PR #40, opened 2020-12-12
fix: remove pointer-events: none, because it causes cursor: not-allow…

vladmonday:
Removing the pointer-events: none for disabled button, because with it the cursor: not-allowed doesn't really work and all disabled button have default cursor instead of the not allowed one.

#### Basic
- [x] PR has description
- [x] Changes to the component are backward compatible (including selectors structure). If not - add to the title of the PR "BREAKABLE_CHANGE""
- [x] All changes to the component are reflected in the ReadMe
#### Style
- [x] CSS selectors are named using [BEM convention](http://getbem.com/naming/) 
- [x] Design is compatible with [Monday Design System](https://design.monday.com/)
#### Storybook
- [x] All the changes to the component should be reflected in the Storybook.
- [x] Component passed Accessibility Plugin checks
#### Tests
- [x] All current tests are passing
- [x] New functionality is covered with tests
- [x] Tests are compliant with TESTING_README.md instructions

### DIFF
--- a/src/components/Button/Button.scss
+++ b/src/components/Button/Button.scss
@@ -139,7 +139,6 @@
           @include theme-prop(background-color, disabled-background-color);
           @include theme-prop(color, disabled-text-color);
           cursor: not-allowed;
-          pointer-events: none;
         }
       }
     }
@@ -209,7 +208,6 @@
         @include theme-prop(border-color, disabled-background-color);
         @include theme-prop(color, disabled-text-color);
         cursor: not-allowed;
-        pointer-events: none;
         &:hover {
           background-color: transparent;
         }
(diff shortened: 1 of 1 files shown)


orrgottlieb:
Hi @vladmonday ,
The reason that we have pointer-events none is - https://github.com/facebook/react/issues/4251
I'm checking to see if it is fixed in 16.13.1 or we might need to upgrade to 17 in order for it to be supported well.
it affects our tooltip when we are wrapping a component with it

orrgottlieb:
Hi @vladmonday,
https://codepen.io/orrgottl/pen/jOMywxL this is actually how React works (attaching the event to the root element), so this is the reason it doesn't work. 
they have several open bugs on the issue.
I rather leave the current issue as it is, We can have a fix for it using native event (which is a much bigger PR)
Let's discuss this
Orr

PR #138, opened 2021-06-06
Feature/hadas/toggle

hadasfa:
https://user-images.githubusercontent.com/72390374/121037247-f77c6780-c7b7-11eb-97ae-fdc76ccba0c8.mov

Description:  add support on new toggle component - with tests and storybook documention

#### Basic
- [ ] Used plop (npm run plop) to create a new component. This command will create a folder for your component and will update index.js file
- [V] PR has description
- [V] New component is functional and using React Hooks. 
- [V] New React Hooks were added to /src/hooks folder (optional)
- [V] Component defines PropTypes
- [ ] There is Readme for the component, which explains the main aspects of how to use it
#### Style
- [V] Styles are added to NewComponent.scss file inside of the NewComponent folder
- [V] CSS selectors are named using [BEM convention](http://getbem.com/naming/) 
- [V] Design is compatible with [Monday Design System](https://design.monday.com/)
#### Storybook
- [V] Stories were added to /src/NewComponent/__stories__/NewComponent.stories.js file
- [ ] Stories include sandbox story with knobs for each property 
- [V] Stories include all flows of using the component
- [V] Component passed Accessibility Plugin checks
#### Tests
- [V] Tests are compliant with TESTING_README.md instructions


src/components/Toggle/Toggle.jsx
```diff
@@ -0,0 +1,93 @@
...
+    onChange,
+    value,
+    name,
+    isDisabled,
+    ariaLabel,
+    ariaControls
+  });
+
+  const className = classNames(`${BASE_TOGGLE_CLASS_NAME}__toggle`, componentClassName, {
+    [`${BASE_TOGGLE_CLASS_NAME}__toggle--selected`]: inputProps.checked,
```
mentaman, line 39: also possible 

{ checked } = inputProps

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,61 @@
...
+.monday-style-toggle{
+  &__wrapper {
+    display: flex;
+    align-items: center;
+  }
+
+  &__toggle {
+    @include reset-button();
+    @include toggle-animation;
+    margin-inline: $spacing-small;
```
mentaman, line 17: image
I think it's not supported in edge which I think we're supposed to support?
https://caniuse.com/?search=margin-inline

src/components/Toggle/Toggle.jsx
```diff
@@ -0,0 +1,93 @@
...
+   */
+  ariaLabel: PropTypes.string,
+  ariaControls: PropTypes.string
+};
+
+Toggle.defaultProps = {
+  id: undefined,
+  componentClassName: "",
+  isDefaultSelected: true,
+  isSelected: undefined,
```
mentaman, line 83: do we need defaultProps for undefined?
I'm okay with it if you are/we do it in other places, but feels a little weird at first look

src/components/Toggle/__tests__/toggle.jest.js
```diff
@@ -0,0 +1,116 @@
...
+    });
+
+    it("renders correctly when selection defined by isSelected (off)", () => {
+      const tree = renderer.create(<Toggle componentClassName="dummy-class-name" isSelected={false} />).toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it("renders correctly when disabled", () => {
+      const tree = renderer.create(<Toggle isDisabled />).toJSON();
+      expect(tree).toMatchSnapshot();
```
mentaman, line 35: where are the snapshots themselves? shouldn't they be commited?
did it pass ci?

src/components/Toggle/Toggle.jsx
```diff
@@ -0,0 +1,93 @@
...
+  isSelected,
+  onChange,
+  value,
+  name,
+  isDisabled,
+  ariaLabel,
+  ariaControls,
+  isHideLabels
+}) => {
+  const { inputProps } = useToggle({
```
mentaman, line 26: not sure I understood what triggers the toggle,
I see it uses checked, but when there is some callback like

onCheckedChange or whatever supposed to be here

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,61 @@
...
+    position: relative;
+    height: 24px;
+    width: 41px;
+    cursor: pointer;
+    border-radius: 100px;
```
mentaman, line 22: why 100px? is it supposed to be that?
there is also border-radius: 100%

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,61 @@
+@import "../../styles/states.scss";
+@import "../../styles/global-css-settings.scss";
+
+@mixin toggle-animation {
+  transition: 100ms $expand-animation-timing;
```
mentaman, line 5: not related to this cr specifically but I see the project use 100ms a lot, maybe worth to keep it as a variable together with $expand-animation-timing ?

src/components/Toggle/__stories__/toggle.stories.js
```diff
@@ -0,0 +1,62 @@
...
+        <Toggle isSelected={false} isDisabled />
+      </StoryStateRow>
+    </StoryWrapper>
+  );
+};
+
+export default {
+  title: "Components/Toggle",
+  component: Toggle,
+  decorators: [withPerformance]
```
mentaman, line 61: do we have components where it's not used?
It's possible to add it to ".storybook/preview.js" instead of explicitly setting it here.
not related to pr though, saw it happens in other stories

mentaman (approved):

src/components/Toggle/Toggle.jsx
hadasfa: Yes this is the standard today :)

src/components/Toggle/Toggle.scss
hadasfa: yes this is not supported :( i have to change it :(

src/components/Toggle/Toggle.scss
```diff
@@ -0,0 +1,61 @@
...
+    align-items: center;
+  }
+
+  &__toggle {
+    @include reset-button();
+    @include toggle-animation;
+    margin-inline: $spacing-small;
+    position: relative;
+    height: 24px;
+    width: 41px;
```
mentaman, line 20: shouldn't it be 4x multipications?

mentaman: or at least 2x

src/components/Toggle/Toggle.jsx
```diff
@@ -0,0 +1,93 @@
...
+
+  // TODO: ask orr about support for all browsers? MARGIN INLINE
+  // TODO: ask about on and off text
+
+  return (
+    <label htmlFor={id} className={`${BASE_TOGGLE_CLASS_NAME}__wrapper`}>
+      <VisuallyHidden>
+        <input disabled={isDisabled} {...inputProps} />
+      </VisuallyHidden>
+      {isHideLabels ? null : <ToggleText>Off</ToggleText>}
```
mentaman, line 52: Off
On

We don't use _t inside monday-ui-react-core?

mentaman, line 52: isHideLabels -> shouldHideLabels, sounds better

src/components/Toggle/Toggle.scss
hadasfa: should be but that's was the figma units... i can double check with evginay

src/components/Toggle/Toggle.jsx
hadasfa: For the open source usage it will be always english.
I will add an  override option and in monday we will use it with (_t("on")) and (_t("off"))

src/components/Toggle/Toggle.scss
hadasfa: @orrgottlieb ?

src/components/Toggle/__stories__/toggle.stories.js
hadasfa: @orrgottlieb what do you think about this?

src/components/Toggle/Toggle.scss
orrgottlieb: i would add times contants - like fast - 50ms, medium - 100ms, slow - 250ms

src/components/Toggle/Toggle.scss
orrgottlieb: either do min-width or have the container decide ?

src/components/Toggle/Toggle.scss
orrgottlieb: why 100px?

PR #154, opened 2021-06-30
fix checkbox focus & add aria labelled by support

hadasfa:
1. fix bug on checkbox focus

https://user-images.githubusercontent.com/72390374/123946450-e6cea400-d9a7-11eb-84cb-00bace26b33c.mov

2. add new prop for checkbox - ariaLabelledBy

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -10,17 +10,17 @@ import "./Checkbox.scss";
 const BASE_CLASS_NAME = "monday-style-checkbox";
 
 export const Checkbox = ({
-                           componentClassName,
-                           label,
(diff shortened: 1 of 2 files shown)


src/components/Checkbox/Checkbox.js
```diff
@@ -70,6 +72,8 @@ export const Checkbox = ({
           iconLabel="checkbox"
           ignoreFocusStyle
           clickable
+          ariaHidden
+          tabindex="-1"
```
orrgottlieb, line 76: you don't need these, just remove the clickable

orrgottlieb (changes requested):

orrgottlieb (approved):

PR #168, opened 2021-07-18
Bug/hadas/a11y feedback

hadasfa:
## Go over this checklist before submitting your Pull Request

- please do not merge before talk with me :) !
- fix toggle curser when disable
- according to Talya feedback about icons: from now on each icon has 3 a11y states:
- **clickble**: tab index :0, aria hidden false, aria label and role:" button"
- **hidden**: _no tab index (instead of -1, because not clickable item should not have any tab index)_, aria-hidden: true , no aria-label 
- not hidden and not clickable (like text): aria-hidden: false, role: img, aria-label, **no tab index (instead of -1, because not clickable item should not have any tab index**
- Custom svg icon is now support ariaLabel and aria hidden (and same logic as above will be activate)
**- we need to notice that after this change icon will not be hidden by default if they have clickable false value**

#### Basic
- [ V] PR has description
- [ V] Changes to the component are backward compatible (including selectors structure). If not - add to the title of the PR "BREAKABLE_CHANGE""
- [ V] All changes to the component are reflected in the ReadMe
- [ v] If component is old and was not compliant with the latest guidelines - it was fixed (optional) 
#### Style
#### Storybook
- [ V] All the changes to the component should be reflected in the Storybook.
- [ V] Component passed Accessibility Plugin checks
#### Tests
- [ v] All current tests are passing
- [ V] Tests are compliant with TESTING_README.md instructions

### DIFF
--- a/src/components/Icon/FontIcon/FontIcon.jsx
+++ b/src/components/Icon/FontIcon/FontIcon.jsx
@@ -2,20 +2,25 @@ import React, { forwardRef } from "react";
 import "./FontIcon.scss";
 import classNames from "classnames";
 
-const FontIcon = forwardRef(({ className, onClick, iconLabel, tabIndex, icon, role = "img", ariaHidden }, iconRef) => {
-  const iconClassName = typeof icon === "function" ? "" : icon;
-  return (
-    <span
-      aria-hidden={ariaHidden}
-      className={classNames(className, "fa", iconClassName)}
-      onClick={onClick}
-      ref={iconRef}
-      aria-label={iconLabel}
-      tabIndex={tabIndex}
-      role={role}
-    >
-      {typeof icon === "function" && icon()}
-    </span>
-  );
-});
+const FontIcon = forwardRef(
+  (
+    { className, onClick, "aria-label": iconLabel, tabIndex, icon, role = "img", "aria-hidden": ariaHidden },
+    iconRef
+  ) => {
+    const iconClassName = typeof icon === "function" ? "" : icon;
+    return (
+      <span
+        aria-hidden={ariaHidden}
+        className={classNames(className, "fa", iconClassName)}
+        onClick={onClick}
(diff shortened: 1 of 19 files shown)


src/components/Icon/FontIcon/FontIcon.jsx
```diff
@@ -2,20 +2,25 @@ import React, { forwardRef } from "react";
...
-      tabIndex={tabIndex}
-      role={role}
-    >
-      {typeof icon === "function" && icon()}
-    </span>
-  );
-});
+const FontIcon = forwardRef(
+  (
+    { className, onClick, "aria-label": iconLabel, tabIndex, icon, role = "img", "aria-hidden": ariaHidden },
```
hadasfa, line 7: aria-hidden, and aria-label is the same props names that IconComponent received (both components created inside the parent Icon component)

src/components/Toggle/Toggle.scss
```diff
@@ -7,6 +7,9 @@
     align-items: center;
     &--disabled {
       opacity: 0.4;
+      & .monday-style-toggle_toggle {
```
hadasfa, line 10: when disabled

src/components/Icon/Icon.jsx
```diff
@@ -25,13 +25,17 @@ const Icon = forwardRef(
...
+    const { screenReaderAccessProps, onClickCallback, computedClassName, iconRef } = useIconProps({
       onClick,
+      iconLabel,
       clickable,
       className,
+      isDecorationOnly: ariaHidden,
       ignoreFocusStyle
     });
 
+    screenReaderAccessProps.tabIndex = externalTabIndex ?? screenReaderAccessProps.tabIndex;
```
orrgottlieb, line 37: why here and not inside the useIconProps?

orrgottlieb (approved):

PR #171, opened 2021-07-21
Feature/hadas/toast link

hadasfa:
Screen Shot 2021-07-21 at 13 40 08

Screen Shot 2021-07-21 at 13 40 20

Screen Shot 2021-07-21 at 13 40 48

Screen Shot 2021-07-21 at 13 41 00

Screen Shot 2021-07-21 at 13 41 27

Description: 
1. few css fixes (for toast looking ok with button and without, with link and without etc)
2. **Before this change:**
Adding a link to toast was not supported.
For adding a button to a toast, you would need to use the standard button component (which includes default colors that do not match the component).
**After this change:**
Link and button components can be added to a toast component by using` <ToastLink/>` and `<ToastButton/>`
At the moment, the API allows you to add a component(button or link) to the content section or to the action section, which is attached to the close button. 
The reason for this not-perfect API is that I did not want to break the component's old API for those who are already using it (which use the action prop for add props that are attached to the right, which is not always what we want to with toast links). Still, later we will need to unify so that the alert banner API will be the same toast API.

### DIFF
--- a/src/components/Toast/__stories__/toast.stories.js
+++ b/src/components/Toast/__stories__/toast.stories.js
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import React, { useState } from "react";
 import { number, select, boolean } from "@storybook/addon-knobs";
 import { TOAST_TYPES } from "../ToastConstants";
 import Toast from "../Toast";
@@ -6,6 +6,8 @@ import Button from "../../Button/Button";
 import Icon from "../../Icon/Icon";
 import Send from "../../Icon/Icons/components/Send";
 import { StoryStateColumn, StoryStateRow } from "../../storybook-helpers";
+import ToastButton from "../ToastButton";
+import ToastLink from "../ToastLink";
 
 export const Toasts = () => {
   const sendIconElement = (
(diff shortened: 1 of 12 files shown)


src/components/Toast/__stories__/toast.stories.js
```diff
@@ -27,35 +29,84 @@ export const Sandbox = () => {
...
-            icon={icon}
-            closeable={knobs.closeable}
-            autoHideDuration={knobs.autoHideDuration}
-            hideIcon={knobs.hideIcon}
-          >
-            Something Happened
-          </Toast>
-        </StoryStateColumn>
-      </StoryStateRow>
+      <StoryStateColumn>
```
MosheZemah, line 50: nice story!

src/published-components.js
```diff
@@ -41,6 +41,8 @@ const publishedComponents = {
   Search: "/src/components/Search/Search.jsx",
   TextField: "/src/components/TextField/TextField.jsx",
   Toast: "/src/components/Toast/Toast.jsx",
+  ToastButton: "/src/components/Toast/ToastButton/ToastButton.jsx",
+  ToastLink: "/src/components/Toast/ToastLink/ToastLink.jsx",
```
MosheZemah, line 45: are you sure we need this components available outside Toast component?

MosheZemah (approved):

src/published-components.js
hadasfa: yes, because the developer need to create the button/or link and pass them to the toast (similar to alert banner)

PR #173, opened 2021-07-25
Fix/tombo/responsive list issues

mentaman:
There were a few issues with the ResponsiveList we found, that caused it to break sometime and draw too much items.
image

1. It didn't draw items that it didn't have place for, so after it calculated their width once,
it wouldn't be able to calculate it again.
since the children here would include only part of the children
image

2. If there were new items added after drawing it, it wouldn't be able to calculate the widthes for the same reason and it'll break

To solve it, I'm drawing it twice now, once for calculation only invisibly with position absolute with the parent width,
and once the actual drawing.


src/components/ResponsiveList/ResponsiveList.jsx
```diff
@@ -14,20 +14,28 @@ const ResponsiveList = forwardRef(
...
       resizeDebounceTime,
+      menuButtonAriaLabel,
       menuButtonProps
     },
     ref
   ) => {
     const componentRef = useRef(null);
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
-    const index = useElementsOverflowingIndex({ ref: componentRef, children, paddingSize, resizeDebounceTime });
+    const index = useElementsOverflowingIndex({
```
mentaman, line 32: From what I saw we're the only one using ResponsiveList and useElementsOverflowingIndex
so there's no risk of breaking changes right?

src/components/ResponsiveList/__stories__/responsiveList.stories.js
```diff
@@ -66,28 +69,51 @@ const DefaultExampleTemplate = (responseListProps) => {
...
     <br />
     <div style={{ width: "80%" }}>
-      <DefaultExampleTemplate menuButtonProps={{component: Bolt}} />
+      <DefaultExampleTemplate menuButtonProps={{component: MoreActions}} />
+    </div>
+  </div>
+}
+
+
+export const ChangeChildrenCheck = () => {
```
mentaman, line 101: without the fix, this story is broken

src/components/ResponsiveList/ResponsiveList.jsx
orrgottlieb: it is open source, so it not known, but this is still a work in progress, i don't mind changing it

orrgottlieb (approved):

PR #175, opened 2021-07-25
responsive list placeholder

mentaman:
Had a bug since it's drawing twice, once visible and once invisible.

Used responsiveListPlaceholder to fix it.

In the storybook example, I checked that it added an id to the placeholder search.
otherwise it had two ids and the accessibility tab shout about it.

### DIFF
--- a/src/components/ResponsiveList/__stories__/responsiveList.stories.js
+++ b/src/components/ResponsiveList/__stories__/responsiveList.stories.js
@@ -39,7 +39,13 @@ const DefaultExampleTemplate = (responseListProps) => {
     <SplitButton size={SplitButton.sizes.MAIN} marginRight secondaryDialogContent={<SecondaryContentComponent />}>
       Add Item
     </SplitButton>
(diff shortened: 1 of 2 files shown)


src/components/ResponsiveList/__stories__/responsiveList.stories.js
```diff
@@ -39,7 +39,13 @@ const DefaultExampleTemplate = (responseListProps) => {
...
     </SplitButton>
-    <div className="responsive-lst-search">
+    <div className="responsive-lst-search" responsiveListPlaceholder={<Search
+      id={"search-icon-inside"}
+      size={Search.sizes.MEDIUM}
+      placeholder="search for content"
+      iconName={() => <SearchIcon />}
+      secondaryIconName={() => <CloseSmall />}
+    />}>
       <Search
```
mentaman, line 49: The invisible now has this id:
image
so no duplicates ids

orrgottlieb (approved):

PR #183, opened 2021-08-01
Feature/tombo/test prop types

mentaman:
- it now tests in the jest and stories, that all the propTypes passed are valid
- eslint had "only-warn" before, so it never failed even on things like "no-undef"
also the one in "npm run eslint" didn't do anything, no relevant file was included.


src/components/Chips/Chips.jsx
```diff
@@ -79,9 +79,9 @@ Chips.propTypes = {
   disabled: PropTypes.bool,
   readOnly: PropTypes.bool,
   /** Icon to place on the right */
-  rightIcon: PropTypes.string,
+  rightIcon: PropTypes.node,
```
orrgottlieb, line 82: this should be string, function or node

orrgottlieb (approved):

PR #199, opened 2021-08-10
Ruti/update multi step indicator to be clickable

rutikevesh:
Now multiStepIndicator will enable clickable steps.
It allows a responsive steps walkthrough, controlled by the component.


src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.jsx
```diff
@@ -90,9 +95,11 @@ const StepIndicator = ({
   return (
     <li
       className={cx(...getClassNamesWithSuffix(""), stepComponentClassName, {
-        [baseClassNameWithAnimation]: statusChangeAnimationState
+        [baseClassNameWithAnimation]: statusChangeAnimationState,
+        clickable: onClick
       })}
       aria-label={ariaLabel}
+      onClick={handleClick}
```
udidoron, line 102: Once an element is clickable, there are a few changes it needs in order to be keyboard-accessible:

- An `onKeyPress` handler (you can use the existing `useKeyEvent` hook to get it easily) to catch 'Enter' and 'Space' key presses, and apply the same `handleClick` function.
- A `tabIndex={0}` attribute

After adding it (if you choose to), make sure it looks OK when tabbing, and that keyboard navigation works.

udidoron (approved):

src/components/MultiStepIndicator/MultiStepIndicator.jsx
```diff
@@ -91,7 +95,8 @@ MultiStepIndicator.defaultProps = {
   type: MultiStepIndicator.types.PRIMARY,
   steps: [],
   fulfilledStepIcon: Check,
-  fulfilledStepIconType: Icon.type.SVG
+  fulfilledStepIconType: Icon.type.SVG,
+  onClick: null
```
orrgottlieb, line 99: why null? why not noop function ?

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.jsx
```diff
@@ -40,13 +43,25 @@ const StepIndicator = ({
...
+  };
+
   // Event listeners for removing animation.
   useEventListener({
     eventName: "animationend",
     callback: disableStatusChangeAnimation,
     ref: componentRef
   });
 
+  const KEYS = [keyCodes.ENTER, keyCodes.SPACE];
```
orrgottlieb, line 57: please move KEYS const outside the function as it will create new reference array each render cycle

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.jsx
```diff
@@ -90,11 +105,13 @@ const StepIndicator = ({
...
       className={cx(...getClassNamesWithSuffix(""), stepComponentClassName, {
-        [baseClassNameWithAnimation]: statusChangeAnimationState
+        [baseClassNameWithAnimation]: statusChangeAnimationState,
+        clickable: onClick
       })}
       aria-label={ariaLabel}
+      onClick={handleClick}
     >
-      <div className={cx(...getClassNamesWithSuffix("__number-container"))} ref={componentRef}>
+      <div className={cx(...getClassNamesWithSuffix("__number-container"))} ref={componentRef} tabIndex="0">
```
orrgottlieb, line 114: please add role="button" or use useButton helper from react-aria

orrgottlieb (changes requested):

src/components/MultiStepIndicator/__tests__/multiStepIndicator.jest.js
```diff
@@ -31,5 +31,34 @@ describe("MultiStepIndicator", () => {
       const tree = renderer.create(<MultiStepIndicator type={MultiStepIndicator.types.SUCCESS} steps={exampleSteps} />);
       expect(tree).toMatchSnapshot();
     });
+
+    it("Render correctly with clickable steps", () => {
```
DorShakedMonday, line 35: We may want to test that clicks are happening instead of the DOM structure.
What do you say?

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.jsx
```diff
@@ -90,11 +106,18 @@ const StepIndicator = ({
   return (
     <li
```
DorShakedMonday, line 107: Maybe change this `li` into a `button`?
When we add this onClick here, it may represent the element better

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.jsx
```diff
@@ -40,13 +46,23 @@ const StepIndicator = ({
 
   const isStatusTransition = useCallback(() => prevStatusRef.current !== status, [prevStatusRef, status]);
 
+  const handleClick = () => {
```
DorShakedMonday, line 49: useCallback?
it is transferred to useKeyEvent

DorShakedMonday (approved):

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.jsx
rutikevesh: But I did add a role of button to the element, and the step is an item in a list so I think it makes sense no?

PR #224, opened 2021-09-17
feat: add support for indeterminate checkbox

fatihict:
Resolves: #223

Screenshot_20210917_163207

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -47,7 +47,7 @@ export const Checkbox = ({
 
   useEffect(() => {
     inputRef.current.indeterminate = indeterminate;
-  }, []);
+  }, [indeterminate]);
 
   return (
     // eslint-disable-next-line jsx-a11y/no-noninteractive-element-interactions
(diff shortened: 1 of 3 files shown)


src/components/Checkbox/Checkbox.js
```diff
@@ -43,6 +45,10 @@ export const Checkbox = ({
     overrideDefaultChecked = false;
   }
 
+  useEffect(() => {
+    inputRef.current.indeterminate = indeterminate;
```
orrgottlieb, line 49: Hi @fatihict ,
Thanks for this improvement!, I have only one question - 
This will render only on the initial render, don't you think it will be better if `indeterminate` will be given to the dependency array, so it could be controlled by the consumer?

Thanks

orrgottlieb (changes requested):

src/components/Checkbox/Checkbox.js
fatihict: Hey @orrgottlieb,
Fair point, I missed that. Fixed your comment in a new commit.

orrgottlieb: thanks! 🤩

PR #307, opened 2021-10-29
added hide button support for attention box

etaylib:
example:
image
image

in stories:
image


hadasfa:
we already merged a pr which handle this :)

PR #310, opened 2021-10-31
stylelint rules

mentaman:
stylelint rules to our scss files


.stylelintrc.json
```diff
@@ -0,0 +1,51 @@
...
+    "property-no-vendor-prefix": null,
+    "scss/at-mixin-argumentless-call-parentheses": null,
+    "color-function-notation": null,
+    "no-empty-source": [true, {"severity": "warning"}],
+    "keyframes-name-pattern": null,
+    "max-line-length": [140, {"severity": "error"}],
+    "function-calc-no-unspaced-operator": null,
+    "no-duplicate-selectors": [true, {"severity": "warning"}],
+
+    "scss/dollar-variable-empty-line-before": null,
```
mentaman, line 28: everything under this line can be autofixed with npm run stylelint-fix

src/components/Tabs/TabsContext/TabsContext.scss
```diff
@@ -1,6 +0,0 @@
-@import "../../../styles/themes";
-@import "../../../styles/typography";
```
sahariko, line 2: These files have side-effects, are you absolutely sure it doesn't mess up anything?

package.json
```diff
@@ -25,6 +25,8 @@
     "fix-lint": "prettier --write \"{,!(node_modules)/**/}*.{js,jsx}\"",
     "eslint": "eslint \"./src/**/*.{js,jsx}\"",
     "eslint-fix": "eslint \"./src/**/*.{js,jsx}\" --fix",
+    "stylelint": "stylelint \"**/*.scss\"",
+    "stylelint-fix": "stylelint \"**/*.scss\" --fix",
```
sahariko, line 29: Not sure this one is needed (I know we have an `eslint-fix`, but I don't think that one's needed as well).

sahariko (approved):
2 small comments, otherwise LGTM!

src/components/Tabs/TabsContext/TabsContext.scss
mentaman: @sahariko it's imported in other places
image
So I think it'd still have these effects. 

I can keep the import just in case.
the linter was annoyed about the empty block. it didn't care about the imports

sahariko: Let's keep it for now then.

orrgottlieb: please keep the import as we support the option to only import a single file from the library - so the basic styles should be added

orrgottlieb (changes requested):
please see other PR https://github.com/mondaycom/monday-ui-react-core/pull/336
waiting for comments an PR:336

src/components/Tabs/TabsContext/TabsContext.scss
mentaman: @orrgottlieb I already put it the back, it's outdated

orrgottlieb (approved):

PR #333, opened 2021-11-15
amitha/fix/text-field-focus-animation-in-dialog

amit-hanoch:
When autofocus in dialog and scrolling to the bottom of the screen the screen "jumps" because the focus collides with the scroll to, adding requestAnimationFrame is the solution.

Read about it a little bit and a lot of UI libraries like material-ui use setTimeout 0 to solve this issue and it's pretty common so  I think we need to make sure all .focus() to happen with request animation frame from now on.

#### Basic
- [ ] PR has description
- [ ] Changes to the component are backward compatible (including selectors structure). If not - add to the title of the PR "BREAKABLE_CHANGE""
- [ ] All changes to the component are reflected in the ReadMe
#### Style
#### Storybook
- [ ] Component passed Accessibility Plugin checks
#### Tests
- [ ] All current tests are passing
- [ ] Tests are compliant with TESTING_README.md instructions


src/components/TextField/TextField.jsx
```diff
@@ -104,6 +105,12 @@ const TextField = forwardRef(
       elementType: "div"
     });
 
+    useEffect(() => {
+      if (inputRef.current && autoFocus) {
```
orrgottlieb, line 109: add `cancelAnimationFrame` on the return function of the `useEffect` please

orrgottlieb (changes requested):

src/components/TextField/TextField.jsx
amit-hanoch: @orrgottlieb found that i missed combo aswell please look there too :)

orrgottlieb (approved):

PR #354, opened 2021-11-23
Multiselect component

sahariko:
### Main Change
Add support for multiple values to our `<Dropdown>` component, via the following new props:
- `multi` - When passed, will render selected options as chips.
- `multiline` - When passed, and the component is constrained with a specific width, will cause it to drop line instead of displaying a counter of unrendered chips.
- `onOptionRemove` - Will be called whenever a chip’s “X” is clicked.
- `onOptionSelect` - Will be called whenever an option is clicked inside the menu.
- `onClear` - Will be called whenever the generic clear button is clicked.

Screen Shot 2021-11-23 at 12 42 38

#### Infrastructure

- Add support for [CSS modules](https://github.com/css-modules/css-modules).

`<Dropdown>`

- Separate test driver to its own file.
- Remove unused (and non-existent) `isOpen` prop from `<Menu>`.

#### `<Chips>`

- Enable rendering without animations (looks bad in the context of multiselect).
- Support passing `data-testid`.

`<Counter>`

- Enable rendering without animations (looks bad in the context of multiselect).
- Support new `onMouseDown` prop, needed to prevent `<Dropdown>` from opening the menu when the counter is clicked.

`<Dialog>`

- Fix bug with `open` prop, which caused it to be “ignored” when the component’s internal state changes.
- Move default tooltip arrow styling from `<Tooltip>` to this component.

`<Icon>`

- Support passing `data-testid`.


src/components/Dialog/Dialog.jsx
```diff
@@ -181,7 +181,7 @@ export default class Dialog extends PureComponent {
     const { isOpen } = this.state;
     const { open } = this.props;
 
-    return isOpen || open;
+    return typeof open !== "undefined" ? open : isOpen;
```
orrgottlieb, line 184: add prop for controlled only (until 2.0)

orrgottlieb (changes requested):

orrgottlieb (approved):

PR #356, opened 2021-11-23
Create hook for a11y consist behavior for notifications

hadasfa:
After talking with Talya and Adi we come to the following conclusions:
When display visual element which include alert aka urgent notification, we will pass role=alert.
Otherwise we will pass role=status only if the element does not include any actions inside (aka buttons).
If this is the case (the element include actions inside) **we will not pass the status role** and we will move the focus to the element (for read the message by screen reader and for give access to the actions to the keyboard users).

#### Basic
- [ v] PR has description
- [ v] Changes to the component are backward compatible (including selectors structure). If not - add to the title of the PR "BREAKABLE_CHANGE""
- [v ] All changes to the component are reflected in the ReadMe
#### Style
#### Storybook
- [v ] Component passed Accessibility Plugin checks
#### Tests
- [ v] All current tests are passing
- [ v] New functionality is covered with tests
- [ v] Tests are compliant with TESTING_README.md instructions


src/hooks/useA11yNotification.js
```diff
@@ -0,0 +1,25 @@
...
+export function useA11yNotification({
+  isUrgent = false,
+  isAriaLiveHandledOutside = false,
+  isIncludeActions = false,
+  ref
+}) {
+  // If notification include actions, focus need to pass move to the notification itself
+  useEffect(() => {
+    if (isIncludeActions && ref && ref.current) {
+      ref.current.focus();
```
hadasfa, line 12: I am not sure if the focus handling should be part of monday-ui-react-core  hook or be handled separately in the monolith.
I thought that it will be better here because than developers could use this hook when developing custom notification components.

src/components/AttentionBox/AttentionBox.jsx
```diff
@@ -7,6 +7,7 @@ import AlertIcon from "../Icon/Icons/components/Alert";
 import { baseClassName, closeClassName, compactClassName, ATTENTION_BOX_TYPES } from "./AttentionBoxConstants";
 import "./AttentionBox.scss";
 import { backwardCompatibilityForProperties } from "../../helpers/backwardCompatibilityForProperties";
+import { useA11yNotification } from "../../hooks/useA11yNotification";
```
hadasfa, line 10: **Please not merge this PR until I return from my vacation (I want to check it in dapulse right after release the new version)**

src/hooks/useA11yNotification.js
```diff
@@ -0,0 +1,13 @@
+import { useMemo } from "react";
+
+export function useA11yNotification({ isUrgent = false, isAriaLiveHandledOutside = false }) {
```
amit-hanoch, line 3: would change it to isAlert and not isUrgent - code is pretty hard to understand when using this

for example - why is Toast.types.NEGATIVE urgent?

amit-hanoch (approved):

PR #362, opened 2021-11-29
Storybook/hadasfa/design fix

hadasfa:
Evgeniy design requests for the new stories template & fix some errors in storybook props & usage multiple line appearance


src/components/Checkbox/Checkbox.js
```diff
@@ -96,7 +96,7 @@ export const Checkbox = ({
 Checkbox.propTypes = {
   id: PropTypes.string,
   className: PropTypes.string,
-  label: PropTypes.string,
+  label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
hadasfa, line 99: in our examples we use elements...  there is a reason why not?

PR #369, opened 2021-12-01
add MenuButton onClick prop

shalomscott:
The MenuButton at its core is.. a button. I can imagine there are plenty of legitimate scenarios where a user would want to perform some logic when it's clicked. What's the harm in exposing an `onClick`? This is in addition to and would not directly affect the "show dialog" trigger. 

Also, when MenuButton was not passed any `children`, instead of rendering a Dialog with an empty div, just don't render the dialog at all. (The Dialog component already gracefully handles being passed `content` of null)

#### Basic
- [x] PR has description
- [x] Changes to the component are backward compatible (including selectors structure). If not - add to the title of the PR "BREAKABLE_CHANGE""
- [ ] All changes to the component are reflected in the ReadMe
#### Style
#### Storybook
- [ ] Component passed Accessibility Plugin checks
#### Tests
- [ ] All current tests are passing
- [ ] Tests are compliant with TESTING_README.md instructions

### DIFF
--- a/src/components/MenuButton/MenuButton.jsx
+++ b/src/components/MenuButton/MenuButton.jsx
@@ -3,6 +3,7 @@ import React, { useCallback, useState, useMemo, useRef, useLayoutEffect } from "
 import PropTypes from "prop-types";
 import cx from "classnames";
 import NOOP from "lodash/noop";
+import isFunction from "lodash/isFunction";
 import Dialog from "../Dialog/Dialog";
 import Menu from "../Icon/Icons/components/Menu";
 import DialogContentContainer from "../DialogContentContainer/DialogContentContainer";
@@ -30,6 +31,7 @@ const MenuButton = ({
   component,
   size,
   open,
+  onClick,
   zIndex,
   ariaLabel,
   closeDialogOnContentClick,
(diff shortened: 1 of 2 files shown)


src/components/MenuButton/MenuButton.jsx
```diff
@@ -139,6 +141,8 @@ const MenuButton = ({
   const onMouseUp = event => {
     if (disabled) {
       event.currentTarget.blur();
+    } else {
+      isFunction(onClick) && onClick();
```
orrgottlieb, line 145:
```
if(disabled) {
      event.currentTarget.blur();
     return
}

onClick(event);
```

src/components/MenuButton/MenuButton.jsx
```diff
@@ -328,6 +333,7 @@ MenuButton.defaultProps = {
   component: Menu,
   size: MenuButtonSizes.SMALL,
   open: false,
+  onClick: undefined,
```
orrgottlieb, line 336: provide default noop function please

orrgottlieb (changes requested):

src/components/MenuButton/MenuButton.jsx
shalomscott: I'll add the default NOOP instead of the isFunction() check?

orrgottlieb (approved):

PR #374, opened 2021-12-05
Fix `<Dropdown>` multi-value bug

sahariko:
At the moment, if the `<Dropdown>` is just a regular, controlled dropdown with a single value, it'll break since we try to run `.reduce` on it.

### DIFF
--- a/src/components/Dropdown/__tests__/dropdown.jest.js
+++ b/src/components/Dropdown/__tests__/dropdown.jest.js
@@ -96,7 +96,7 @@ describe("Dropdown", () => {
     });
   });
 
-  describe.only("Controlled", () => {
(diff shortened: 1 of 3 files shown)


src/components/Dropdown/__tests__/dropdown.jest.js
```diff
@@ -96,6 +96,39 @@ describe("Dropdown", () => {
     });
   });
 
+  describe.only("Controlled", () => {
```
orrgottlieb, line 99: please remove only :)

orrgottlieb (changes requested):

orrgottlieb (approved):

PR #385, opened 2021-12-13
expose colors-utils

ofersh-monday:
expose colors-utils so we could use storybook colors in monday repo


src/components/index.js
```diff
@@ -70,3 +70,4 @@ export { default as TextWithHighlight } from "./TextWithHighlight/TextWithHighli
 export { default as Accordion } from "./Accordion/Accordion/Accordion";
 export { default as AccordionItem } from "./Accordion/AccordionItem/AccordionItem";
 export { default as Clickable } from "./Clickable/Clickable";
+export * from "../utils/colors-utils";
```
hadasfa, line 73: its not a component and therefore I think you can remove this

ofersh-monday: @orrgottlieb wdyt?

hadasfa: in the future we will create separate index files for hooks, utils and components

hadasfa (approved):

PR #390, opened 2021-12-14
Fix bugs on allow multiple on click

hadasfa:
Until now the allow multiple functionality was buggy because the on click functionality was not working as excepted

### DIFF
--- a/src/components/Accordion/Accordion/Accordion.jsx
+++ b/src/components/Accordion/Accordion/Accordion.jsx
@@ -33,9 +33,12 @@ const Accordion = forwardRef(({ children: originalChildren, allowMultiple, defau
   const onChildClick = useCallback(
     itemIndex => {
       if (allowMultiple) {
-        const newExpandedItems = expandedItems;
+        const newExpandedItems = [...expandedItems];
         if (isChildExpanded(itemIndex)) {
-          newExpandedItems.pop(itemIndex);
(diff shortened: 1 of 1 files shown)


src/components/Accordion/Accordion/Accordion.jsx
```diff
@@ -33,9 +33,12 @@ const Accordion = forwardRef(({ children: originalChildren, allowMultiple, defau
   const onChildClick = useCallback(
     itemIndex => {
       if (allowMultiple) {
-        const newExpandedItems = expandedItems;
+        const newExpandedItems = [...expandedItems];
```
hadasfa, line 36: the first bug was that changing the state does not cause a render because the state was mutable

src/components/Accordion/Accordion/Accordion.jsx
```diff
@@ -33,9 +33,12 @@ const Accordion = forwardRef(({ children: originalChildren, allowMultiple, defau
...
         if (isChildExpanded(itemIndex)) {
-          newExpandedItems.pop(itemIndex);
+          const index = newExpandedItems.indexOf(itemIndex);
+          if (index > -1) {
+            newExpandedItems.splice(index, 1);
```
hadasfa, line 40: the second bug was that clicking on already selected drawer caused removing the last opened drawer instead of the one we click on.

MosheZemah (approved):

PR #402, opened 2021-12-21
Storybook/hadas/controls infra

hadasfa:
until now we can't set enums in the controls because storybook can't take automatically enums values. therefore we need to set the enums props fields custom. I create a function for make this easier and used it in few stories (not done fo over all of them, we can do this next week in the foundation iteration part 2).
Also add support there for adding icons in the controls.


src/components/SplitButton/__stories__/splitButton.stories.mdx
```diff
@@ -12,6 +12,11 @@ import "./splitButton.stories.scss";
 <Meta
   title="Buttons/Split Button"
   component={ SplitButton }
+  argTypes={{
+    d: {
```
orrgottlieb, line 16: ?

orrgottlieb (approved):
amazing!!! very very very cool!

src/components/SplitButton/__stories__/splitButton.stories.mdx
hadasfa: deleted :)

PR #417, opened 2021-12-28
change className props to consist naming

hadasfa:
change className props to consist naming #417

### DIFF
--- a/src/components/ButtonGroup/ButtonGroup.jsx
+++ b/src/components/ButtonGroup/ButtonGroup.jsx
@@ -132,7 +132,7 @@ ButtonGroup.sizes = Button.sizes;
 ButtonGroup.kinds = Button.kinds;
 
 ButtonGroup.defaultProps = {
-  className: "",
+  className: undefined,
   value: "",
   name: "",
   disabled: false,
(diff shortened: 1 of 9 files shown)


src/components/ButtonGroup/ButtonGroup.jsx
```diff
@@ -128,7 +132,7 @@ ButtonGroup.sizes = Button.sizes;
 ButtonGroup.kinds = Button.kinds;
 
 ButtonGroup.defaultProps = {
-  componentClassName: "",
+  className: undefined,
```
sahariko, line 135: Don't think it's needed.

src/components/Link/Link.jsx
```diff
@@ -89,7 +93,7 @@ Link.defaultProps = {
   /**
    * Classname to be added to the component container
    */
-  componentClassName: "",
+  className: undefined,
```
sahariko, line 96: Same.

MosheZemah (approved):

src/components/ButtonGroup/ButtonGroup.jsx
hadasfa: That reason for this is because if the  class name will contain default value we will ignore the value of the deprecated componentClassName.

hadasfa: I can set a default value check for the backward compatibility function but I think it's not really matter (this way or another)

PR #424, opened 2022-01-02
feat(slider): create new component - Slider

niksa-monday:
Add new component Slider.

https://monday.monday.com/boards/245345663/views/39363783/pulses/2091381927

1. Slider can be draggable or Jump clickable
2. Need an ability to adjust steps on the slider
3. Slider may be easily color adjusted form our keys  
4. Keyboard focus must work with tab  and the arrow must be able to adjust the slider by 10%
5. After adjustment the one i clicked will be keyboard movable

Slider-Requirements

#### Basic
- [x] PR has description.
#### Style
- [ ] Styles are added to `NewComponent.modules.scss` file inside of the `NewComponent` folder.  - will be done lated in separate Improv PR
- [ ] Component uses CSS Modules. - will be done lated in separate Improv PR
#### Storybook
#### Tests


src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+import { SliderProvider } from "./SliderContext";
+import SliderBase from "./SliderBase/SliderBase";
+import SliderInfix from "./SliderInfix";
+import { createBemBlockHelper } from "../../helpers/bem-helper";
+
+const Slider = forwardRef(
+  (
+    {
+      ariaLabel,
+      ariaLabeledBy,
```
hadasfa, line 17: please change to ariaLabelledby (to be as close as possible to native html syntax aria-labelledby

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
+import React, { forwardRef, useRef } from "react";
+import PropTypes from "prop-types";
+import { COLORS_BASIC, SIZES_BASIC } from "../../constants";
+import useMergeRefs from "../../hooks/useMergeRefs";
+import "./Slider.scss";
```
hadasfa, line 5: please use css modules while implement new components instead of bem (this is a new convention and therefore we it is hard to find references)

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+      infixOptions: {
+        prefix,
+        postfix,
+        indicateSelection
+      }
+    };
+    return (
+      <SliderProvider {...providerProps}>
+        <div
+          className={bem("", { disabled, "value-shown": showValue }, className)}
```
hadasfa, line 72: I kind of not understand what is the new bem helper and how it is different from the old one. But because now we move to use CSS Modules I am not sure it is still necessary

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+   *   - should be used in Controlled Mode only
+   *   - in isRange mode should be an array of two numbers
+   */
+  value: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.number), PropTypes.number]),
+  /**
+   * Default `value` if value not specified
+   *  - should be used in NON-Controlled mode to set initial Value
+   *  - in isRange mode should be an array of two numbers
+   */
+  valueDefault: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.number), PropTypes.number]),
```
hadasfa, line 166: let's change prop name to defaultValue for creating a consistent API with other components props

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Options for initial/start/prefix element, it can be one of:
+   *  - Any Component (react component, node, text, number etc.)
+   *  - Or it can be an object of options for Icons component (see Icon components props)
+   *  - Or it can be an object for Label (Icon, Title - and other components)
+   *  - Or it can be Render Props Function witch are getting value and valueText
+   */
+  prefix: PropTypes.oneOfType([
+    PropTypes.shape({
+      icon: PropTypes.string.isRequired
```
hadasfa, line 192: Is it the our design system icon? If so it can be also a function

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+  min: PropTypes.number,
+  /**
+   * Optional onChange callback (for outer trigger or Controlled mode)
+   * - required in Controlled Mode
+   */
+  onChange: PropTypes.func,
+  /**
+   * Step for Component (Slider) changes
+   */
+  step: PropTypes.number,
```
hadasfa, line 146: I did not understand what this prop does. I tried to play with it in the controls and nothing happens. what it should do?

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+    };
+    return (
+      <SliderProvider {...providerProps}>
+        <div
+          className={bem("", { disabled, "value-shown": showValue }, className)}
+          data-testid={dataTestId}
+          id={id}
+          ref={mergedRef}
+        >
+          <SliderInfix kind="prefix" />
```
hadasfa, line 77: Why not setting a enum for the kind prop instead of set a string value?
Also, I guess the infix is the slider line and the slider base is the circle between the slide lines?
why not using simpler names for the slider sub components?

src/components/Slider/SliderBase/SliderBase.jsx
```diff
@@ -0,0 +1,90 @@
...
+import SliderThumb from "./SliderThumb";
+import { isArrowDownEvent, isArrowLeftEvent, isArrowRightEvent, isArrowUpEvent } from "../../../utils/dom-event-utils";
+
+function getKey(index) {
+  return index;
+}
+
+const SliderBase = forwardRef(({ className }, ref) => {
+  const { color, disabled, dragging, size, consumerBem, shapeTestId } = useSliderUi();
+  const { isRange, min, max, step, value } = useSliderSelection();
```
hadasfa, line 20: I really love the separation to hooks here 😍

src/components/Slider/SliderBase/SliderBase.jsx
```diff
@@ -0,0 +1,90 @@
...
+  const { changeValue, increaseValue, decreaseValue } = useSliderActions();
+  const { railCoords, railRef } = useSliderRail({ min, max, step, ref });
+  const { dimension, offset, positions } = calcDimensions({ isRange, max, min, value });
+
+  function handlePointerMove(e) {
```
hadasfa, line 25: If I remember correctly, when the functions defined inside the functional component it will create new instance of the functions for each render.
Isn't it will be more efficient to use here useCallback here for define the functions?

src/components/Slider/SliderBase/SliderBase.scss
```diff
@@ -0,0 +1,199 @@
...
+          width: 24px;
+        }
+      }
+    }
+
+    &--large {
+      padding: 0 12px;
+      .monday-slider {
+        &__rail {
+          border-radius: 8px;
```
hadasfa, line 134: we have border radius variables for few fixed border radius sizes

src/components/Slider/SliderBase/SliderRail.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { forwardRef } from "react";
+import PropTypes from "prop-types";
+import { bem } from "../SliderCommons";
+
+const SliderRail = forwardRef(({ className, children, onClick }, ref) => {
+  function handleClick(e) {
```
hadasfa, line 6: I think create function inside the render function will create new instance for each render of the slider rail. Why not use useCallback ?

src/components/Slider/SliderBase/SliderThumb.jsx
```diff
@@ -0,0 +1,94 @@
...
+const tooltipShowDelay = 300;
+const tooltipPosition = DialogPositions.TOP;
+
+const SliderThumb = ({ className, index, position }) => {
+  const { max, min, value, valueText } = useSliderSelection(index);
+  const { ariaLabel, ariaLabeledBy, disabled, focused, shapeTestId, showValue } = useSliderUi();
+  const { setFocused, setDragging } = useSliderActions();
+  const ref = useRef(null);
+
+  function handleBlur() {
```
hadasfa, line 17: same comment about the functions inside the component render function :)

src/components/Slider/SliderCommons.js
```diff
@@ -0,0 +1,10 @@
+import { createBemBlockHelper } from "../../helpers/bem-helper";
+
```
hadasfa, line 2: In other files we use the following convention for constants files - ComponentNameConstants.js
Let's rename the file name to be according to this convention.

src/components/Slider/SliderContext.js
```diff
@@ -0,0 +1,192 @@
...
+  const selectionContextValue = {
+    isRange,
+    max,
+    min,
+    step,
+    value: actualValue,
+    valueText: actualValueText
+  };
+
+  function increaseValue() {
```
hadasfa, line 100: same question regarding define functions inside the functional component

src/components/Slider/SliderHelpers.js
```diff
@@ -0,0 +1,52 @@
+export function calcDimensions({ isRange, max, min, value }) {
+  if (!isRange) {
```
hadasfa, line 2: why not convert it to hook and perform this calculations only when one of the arguments are changes?

hadasfa: or at least create a hook that using this function instead of using it directly

src/components/Slider/SliderHooks.js
```diff
@@ -0,0 +1,83 @@
...
+  if (!Array.isArray(value)) {
+    return ensureSingleValueText(valueText, value, formatter);
+  }
+  return value.map((valueSingle, index) => {
+    const valueTextSingle = Array.isArray(valueText) ? valueText[index] : undefined;
+    return ensureSingleValueText(valueTextSingle, valueSingle, formatter);
+  });
+}
+
+export function useControlledOrInternal(value) {
```
hadasfa, line 26: suggestion for simpler function name: "useIsStateControlledFromOutside" 
the current name sounds like it will return an enum

src/components/Slider/SliderInfix.jsx
```diff
@@ -0,0 +1,69 @@
...
+import Label from "../Label/Label";
+import Icon from "../Icon/Icon";
+
+const defaultIconProps = {
+  clickable: false,
+  ignoreFocusStyle: true
+};
+
+function getCurrentLabel({ isPostfix, isRange, valueText }) {
+  if (!isRange) {
```
hadasfa, line 14: maybe will be prettier to set it like this?

if (isRange) return isPostfix ? valueText[1] : valueText[0];
else return valueText;

src/components/Slider/SliderInfix.jsx
```diff
@@ -0,0 +1,69 @@
...
+  if (typeof infix === "function") {
+    return [true, "", infix(value, valueText)];
+  }
+  if (typeof infix === "string") {
+    return [true, "txt", infix];
+  }
+  if (typeof infix === "undefined") {
+    return [false, "", null];
+  }
+  return [true, "", infix];
```
hadasfa, line 47: I would return an object with named fields as return value because it would be more readable and eaiser to understand what is true means and what is "" means.

src/components/Slider/SliderInfix.jsx
```diff
@@ -0,0 +1,69 @@
...
+    return valueText;
+  }
+  if (isPostfix) {
+    return valueText[1];
+  }
+  return valueText[0];
+}
+
+function useSliderInfixComponent(kind) {
+  const { prefix, postfix, indicateSelection } = useSliderInfix();
```
hadasfa, line 24: please move the hooks to a separate file

src/components/Slider/__stories__/Slider.stories.mdx
```diff
@@ -0,0 +1,219 @@
...
+  </Story>
+</Canvas>
+
+## Props
+
+<ArgsTable of={Slider} />
+
+## Usage
+
+<UsageGuidelines guidelines={["Usage guideline 1", "Usage guideline 2", "Usage guideline 3"]} />
```
hadasfa, line 50: Please check with Evigny if he have any usage and tip to insert to the component story. if he does not have any we can remove this section.

src/components/Slider/__stories__/Slider.stories.mdx
```diff
@@ -0,0 +1,219 @@
...
+    <Slider indicateSelection={true} valueDefault={12} size={Slider.sizes.SMALL} />
+    <Slider indicateSelection={true} isRange={true} valueDefault={[12, 55]} size={Slider.sizes.SMALL} />
+    <Slider prefix={{ icon: Sound }} indicateSelection={true} valueDefault={70} size={Slider.sizes.SMALL} />
+    <Slider prefix={{ icon: Video }} postfix={{ icon: Sound }} valueDefault={45} size={Slider.sizes.MEDIUM} />
+    <Slider prefix="Vol" indicateSelection={true} valueDefault={0} size={Slider.sizes.LARGE} />
+  </Story>
+</Canvas>
+
+## Do’s and Don’ts
+<ComponentRules
```
hadasfa, line 119: Same here. If we don't have any requests from Evigeny for dos and don'ts we can remove this section

src/components/Slider/__stories__/Slider.stories.mdx
```diff
@@ -0,0 +1,219 @@
...
+      },
+      negative: {
+        component:<Slider />,
+        description:"Bad practice description."
+      }
+    }
+  ]}
+/>
+
+## Use cases and examples
```
hadasfa, line 144: same here

src/constants/sizes.js
```diff
@@ -6,6 +6,12 @@ export const SIZES = Object.freeze({
   LARGE: "large"
 });
 
+export const SIZES_BASIC = Object.freeze({
```
hadasfa, line 9: That's great!
but maybe let's move the SIZES_BASIC to be before the SIZES init and do something like this:
SIZES = Object.freeze({
  XXS: "xxs",
  XS: "xs",
  SMALL: "small",
...SIZES_BASIC
});

src/helpers/bem-helper.js
```diff
@@ -6,3 +9,17 @@ export function BEMClass(componentName) {
     return className;
   };
 }
+
+export function createBemBlockHelper(block, { isConsume } = {}) {
```
hadasfa, line 13: I think it will be better not to add this helper right now because we moving to css modules

src/constants/general.js
```diff
@@ -0,0 +1,7 @@
+export const BEM_PREFIX = "monday";
+
+export const COLORS_BASIC = {
```
hadasfa, line 3: that's great 🚀

src/components/Slider/SliderBase/SliderRail.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { forwardRef } from "react";
+import PropTypes from "prop-types";
+import { bem } from "../SliderCommons";
+
+const SliderRail = forwardRef(({ className, children, onClick }, ref) => {
```
hadasfa, line 5: Is this name is standard? Maybe consider to rewrite is to SliderClickSurface?

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Attribute `id` to be added to the component-root-node
+   */
+  id: PropTypes.string,
+  /**
+   * If true switch slider to RRange mode (two Thumbs)
+   */
+  isRange: PropTypes.bool,
```
hadasfa, line 129: we don't use anymore on is prefix. I think we can remove this prop ad calculate if the slide is in range value according to it's value, if the value is array with 2 value therefore we will display a range type slider.

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+  ariaLabeledBy: PropTypes.string,
+  /**
+   * Custom `class name` to be added to the component-root-node
+   */
+  className: PropTypes.string,
+  /**
+   * Custom base of class names for add to component-root and internal-components nodes
+   *  - check Dev Tools for found available Custom/Consumer Classes
+   */
+  classNameBase: PropTypes.string,
```
hadasfa, line 107: as we talk we will remove this because we want to convert all our components to css modules.

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,242 @@
...
+  /**
+   * Color Mode (primary/positive/negative) of the component (Slider)
+   */
+  color: PropTypes.oneOf(Object.values(Slider.colors)),
+  /**
+   * Unique TestId - can be used as Selector for integration tests and other needs (tracking, etc)
+   */
+  "data-testid": PropTypes.string,
```
hadasfa, line 115: let use the same capital and small letters syntax as in the other props: dataTestId

src/components/Slider/Slider.jsx
hadasfa: let's keep this prop but add more clear comment about what is purpose is?

src/components/Slider/Slider.jsx
niksa-monday: Fixed according to comment. 
(https://github.com/mondaycom/monday-ui-react-core/pull/424/commits/1031367619056ae9a182bb88a707aaeefd3c3f04)

src/components/Slider/Slider.jsx
niksa-monday: Fixed according to comment. 
(https://github.com/mondaycom/monday-ui-react-core/pull/424/commits/1031367619056ae9a182bb88a707aaeefd3c3f04)

src/components/Slider/Slider.jsx
niksa-monday: For a while I've moved BEM helper inside component itself (from commons reusable helpers).
(https://github.com/mondaycom/monday-ui-react-core/pull/424/commits/441ef12a7478e650c4cf431724c6b8ad3d6acc9a)
When I'll finished with other sub-tasks/issues I'll start work on switch to CSS Modules.

src/components/Slider/SliderBase/SliderBase.jsx
niksa-monday: Thanks

src/components/Slider/Slider.jsx
niksa-monday: I'm changing description.
step it pretty common property for sliders:
https://mui.com/api/slider/
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role

src/components/Slider/SliderBase/SliderBase.scss
niksa-monday: For border radius we have 4 different patterns of usage:
Screen Shot 2022-01-05 at 13 58 04
Probably hardcoded and SASS vars are deprecated.
But here I have a couple of questions:
1. What to use direct Custom CSS property: `var(--border-radius-small)` or helper `getCSSVar("border-radius-small"),`
2. In Figma requirements we have 2px border-radius - but at  [UI Styles](https://github.com/mondaycom/monday-ui-style/blob/master/src/border-radius.scss) we only have 4-8-16 - probably we need to add extra-small options

Waiting you answer and I'll continue with this issue :)

src/components/Slider/SliderHelpers.js
niksa-monday: This approach is more performant and flexible - it pure function - easy for reuse, easy for change easy for tests.
We already have all needed variables from hooks - so I thought that simple helper is better for use.

src/components/Slider/SliderInfix.jsx
niksa-monday: I don't think.
I personally don't like mixes of if's and ternary operator.
I prefer early return than if - else construction.
Of course each approach have cons and pros:
- https://medium.com/swlh/return-early-pattern-3d18a41bba8
- https://betterprogramming.pub/are-early-returns-any-good-eed4b4d03866
But I really think early return much more better :)

src/components/Slider/SliderInfix.jsx
niksa-monday: As at previous topics we have here pros and cons of each approach.
- currently we just using it in same file, it pretty clear how it works
- hook itself much more readable and short
- useState use same pattern - in future in case of reuse it provide easy ability to provide names according to they needs.

In any case I don't think it critical. I can change if if you prefer.

src/components/Slider/SliderInfix.jsx
niksa-monday: You prefer to move it to SliderHooks or at separate SliderInfixHooks?

src/components/Slider/__stories__/Slider.stories.mdx
niksa-monday: Yes we are working on it.
Part of things already done.
Issue can be resolved.

src/components/Slider/__stories__/Slider.stories.mdx
niksa-monday: Evgeniy working on it.

src/helpers/bem-helper.js
niksa-monday: For a while it moved locally to slider.
I'll remove it during switch from BEM to CSS Modules (see one of previous comments)

_Issue can be resolved._

src/components/Slider/Slider.jsx
niksa-monday: We can change `isRange` to `ranged` (by analogy with `disabled`)

First of all `value` is pretty rare use case for controlled version of slider.
So probably we are speaking about defaultValue

But an Idea was  exactly wise versa - we just providing `ranged` props and getting defaultValue=[min, max] or  (exactly as min for non-ranged)

In other case we suddenly have to provide pretty complex explanation why defaultValue switching mode.
And users need to provide defaultValue=[min, max] - even id they didn't need it. 

ranged - looks more self explaining and simple for most use-cases.

src/components/Slider/Slider.jsx
niksa-monday: Done!

src/components/Slider/Slider.jsx
niksa-monday: Done!

src/components/Slider/Slider.jsx
niksa-monday: Done!

src/components/Slider/Slider.jsx
niksa-monday: Done!

src/components/Slider/SliderCommons.js
niksa-monday: Done!

src/components/Slider/SliderHooks.js
niksa-monday: Done!

src/constants/sizes.js
niksa-monday: Done!

src/components/Slider/SliderBase/SliderRail.jsx
niksa-monday: This name is not actual standard. But pretty wildly used in different realisations.
More then that it not only for Click it organize `Rail` for Track and FilledTrack

src/components/Slider/SliderBase/SliderBase.scss
hadasfa: regarding the first question - the most updated best practice / guideline we recommend is to use the var directly: var(--border-radius-small)

regarding the second question - if Evigny thinks we need to add the extra small unit so let's add it :)

src/components/Slider/Slider.jsx
niksa-monday: We are using data-testid in such a way in all other components.
Screen Shot 2022-01-05 at 15 07 01

And it have good reasons for it:

```
<MyComponent dataTestId="my-comp">
  <div data-testid="my-comp-child1" />
  <SubComponent dataTestId="my-comp-child2" />
</MyComponent>
```
compare to:
```
<MyComponent data-testid="my-comp">
  <div data-testid="my-comp-child1" />
  <SubComponent data-testid="my-comp-child2" />
</MyComponent>
```

Also sometimes you need to enhance children components with props from parent component and not always you know type of child dom-node or react-component

src/components/Slider/SliderInfix.jsx
hadasfa: all the hooks can be in one SliderHooks file

src/components/Slider/Slider.jsx
hadasfa: let's change the prop name to ranged

src/components/Slider/Slider.jsx
hadasfa: so let's keep the prop name like this :)

src/components/Slider/SliderInfix.jsx
hadasfa: So I would prefer to change it to object structure in that case, still think it is much more readable and easier to understand what is each value in the returning object means.

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,231 @@
...
+      // ------ Additional subcomponents' props
+      indicateSelection,
+      prefix,
+      postfix
+    },
+    ref
+  ) => {
+    const componentRef = useRef(null);
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    const providerProps = {
```
orrgottlieb, line 41: @niksa-monday IMO it is less clear creating an object and then spreading it on the Element.

I think it is clearer then just adding it to the component

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,231 @@
...
+      onChange,
+      ranged,
+      showValue,
+      size,
+      step,
+      value,
+      defaultValue: ensureDefaultValue({ defaultValue, min, max, ranged }),
+      valueFormatter,
+      valueText,
+      infixOptions: {
```
orrgottlieb, line 58: creating a new object reference will trigger rerender of the Infix.Provider every render

src/components/Slider/SliderBase/SliderBase.jsx
```diff
@@ -0,0 +1,107 @@
...
+  isArrowLeftEvent,
+  isArrowRightEvent,
+  isArrowUpEvent,
+  isPageDownEvent,
+  isPageUpEvent,
+  isEndEvent,
+  isHomeEvent
+} from "../../../utils/dom-event-utils";
+
+function getKey(index) {
```
orrgottlieb, line 22: why ? it is used once  and it is the identity function

src/components/Slider/SliderBase/SliderBase.jsx
```diff
@@ -0,0 +1,107 @@
...
+  return index;
+}
+
+const SliderBase = forwardRef(({ className }, ref) => {
+  const { color, disabled, dragging, size, shapeTestId } = useSliderUi();
```
orrgottlieb, line 27: Love the structure  and readability- 😍

src/components/Slider/SliderBase/SliderBase.jsx
```diff
@@ -0,0 +1,107 @@
...
+  function handlePointerMove(e) {
+    if (dragging === null) {
+      return;
+    }
+    const offsetInPx = Math.round(e.clientX - railCoords.left);
+    const newValue = moveToPx({ offsetInPx, min, max, railCoords, step });
+    changeValue(newValue);
+  }
+
+  function handleRailClick(e) {
```
orrgottlieb, line 42: as this specific function is being passed to a child component - lets wrap it with useCallback - or lets open this issue up for discussion in the group ?

src/components/Slider/SliderBase/SliderFilledTrack.jsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { bem } from "../SliderHelpers";
+
+function defineFilledTrackProps({ dimension, offset, reverse }) {
```
orrgottlieb, line 5: if this is used just for style maybe reconsider "flattening" the object a bit and just return the style properties ?

src/components/Slider/SliderBase/SliderRail.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { forwardRef } from "react";
+import PropTypes from "prop-types";
+import { bem } from "../SliderHelpers";
+
+const SliderRail = forwardRef(({ className, children, onClick }, ref) => {
+  function handleClick(e) {
+    if (typeof onClick === "function") {
```
orrgottlieb, line 7: just a question what do you say about adding this prop as required and supply a default noop - and remove the typeof ?

src/components/Slider/SliderBase/SliderRail.jsx
```diff
@@ -0,0 +1,36 @@
...
+  className: PropTypes.string,
+  /**
+   * onClick callback function
+   */
+  onClick: PropTypes.func
+};
+
+SliderRail.defaultProps = {
+  className: "",
+  onClick: undefined
```
orrgottlieb, line 33: NOOP ?

src/components/Slider/SliderBase/SliderThumb.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <Tooltip
+      open={active === index || dragging === index}
+      content={showValue ? null : valueText}
+      position={tooltipPosition}
+      showDelay={tooltipShowDelay}
+    >
+      <div
+        aria-label={ariaLabel}
```
orrgottlieb, line 56: Beautiful !!!!

src/components/Slider/SliderHooks.js
```diff
@@ -0,0 +1,83 @@
...
+    return ensureSingleValueText(valueText, value, formatter);
+  }
+  return value.map((valueSingle, index) => {
+    const valueTextSingle = Array.isArray(valueText) ? valueText[index] : undefined;
+    return ensureSingleValueText(valueTextSingle, valueSingle, formatter);
+  });
+}
+
+export function useIsStateControlledFromOutside(value) {
+  const [isControlled] = useState(typeof value !== "undefined");
```
orrgottlieb, line 27: if this can't be changed, why not just use useRef ?

src/components/Slider/SliderInfix.jsx
orrgottlieb: if the hooks are only for a specific component - IMO they could be in a file SliderInfixHooks

src/constants/sizes.js
```diff
@@ -1,11 +1,15 @@
-export const SIZES = Object.freeze({
-  XXS: "xxs",
-  XS: "xs",
+export const SIZES_BASIC = Object.freeze({
   SMALL: "small",
   MEDIUM: "medium",
   LARGE: "large"
 });
 
+export const SIZES = Object.freeze({
```
orrgottlieb, line 7: love it!

orrgottlieb (commented):
@niksa-monday amazing job!!!

very detailed!
i have a few comments some of them could be discussed in a team discussion as they are probably guidelines that we should write

src/components/Slider/Slider.jsx
```diff
@@ -0,0 +1,225 @@
...
+      indicateSelection,
+      prefix,
+      postfix
+    },
+    ref
+  ) => {
+    const componentRef = useRef(null);
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    const consumerBem = createBemBlockHelper(classNameBase, { isConsume: true });
+    function handleChange(newValue) {
```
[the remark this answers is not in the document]
niksa-monday: Moved and wraped

src/components/Slider/SliderHooks.js
```diff
@@ -0,0 +1,86 @@
...
+// TODO: can be used as global common/shared util-hooks
+export function useSliderResize(onResize) {
+  function handleResize() {
+    if (typeof onResize === "function") {
+      onResize();
+    }
+  }
+  useEffect(() => {
+    // TODO: enhance by ResizeObserve
+    window.addEventListener("resize", handleResize);
```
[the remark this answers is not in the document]
niksa-monday: Checking ...

src/components/Slider/SliderHooks.js
```diff
@@ -0,0 +1,86 @@
...
+    return () => window.removeEventListener("resize", handleResize);
+  }, [onResize]);
+}
+
+export function useSliderValue({ isControlled, value, valueDefault }) {
+  const initialValue = isControlled ? value : valueDefault;
+  const [internalStateValue, setInternalStateValue] = useState(initialValue);
+  if (isControlled) {
+    return [value, () => {}];
```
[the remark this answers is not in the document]
niksa-monday: Changing ...

src/components/Slider/Slider.jsx
niksa-monday: Changing ...

src/components/Slider/SliderBase/SliderRail.jsx
niksa-monday: Changing...

src/components/Slider/SliderHooks.js
```diff
@@ -0,0 +1,86 @@
...
+
+// TODO: can be used as global common/shared util-hooks
+export function useSliderResize(onResize) {
+  function handleResize() {
+    if (typeof onResize === "function") {
+      onResize();
+    }
+  }
+  useEffect(() => {
+    // TODO: enhance by ResizeObserve
```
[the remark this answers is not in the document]
niksa-monday: Checking ...

src/components/Slider/SliderHooks.js
niksa-monday: Technically it can be changed from consumer - not recommended, and probably not real use case.
I need to think about it.

src/components/Slider/SliderBase/SliderRail.jsx
niksa-monday: I always thought that technically it not type checking in run-time - I still can pass (as a consumer of component) not a function by mistake - and get an error.
But maybe it is ok - if consumer passing not a function - maybe we need to through error and not to take care on it silently.
Probably I'll change it ...

src/components/Slider/SliderBase/SliderBase.jsx
niksa-monday: Done!

src/components/Slider/Slider.jsx
niksa-monday: Looks like missed it after some changes/refactoring 🤦🏻‍♂️ 😆
Good catch!
Changing....

src/components/Slider/SliderBase/SliderBase.jsx
niksa-monday: Looks like missed it after some changes/refactoring 🤦🏻‍♂️ 😆
Good catch!
Changing....

src/components/Slider/Slider.jsx
niksa-monday: Missed it...
Changing ...

src/components/Slider/SliderHooks.js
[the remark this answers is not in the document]
niksa-monday: Done!

src/components/Slider/Slider.jsx
niksa-monday: Done!

src/components/Slider/Slider.jsx
niksa-monday: Done!

src/components/Slider/SliderBase/SliderBase.jsx
niksa-monday: Done!

src/components/Slider/SliderBase/SliderRail.jsx
niksa-monday: Done!

src/components/Slider/SliderBase/SliderRail.jsx
niksa-monday: Done

src/components/Slider/Slider.jsx
niksa-monday: According to our agreements move it to Backlog: Slider Improvements tasks

src/components/Slider/SliderHooks.js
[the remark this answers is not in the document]
niksa-monday: Done!

src/components/Slider/SliderHooks.js
niksa-monday: According to our agreements moved to Backlog: Slider Improvements tasks

src/components/Slider/__tests__/slider-ranged-tests.jest.js
```diff
@@ -0,0 +1,260 @@
+import React from "react";
+import { act, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { snapshotDiff } from "../../../../jest/utils";
+import { renderSliderInRangeMode } from "./slider-tests.utils";
+
+jest.useFakeTimers();
+
+jest.mock("../../TextField/TextField", () => {
```
laviomri, line 9: If I understand correctly you mock the implementation so it won't appear in the diff output. 
This means that each time that we'll want to add a new snapshot test, we'll need to replace the internal components like you did here. Is there a way to reduce the code duplication for these cases? Perhaps a helper function that gets a `path` parameter and calls `jest.mock` with a similar implementation like you have here?

niksa-monday: I mock other components - because I don't want to update Snapshots if there are changes at TextField / Icon component. It not at Slider component responsibility.
And yes, we can think about reuse of components mocks according to your suggestion.

src/components/Slider/SliderBase/SliderThumb.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+const SliderThumb = ({ className, index, onMove, position }) => {
+  const { max, min, ranged, value: valueOrValues, valueText: valueOrValuesText } = useSliderSelection();
+  const value = ranged ? valueOrValues[index] : valueOrValues;
+  const valueText = ranged ? valueOrValuesText[index] : valueOrValuesText;
+  const { active, ariaLabel, ariaLabelledby, disabled, dragging, focused, shapeTestId, showValue } = useSliderUi();
+  const { setActive, setFocused, setDragging } = useSliderActions();
+  const ref = useRef(null);
+
+  function handleBlur() {
```
hadasfa, line 20: I think that according to our new conventions from the meeting yesterday the functions should be defined outside the component. Am I right?

src/components/Slider/SliderBase/SliderThumb.jsx
```diff
@@ -0,0 +1,115 @@
+import React, { useEffect, useRef } from "react";
+import PropTypes from "prop-types";
+import { DialogPositions } from "../../../constants";
+import { NOOP } from "../../../utils/function-utils";
+import Tooltip from "../../Tooltip/Tooltip";
+import { bem } from "../SliderHelpers";
+import { useSliderActions, useSliderSelection, useSliderUi } from "../SliderContext";
+
+const tooltipShowDelay = 300;
```
hadasfa, line 9: Maybe move it to the slider constants?

src/components/Slider/SliderBase/SliderThumb.jsx
niksa-monday: No, it not pure function it uses components functions: setFocused and setActive

src/components/Slider/SliderBase/SliderThumb.jsx
niksa-monday: Done!

PR #431, opened 2022-01-04
Infra/orr/update deps

orrgottlieb:
Upgrade most of our dev and prod dependencies

Major upgrades:

- node to version 16
- webpack -> latest
- babel -> latest
- sass-loader -> latest
- style-loader -> latest


src/components/EditableHeading/EditableHeading.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/destructuring-assignment */
```
sahariko, line 1: Why not fix it and be done with it?

sahariko (approved):

src/components/EditableHeading/EditableHeading.jsx
orrgottlieb: @sahariko this component needs too be refactored IMO

sahariko: Alright, it's not a blocker or anything.

PR #445, opened 2022-01-10
Storybook actions

laviomri:
Added [Storybook actions](https://storybook.js.org/docs/react/essentials/actions), which can be easily configured to show callback examples.
I added it only to two components for now, just as a test case.
image

I found it useful for my current task, but please let me know if you think it shouldn't be used.


orrgottlieb (approved):
Very cool

PR #451, opened 2022-01-12
Added controlled input decorator

laviomri:
Added a convenient decorator which allows adding more interactivity to controlled component. 

[Before](https://user-images.githubusercontent.com/96776835/149083361-3643082d-c318-4f93-ac43-c683201bc905.mov)  ->  [After](https://user-images.githubusercontent.com/96776835/149083425-7e50d0ca-f467-4db1-ac14-6b6bda5056c5.mov)

I found it difficult to understand the behavior of some components without actually seeing a proper "flow". 
I think this decorator will be useful in more cases. However, if you think it's open for abuse I'm ok with not adding a generic decorator, and I'll add a specific decorator for my component.


laviomri:
closed in favor of https://github.com/mondaycom/monday-ui-react-core/pull/458

PR #455, opened 2022-01-12
Fix/eslint

orrgottlieb:


.eslintrc.js
```diff
@@ -1,13 +1,32 @@
...
+module.exports = {
+  globals: {
+    jest: "writeable",
+    describe: "readable",
+    it: "readable",
+    expect: "writeable",
+    afterEach: "writeable",
+    beforeEach: "writeable",
+    beforeAll: "readable",
+    afterAll: "readable"
```
sahariko, line 10: These globals are relevant only for test files, and can be achieved more easily with:
```js
{
  overrides: [
    {
      files: [testFilePattern],
      environment: {
        jest: true
      }
    }
  ]
}
```

.eslintrc.js
```diff
@@ -1,13 +1,32 @@
...
+  },
+  extends: ["plugin:react/recommended", "airbnb", "plugin:react-hooks/recommended"],
+  parserOptions: {
+    ecmaFeatures: {
+      jsx: true
+    },
+    ecmaVersion: 13,
+    sourceType: "module"
+  },
+  plugins: ["import", "react", "jsx-a11y", "json", "markdown"],
```
sahariko, line 24: I think we still want the `react-hooks` plugin.

orrgottlieb: @sahariko 
image

sahariko: Aaaah I see, since we're not extending anything, it's not needed.

sahariko (approved):

PR #458, opened 2022-01-15
Added optional mapping from action to input props

laviomri:
* As discussed following to [this PR](https://github.com/mondaycom/monday-ui-react-core/pull/451), I've enhanced the existing `createStoryMetaSettings` instead of adding a new function. 
* `createStoryMetaSettings` will now return an object containing `argTypes` and `decorators`. I've aligned the existing usage of all stories.
* As mentioned in the original PR, the new functionality allows adding more interactivity to the stories. I've added the `ColorPicker` story as an example (until now, it was impossible for the user to understand the behavior of the `isMultiselect` option, since changes didn't update the state of the component)

- [x] Optional - we can update the plop storybook (`plop/general/component-stories-mdx.txt`) to include a call to `createStoryMetaSettings`. @hadasfa WDYT?

### DIFF
--- a/plop/general/component-stories-mdx.txt
+++ b/plop/general/component-stories-mdx.txt
@@ -1,11 +1,20 @@
 import {{properCase componentName}} from "../{{properCase componentName}}";
 import { ArgsTable, Story, Canvas, Meta } from "@storybook/addon-docs";
-import { createComponentTemplate } from "../../../storybook/functions/create-component-story";
+import { createComponentTemplate, createStoryMetaSettings } from "../../../storybook/functions/create-component-story";
 import "./{{properCase componentName}}.stories.scss";
+
+export const metaSettings = createStoryMetaSettings({
+  component: {{properCase componentName}},
+  enumPropNamesArray: [], // List enum props here
+  iconPropNamesArray: [], // List props that are typed as icons here
+  actionPropsArray: [], // List the component's actions here
+});
 
 <Meta
   title="Components/{{properCase componentName}}/{{properCase componentName}}"
   component={ {{properCase componentName}} }
+  argTypes={ metaSettings.argTypes }
+  decorators={ metaSettings.decorators }
 />
 
 <!--- Component template -->
--- a/src/components/ColorPicker/__stories__/ColorPicker.stories.mdx
+++ b/src/components/ColorPicker/__stories__/ColorPicker.stories.mdx
@@ -10,7 +10,11 @@ export const metaSettings = createStoryMetaSettings({
   actionPropsArray: ["onSave"]
 });
 
-<Meta title="Pickers/ColorPicker" component={ColorPicker} argType={ metaSettings.argTypes }/>
+<Meta 
+  title="Pickers/ColorPicker"
+  component={ ColorPicker }
+  argType={ metaSettings.argTypes }
+/>
 
 <!--- Component template -->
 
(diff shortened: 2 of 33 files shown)


src/components/ColorPicker/__stories__/ColorPicker.stories.mdx
```diff
@@ -4,13 +4,18 @@ import { createComponentTemplate, createStoryMetaSettings } from "../../../story
 import TextColorIndicator from "../../Icon/Icons/components/TextColorIndicator";
 import Check from "../../Icon/Icons/components/Check";
 
-export const argTypes = createStoryMetaSettings({
+export const metaSettings = createStoryMetaSettings({
   component: ColorPicker,
   enumPropNamesArray: ["colorStyle", "colorSize"],
-  actionPropsArray: ["onSave"]
+  actionPropsArray: [{name: 'onSave', linkToValueProp: 'value'}]
```
hadasfa, line 10: I think linter should warn about using different " / ' styles in the same page

src/storybook/functions/create-component-story.js
```diff
@@ -41,9 +44,43 @@ export function createStoryMetaSettings({ component, enumPropNamesArray, iconPro
...
+/**
+ * Creates a decorator which maps a callback prop to an input prop.
+ * For example: mapping the onChange callback of a Select component to its currentValue prop.
+ * Useful for adding interactivity to stories of controlled components.
+ * Additionally, the callback will trigger a Storybook action, that can be seen on the Actions tab.
+ * @param {string} actionName - the name of the action prop of the component in the story. For example, "setValue" or "onChange".
+ * @param {string} linkToPropName - the name of the prop which should be updated when the prop of "actionName" is called. For example, "value".
+ * @returns A decorate for storybook which updates the {@link linkToPropName} input of the component, whenever {@link actionName} is called.
+ */
+function createMappedActionToInputPropDecorator(actionName, linkToPropName) {
```
hadasfa, line 68: the only thing I would change is the name of the second parameter to be the same as in the createStoryMetaSettings structure. In other words, I would change "linkToPropName" to "linkToValueProp"

and maybe change the param name to: "linkedToPropValue" ?

hadasfa (commented):
Great PR, wrote 2 small comments.

Regarding the idea to update the plop to include the createStoryMetaSettings, I think this is a great idea :)

src/components/ColorPicker/__stories__/ColorPicker.stories.mdx
laviomri: by looking at `eslintrc.js` it looks like the rule is disabled
image
do you think we should change it?

regardless, I'll fix this file :)

src/storybook/functions/create-component-story.js
laviomri: sure, it makes sense. I'll do it now :)

plop/general/component-stories-mdx.txt
```diff
@@ -1,11 +1,20 @@
 import {{properCase componentName}} from "../{{properCase componentName}}";
 import { ArgsTable, Story, Canvas, Meta } from "@storybook/addon-docs";
-import { createComponentTemplate } from "../../../storybook/functions/create-component-story";
+import { createComponentTemplate, createStoryMetaSettings } from "../../../storybook/functions/create-component-story";
 import "./{{properCase componentName}}.stories.scss";
 
+export const metaSettings = createStoryMetaSettings({
```
hadasfa, line 6: yay 🥳

hadasfa (approved):

PR #460, opened 2022-01-17
Fix tabs rendering non active tabs

MosheZemah:
Add a prop to allow rendering only the active tabs and not all of them

### DIFF
--- a/src/components/Tabs/TabPanel/TabPanel.jsx
+++ b/src/components/Tabs/TabPanel/TabPanel.jsx
@@ -4,12 +4,12 @@ import cx from "classnames";
 import useMergeRefs from "../../../hooks/useMergeRefs";
 import "./TabPanel.scss";
 
-const TabPanel = forwardRef(({ className, id, children }, ref) => {
+const TabPanel = forwardRef(({ className, id, children, key }, ref) => {
   const componentRef = useRef(null);
   const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
 
   return (
-    <div key={id} ref={mergedRef} className={cx("tab-panel--wrapper", className)} id={id} role="tabpanel">
+    <div key={key} ref={mergedRef} className={cx("tab-panel--wrapper", className)} id={id} role="tabpanel">
       {children}
     </div>
   );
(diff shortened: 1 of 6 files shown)


src/components/Tabs/TabPanel/TabPanel.jsx
```diff
@@ -4,12 +4,12 @@ import cx from "classnames";
...
 import "./TabPanel.scss";
 
-const TabPanel = forwardRef(({ className, id, children }, ref) => {
+const TabPanel = forwardRef(({ className, id, children, index }, ref) => {
   const componentRef = useRef(null);
   const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
 
   return (
-    <div ref={mergedRef} className={cx("tab-panel--wrapper", className)} id={id} role="tabpanel">
+    <div key={id || index} ref={mergedRef} className={cx("tab-panel--wrapper", className)} id={id} role="tabpanel">
```
hadasfa, line 12: this could break old tabs implementations if someone pass numeric id for only some of the tabs (few tabs will have the same numeric key and it can cause weird behivior) 
what about a combine  key with the index and the id?

src/components/Tabs/TabPanels/TabPanels.jsx
```diff
@@ -1,27 +1,34 @@
...
-  return (
-    <div ref={mergedRef} className={cx("tab-panels--wrapper", className)} id={id}>
-      {React.Children.map(children, (child, index) => {
+const TabPanels = forwardRef(
+  ({ className, id, activeTabId, animationDirection, children, renderOnlyActiveTab }, ref) => {
+    const componentRef = useRef(null);
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    const renderedTabs = useMemo(() => {
+      return React.Children.map(children, (child, index) => {
         if (activeTabId === index) {
```
hadasfa, line 13: we can do here  if (activeTabId === index || ! renderOnlyActiveTab
for reduce copies of the same code (the return line) and therefore it will be more simple.

src/components/Tabs/TabPanel/TabPanel.jsx
orrgottlieb: @hadasfa numeric id's are very bad practice and can break some browser api's - 

> Note: Technically, in HTML5, the value for an id attribute may contain any character, except whitespace characters. However, to avoid inadvertent errors, only ASCII letters, digits, '_', and '-' should be used and the value for an id attribute should start with a letter. For example, . has a special meaning in CSS (it acts as a class selector). Unless you are careful to escape it in the CSS, it won't be recognised as part of the value of an id attribute. It is easy to forget to do this, resulting in bugs in your code that could be hard to detect.

so i wouldn't worry about it.

orrgottlieb (approved):

PR #462, opened 2022-01-19
Added GridKeyboardNavigationContext

laviomri:
[Task](https://monday.monday.com/boards/470024321/pulses/2123616392)
Added a context which allows navigation between multiple navigable sections. 
A GIF is worth a thousand words :)
navigation_context

In some cases where the navigated components are not in the same dimensions, the "jump" between the two components may look clanky (as seen in the gif). It's very difficult to fix, but probably possible.

#### Basic
- [x] PR has description.
#### Style
- [ ] ~Styles are added to `NewComponent.modules.scss` file inside of the `NewComponent` folder.~  Irrelevant
- [ ] ~Component uses CSS Modules.~  Irrelevant
#### Storybook
#### Tests


src/components/GridKeyboardNavigation/__stories__/useGridKeyboardNavigationContext.stories.mdx
```diff
@@ -0,0 +1,109 @@
...
+  <Story name="Overview">{
+    () => {
+      const wrapperRef = useRef(null);
+      const leftElRef = useRef(null);
+      const rightElRef = useRef(null);
+      const keyboardContext = useGridKeyboardNavigationContext([{leftElement: leftElRef, rightElement: rightElRef}], wrapperRef)
+      return (
+        <GridKeyboardNavigationContext.Provider value={keyboardContext}>
+          <Flex ref={wrapperRef}>
+            <DummyNavigableGrid ref={leftElRef} itemsCount={15} numberOfItemsInLine={4} itemPrefix="L "/>
```
laviomri, line 43: The story doesn't look too good, but I wanted to emphasize the layout structure so the behavior will be clear. (you can check the gif in the PR's description).
Do you think I should change it?

src/components/GridKeyboardNavigationContext/GridKeyboardNavigationContextHelper.js
```diff
@@ -0,0 +1,51 @@
+import { NAV_DIRECTIONS } from "../../hooks/useFullKeyboardListeners";
+
+export const getDirectionMaps = positions => {
+  const directionMaps = {
+    [NAV_DIRECTIONS.RIGHT]: new Map(),
+    [NAV_DIRECTIONS.LEFT]: new Map(),
+    [NAV_DIRECTIONS.UP]: new Map(),
+    [NAV_DIRECTIONS.DOWN]: new Map()
```
MosheZemah, line 8: why not simple {} ?

MosheZemah (approved):

src/components/GridKeyboardNavigationContext/GridKeyboardNavigationContextHelper.js
laviomri: since I use Objects (refs) as the keys. If I'll use a regular object as a map, the actual value which will be used as key is the result of `.toString()`:
image

on the contrary, [a `Map` can handle objects as keys](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map#objects_vs._maps)

src/components/GridKeyboardNavigationContext/GridKeyboardNavigationContextHelper.js
```diff
@@ -0,0 +1,51 @@
+import { NAV_DIRECTIONS } from "../../hooks/useFullKeyboardListeners";
```
sahariko, line 1: [PETTY] This file can be simply called `helpers.js`.

src/components/GridKeyboardNavigationContext/GridKeyboardNavigationContextHelper.js
```diff
@@ -0,0 +1,51 @@
...
+    // no relations were registered for this direction - fallback to a different direction
+    if ([NAV_DIRECTIONS.LEFT, NAV_DIRECTIONS.RIGHT].includes(direction)) {
+      // there are no registered horizontal relations registered, try vertical relations. Get the top-most element.
+      return getOutmostElementInDirection(directionMaps, NAV_DIRECTIONS.UP);
+    }
+    // there are no registered vertical relations registered, try horizontal relations. Get the left-most element.
+    return getOutmostElementInDirection(directionMaps, NAV_DIRECTIONS.LEFT);
+  }
+  let result = firstEntry?.[0];
+  while (directionMap.get(result)) {
```
sahariko, line 46: Just a thought - is there a way for someone to register elements in a manner that will create a circular loop?

src/components/GridKeyboardNavigationContext/GridKeyboardNavigationContextHelper.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+  });
+  return directionMaps;
+};
+
+export const getOppositeDirection = direction => {
+  if (direction === NAV_DIRECTIONS.LEFT) return NAV_DIRECTIONS.RIGHT;
+  if (direction === NAV_DIRECTIONS.RIGHT) return NAV_DIRECTIONS.LEFT;
+  if (direction === NAV_DIRECTIONS.UP) return NAV_DIRECTIONS.DOWN;
+  if (direction === NAV_DIRECTIONS.DOWN) return NAV_DIRECTIONS.UP;
```
sahariko, line 27: We should probably throw an explicit error if passed a non-supported direction, otherwise there will be no return value, and debugging would be harder.

Also [PETTY] Can be a `switch` case.

src/components/GridKeyboardNavigationContext/GridKeyboardNavigationContextHelper.js
laviomri: yup, it's possible. and it will cause an infinite loop in some edge cases. I'll add a guard and throw an error in that case 👍

src/components/GridKeyboardNavigationContext/GridKeyboardNavigationContextHelper.js
laviomri: done 👍

PR #463, opened 2022-01-19
AttentionBox with icon on compact mode support & Docs fixes in attention box and typography

hadasfa:
- AttentionBox with icon on compact mode support 
- Fix spacing between attention box header and content
- Fix attention box story ad the assets in typography story


.storybook/preview.js
```diff
@@ -74,11 +74,7 @@ addParameters({
         "*",
         "Accessibility",
         "Hooks"
-      ],
-      includeName: false,
```
hadasfa, line 78: that's not work

src/components/AttentionBox/AttentionBox.jsx
```diff
@@ -13,6 +13,8 @@ const AttentionBox = ({
   className,
   // Backward compatibility for props naming
   componentClassName,
+  // Will remove when releasing version 2 as BREAKING CHANGES
+  withIconWithoutHeader,
```
hadasfa, line 17: because until now we did not support displaying icon without title.
we still don't support it in regular mode but in compact mode it is looking good and needed. (example in screenshot in the pr)

hadasfa: that is a breaking change and that's why we add the flag

src/components/AttentionBox/AttentionBox.scss
```diff
@@ -62,11 +66,9 @@
     display: flex;
     align-items: center;
     margin-top: 0;
-    margin-bottom: $spacing-medium;
+    margin-bottom: var(--spacing-small);
```
hadasfa, line 69: evigany request

package.json
```diff
@@ -82,19 +82,19 @@
...
-    "@storybook/addons": "6.5.0-alpha.5",
-    "@storybook/builder-webpack5": "6.5.0-alpha.5",
-    "@storybook/manager-webpack5": "6.5.0-alpha.5",
+    "@storybook/addon-a11y": "6.5.0-alpha.18",
+    "@storybook/addon-actions": "6.5.0-alpha.18",
+    "@storybook/addon-docs": "6.5.0-alpha.18",
+    "@storybook/addon-links": "6.5.0-alpha.18",
+    "@storybook/addon-storyshots": "6.5.0-alpha.18",
+    "@storybook/addons": "6.5.0-alpha.18",
+    "@storybook/builder-webpack5": "6.5.0-alpha.18",
```
hadasfa, line 91: we want to upgrade the storybook version to the latest and check how if it fix few storybook bugs

laviomri (approved):

package.json
laviomri: do we have a list of bugs we'd like to get fixed?

hadasfa: its mainly a bug with the order of the stories - the stories displayed by alphabet instead according to their load order even we did not defined it in the preview

PR #464, opened 2022-01-20
Keyboard nav for ColorPicker

laviomri:
[Task](https://monday.monday.com/boards/470024321/pulses/2174881648)
Added keyboard navigation for the `ColorPicker` component. Also, added a circle-shape mode (to be used as a color picker for the workspace).

image

- [ ] Requires Evgeny's review


src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
```diff
@@ -19,43 +20,26 @@ const ColorPickerItemComponent = ({
...
+  isActive,
+  colorShape
 }) => {
   const colorAsStyle = ColorUtils.getMondayColorAsStyle(color, colorStyle);
   const itemRef = useRef(null);
 
   const onMouseDown = useCallback(e => e.preventDefault(), []);
   const onClick = useCallback(() => onValueChange(color), [onValueChange, color]);
 
-  useEffect(() => {
```
laviomri, line 30: this effect was used to change the color to "selected" mode when hovering. This behavior has changed - we now add a border as a hovering feedback, instead of changing the color

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.scss
```diff
@@ -1,16 +1,56 @@
+@import "../../../../styles/states.scss";
+@import "../../../../styles/global-css-settings.scss";
+
 .monday-style-color-item-wrapper {
+  --border-radius-multiplier: 1;
+
+  &.circle {
+    --border-radius-multiplier: 100;
```
sahariko, line 8: Is `100` here just to make sure the corder radius is a perfect circle? If so, and it's not critical, we might as well go with a ridiculously big number, just to make sure it doesn't break in bigger sizes.

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContentComponent.jsx
```diff
@@ -2,15 +2,102 @@ import cx from "classnames";
...
+import { COLOR_SHAPES, DEFAULT_NUMBER_OF_COLORS_IN_LINE } from "../../ColorPickerConstants";
 import { calculateColorPickerWidth } from "../../services/ColorPickerStyleService";
 import "./ColorPickerContentComponent.scss";
+import useGridKeyboardNavigation from "../../../../hooks/useGridKeyboardNavigation/useGridKeyboardNavigation";
+import {
+  GridKeyboardNavigationContext,
+  useGridKeyboardNavigationContext
+} from "../../../GridKeyboardNavigationContext/GridKeyboardNavigationContext";
+
+const Colors = React.forwardRef(
```
sahariko, line 20: Both this and `ClearButton` take up quite a bit of this file, consider separating them to dedicated files.

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContentComponent.jsx
```diff
@@ -27,12 +114,22 @@ const ColorPickerContentComponent = ({
   isMultiselect,
   colorSize,
   numberOfColorsInLine,
-  tooltipContentByColor
+  tooltipContentByColor,
+  focusOnMount,
```
sahariko, line 118: Missing in `propTypes`.

sahariko (approved):

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.scss
laviomri: sure, I'll change it to 99999 (I think it will also be clearer for devs to understand its an arbitrary number)

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContentComponent.jsx
laviomri: good catch! added 😄

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContentComponent.jsx
laviomri: good idea. I split it.
as a more general question - do we usually add PropTypes to such "internal" components which are used only in a specific component? (I haven't used PropTypes so much in the past)

sahariko: Yes, I think that's the best practice, even though not all components implement it fully.

PR #467, opened 2022-01-20
support keep open mode for dropdown

etaylib:
image

if in multi mode - keep open, otherwise, close, as before


src/components/Dropdown/Dropdown.jsx
```diff
@@ -249,6 +250,7 @@ const Dropdown = ({
       tabIndex={tabIndex}
       id={id}
       autoFocus={autoFocus}
+      closeMenuOnSelect={!keepDropdownOpenOnSelection}
```
etaylib, line 253: currently it's being closed automatically, so to prevent breaking, we pass with "true" and enable it

src/components/Dropdown/Dropdown.jsx
```diff
@@ -274,7 +276,8 @@ Dropdown.defaultProps = {
   extraStyles: defaultCustomStyles,
   tabIndex: "0",
   id: undefined,
-  autoFocus: false
+  autoFocus: false,
+  keepDropdownOpenOnSelection: false
```
sahariko, line 280: We're already in the "dropdown" context, so we can be less verbose:

```suggestion
  keepOpenOnSelection: false
```

src/components/Dropdown/Dropdown.jsx
```diff
@@ -55,7 +55,8 @@ const Dropdown = ({
   multiline = false,
   onOptionRemove: customOnOptionRemove,
   onOptionSelect,
-  onClear
+  onClear,
```
hadasfa, line 58: please fix linter error

src/components/Dropdown/Dropdown.jsx
hadasfa: Agree with sahar. also, please set default value to undefined and set the default value depend if the drop down is multi or not (multi - true, single: false) - according to evigany request.

src/components/Dropdown/Dropdown.jsx
etaylib: which one?

src/components/Dropdown/Dropdown.jsx
```diff
@@ -56,7 +56,8 @@ const Dropdown = ({
   onOptionRemove: customOnOptionRemove,
   onOptionSelect,
   onClear,
-  onInputChange
+  onInputChange,
+  closeMenuOnSelect
```
sahariko, line 60:
```suggestion
  closeMenuOnSelect = !multi
```

src/components/Dropdown/Dropdown.jsx
```diff
@@ -249,6 +250,7 @@ const Dropdown = ({
       tabIndex={tabIndex}
       id={id}
       autoFocus={autoFocus}
+      closeMenuOnSelect={closeMenuOnSelect === undefined ? !multi : closeMenuOnSelect}
```
sahariko, line 253: See comment above.

src/components/Dropdown/Dropdown.jsx
```diff
@@ -446,7 +449,8 @@ Dropdown.propTypes = {
   /**
    * If set to true together with `multi`, it will make the dropdown expand to multiple lines when new values are selected.
    */
-  multiline: PropTypes.bool
+  multiline: PropTypes.bool,
+  closeMenuOnSelect: PropTypes.bool
```
sahariko, line 453: Needs description.

hadasfa (approved):

PR #471, opened 2022-01-22
feat: fix lint and absolute paths

orrgottlieb:


MosheZemah (approved):

laviomri (approved):

.eslintrc.js
```diff
@@ -82,6 +82,7 @@ module.exports = {
     "jsx-a11y/no-noninteractive-tabindex": "error",
     "react/jsx-boolean-value": "off",
     "max-len": "off",
+    indent: "off",
```
laviomri, line 85: why do we want to turn this rule off?

PR #482, opened 2022-01-24
 Add typography variables

hadasfa:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


src/components/Avatar/AvatarContent.scss
```diff
@@ -12,9 +15,9 @@ $avatar-icon-small: 12px;
   }
 
   &_text {
-    font-family: $primary-font;
-    font-weight: $font-weight-normal;
-    line-height: $line-height-for-size-18;
+    font-family: var(--font-family);
+    font-weight: var(--font-weight-normal);
+    line-height: 21px;
```
hadasfa, line 20: we usually put different padding for 18px and therefore removed the variable

src/components/Button/Button.scss
```diff
@@ -8,7 +8,6 @@
   outline: none;
   border: none;
   height: auto;
-  font-family: Roboto, sans-serif;
```
hadasfa, line 11: because  we use mixins for fonts here for each button size which include font family setting

src/components/Avatar/AvatarContent.scss
```diff
@@ -3,6 +3,9 @@
 $avatar-icon-large: 28px;
 $avatar-icon-medium: 18px;
 $avatar-icon-small: 12px;
+$font-size-avatar-large: 18px;
```
hadasfa, line 6: moved out of typography because its not the default sizes which in css variables now

src/storybook/styles/typography.scss
```diff
@@ -2,7 +2,7 @@
 
 // font family
 $header-primary-font: "Sofia Pro", Roboto, sans-serif;
-$text-primary-font: Roboto, sans-serif;
+$text-primary-font: var(--font-family);
```
hadasfa, line 5: its ok because we don't want our storybook to be depended at the system font

src/storybook/stand-alone-documentaion/typography/text-styles/text-styles.jsx
```diff
@@ -74,9 +74,9 @@ export const TextStyles = () => {
       </VisualDescription>
       <VisualDescription
         ariaLabel="H5"
-        title="Fourth heading (Roboto 18px bold)"
+        title="Fourth heading (Roboto 16px bold)"
```
laviomri, line 77: (might be out of this PR's scope) looks like the details can easily get out of sync with the actual rules (meaning, the rules might change to be `Roboto 16px semi-bold`, but no one will remember to update this story).
Is there a way to automatically sync between the rules and the stories? perhaps [exporting a SASS variable](https://css-tricks.com/getting-javascript-to-talk-to-css-and-sass/#aa-sass-variables-and-javascript)?

src/styles/typography.scss
```diff
@@ -29,129 +9,81 @@ $p-empty-state-height: 16px;
 $h6-empty-state-height: 12px;
 $small-empty-state-height: 12px;
 
+@mixin base-font {
```
laviomri, line 12: looks like the usage of `base-font` always comes along with a `font` rule. What do you think about adding the font rule as an argument to `base-font`, so that future devs will never forget that this is required?

laviomri (approved):
Cool! had some questions :)

src/storybook/stand-alone-documentaion/typography/text-styles/text-styles.jsx
hadasfa: We exporting the css variables for this exactly. ׳e need to add them to this doc and i will talk with eivgany about the design of the doc with the variables.

src/styles/typography.scss
hadasfa: Agreed, will be fixed

PR #492, opened 2022-01-26
Menu item small refactor

laviomri:
Small refactor, no actual change - mostly prettier. I split it to avoid big PRs.

### DIFF
--- a/src/components/Menu/Menu/hooks/useMenuKeyboardNavigation.jsx
+++ b/src/components/Menu/Menu/hooks/useMenuKeyboardNavigation.jsx
@@ -22,7 +22,6 @@ export default function useMenuKeyboardNavigation(
   setActiveItemIndex,
   isVisible,
   ref,
-  resetOpenSubMenuIndex,
   useDocumentEventListeners
 ) {
   const onArrowKeyEvent = useCallback(
--- a/src/components/Menu/MenuItem/MenuItem.jsx
+++ b/src/components/Menu/MenuItem/MenuItem.jsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-props-no-spreading */
-import React, { useCallback, useRef, useLayoutEffect, useMemo, useEffect } from "react";
+import React, { useCallback, useRef, useLayoutEffect, useMemo, useEffect, forwardRef } from "react";
 
 import PropTypes from "prop-types";
 import isFunction from "lodash/isFunction";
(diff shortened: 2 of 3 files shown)


src/components/Menu/Menu/hooks/useMenuKeyboardNavigation.jsx
```diff
@@ -22,7 +22,6 @@ export default function useMenuKeyboardNavigation(
   setActiveItemIndex,
   isVisible,
   ref,
-  resetOpenSubMenuIndex,
```
laviomri, line 25: it's unused

src/components/Menu/MenuItem/MenuItem.jsx
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-props-no-spreading */
-import React, { useCallback, useRef, useLayoutEffect, useMemo, useEffect } from "react";
+import React, { useCallback, useRef, useLayoutEffect, useMemo, useEffect, forwardRef } from "react";
```
laviomri, line 2: I added forwardRef, and applied the prettier fixes

src/components/Menu/Menu/hooks/useMenuKeyboardNavigation.jsx
MosheZemah, line 25: you can't remove it since it is not an object (it is in method params)
@sahariko

MosheZemah (changes requested):

src/components/Menu/Menu/hooks/useMenuKeyboardNavigation.jsx
laviomri: I'm not sure I understood you - there's only one place that references this function (in [Menu.jsx](https://github.com/mondaycom/monday-ui-react-core/pull/492/files#diff-bd738b83ed2d4d19d546bde5ecfd6ec076a1d38691922f6098b3b79bdb8931deL86)), and I removed it too in this PR.. where do you think this may still be used?

sahariko: This would be considered a breaking change, resulting in a major version bump of this package.

I think that we can create a milestone and aggregate all of these breaking changes, but for now let's hold back on this.

laviomri: I'm not 100% (I don't think we can be 100% sure). 
It is however a highly internal hook for the menu component, which isn't exposed directly from the package. If someone external imports this hook directly (e.g. `require("dist/src/components.....")`, should we avoid changing it?

I have no problem with reverting this change, but I'm pretty sure I'm missing something :)

sahariko: Hmmm good point, I think we can make this change then 🙂

MosheZemah (approved):

PR #494, opened 2022-01-26
supporting custom color picker - v1

etaylib:
https://user-images.githubusercontent.com/17691874/151213021-283da879-6d90-4816-9dda-04348546e87c.mov


src/components/ColorPicker/components/ColorPickerContent/ColorPickerColorsGrid.jsx
```diff
@@ -22,11 +22,33 @@ export const ColorPickerColorsGrid = React.forwardRef(
...
       tooltipContentByColor,
-      colorShape
+      colorShape,
+      showCustomColorPicker,
+      initialCustomColor
     },
     ref
   ) => {
+    const [customColor, setCustomColor] = useState(initialCustomColor);
+    const [showColorPicker, setShowColorPicker] = useState();
```
laviomri, line 32: consider being more explicit here
```suggestion
    const [showColorPicker, setShowColorPicker] = useState(false);
```

and same comment on lines 46-47 (`setShowColorPicker()`...)

src/components/ColorPicker/components/ColorPickerContent/ColorPickerColorsGrid.jsx
```diff
@@ -22,11 +22,33 @@ export const ColorPickerColorsGrid = React.forwardRef(
...
+    const onCustomColorClicked = useCallback(
+      colorHex => {
+        onSelectionAction(colorsToRender.length);
+        setCustomColor(colorHex);
+        onColorClicked(colorHex, customColor);
+      },
+      [colorsToRender.length, customColor, onColorClicked, onSelectionAction]
+    );
+
+    const onClearSelectedColor = useCallback(
```
laviomri, line 44:
```suggestion
    const onClearCustomColor = useCallback(
```

src/components/ColorPicker/components/ColorPickerContent/ColorPickerColorsGrid.jsx
```diff
@@ -22,11 +22,33 @@ export const ColorPickerColorsGrid = React.forwardRef(
...
+      colorToRemove => {
+        setShowColorPicker();
+        setCustomColor();
+        onColorClicked(colorToRemove);
```
laviomri, line 48: won't this trigger a color selection for the custom color that we are currently removing?

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.scss
```diff
@@ -1,101 +1,142 @@
...
-    }
+    .color-item {
+      display: flex;
+      justify-content: center;
+      align-items: center;
+      border: 2px solid transparent;
+      border-radius: calc(4px * var(--border-radius-multiplier, 1));
+      cursor: pointer;
+
+      &.rainbow {
```
laviomri, line 68: maybe it's just me, but I suggest giving it a more "mild" colorization 😄

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.scss
```diff
@@ -1,101 +1,142 @@
...
       }
     }
+
+    .color-item-text-mode:hover {
+      background-color: var(--primary-background-hover-color) !important;
+    }
+  }
+
+  .custom-color-picker {
+    transform: translateY(-8);
```
laviomri, line 126: should this be ` transform: translateY(-8px);`? 
also, if it looks ok as it is right now, consider just removing this line :)

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
```diff
@@ -55,35 +75,56 @@ const ColorPickerItemComponent = ({
...
+        </li>
+      </Tooltip>
+      {isColorPickerItem && isSelected && (
+        <div className="clear-color-picker" onClick={() => clearSelectedColor(customColor)}>
+          <Icon icon={Delete} iconType={Icon.type.SVG} ignoreFocusStyle iconSize={14} />
+        </div>
+      )}
+      {isColorPickerItem && showColorPicker && (
+        <div className="custom-color-picker">
+          <ChromePicker color={{ hex: customColor }} onChange={newColor => chooseCustomColor(newColor.hex)} />
```
laviomri, line 124: consider using useMemo for the color, since otherwise a new object will be created on each render call, causing unnecessary renders of `ChromePicker`. (using `useCallback` for the onChange will also be nice 😄 )

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
```diff
@@ -55,35 +75,56 @@ const ColorPickerItemComponent = ({
...
+            <div className="color-indicator-wrapper" style={colorIndicatorWrapperStyle}>
+              {shouldRenderIcon && (
+                <Icon icon={shouldRenderSelectedIcon ? SelectedIndicatorIcon : ColorIndicatorIcon} ignoreFocusStyle />
+              )}
+            </div>
+          </Clickable>
+        </li>
+      </Tooltip>
+      {isColorPickerItem && isSelected && (
+        <div className="clear-color-picker" onClick={() => clearSelectedColor(customColor)}>
```
laviomri, line 118: how will keyboard users be able to interact with this button?
(BTW, consider using a button and not a div, so it will be tab-able..)

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
```diff
@@ -22,13 +24,31 @@ const ColorPickerItemComponent = ({
   colorSize,
   tooltipContent,
   isActive,
-  colorShape
+  colorShape,
```
laviomri, line 27: it feels like the custom color picker might be better off as a separate component.
This way, we can end up with two "self contained" components (one for regular color and one for custom picker), instead of having one component with many "if/else".

I'm not entirely sure about this suggestion, but I think it's worth considering...

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContentComponent.jsx
```diff
@@ -51,19 +53,24 @@ const ColorPickerContentComponent = forwardRef(
     }, [isBlackListMode, colorsList]);
 
     const onColorClicked = useCallback(
-      color => {
+      (color, removeOnAdd) => {
```
laviomri, line 56: this doesn't feel right  - usually, an "onClick" callback only receives the item which was clicked. Maybe we need to rename this function? (or perhaps the variable?)

PR #495, opened 2022-01-26
fix close when click outside in drop down and display correct options when delete

hadasfa:
1. Update react select to most update 4 version for fixing issue - drop down does not display the right options after choosing all of them and then cancel one selection. 
2. Move value container to different component instead of renderer which created on run time because of its create a bug (drop down menu will not close in multi mode when click outside(
before:

https://user-images.githubusercontent.com/72390374/151222646-199c15ce-9e19-462b-9c9c-b8f5a854c88c.mov

after:

https://user-images.githubusercontent.com/72390374/151222978-31e6660f-56cb-4969-aa65-1fbdcab5f73b.mov


package.json
```diff
@@ -63,7 +63,7 @@
     "prop-types": "^15.8.0",
     "react-inlinesvg": "^2.3.0",
     "react-popper": "^2.2.5",
-    "react-select": "^3.1.0",
+    "react-select": "^4.0.0",
```
orrgottlieb, line 66: please use the correct node version, run `nvm use` before and than `npm i`

orrgottlieb (changes requested):

package.json
hadasfa, line 66: as far as i seen they have only 1 big change which not relevant to the way we implement our drop down - the change is in the on change function 1 param value: https://react-select.com/upgrade#from-v3-to-v4

orrgottlieb (changes requested):

package.json
hadasfa: I have run it but did not understand what it did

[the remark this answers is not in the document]
hadasfa: now it is not relevant anymore right?

orrgottlieb (approved):

PR #506, opened 2022-02-01
Menu grid item

laviomri:
Added `MenuGridItem`
MenuGridItem demo

#### Basic
- [x] PR has description.
#### Style
- [x] Styles are added to `NewComponent.modules.scss` file inside of the `NewComponent` folder.
#### Storybook
#### Tests


src/components/Menu/Menu/Menu.jsx
```diff
@@ -91,12 +95,10 @@ const Menu = forwardRef(
       setIsInitialSelectedState(true);
     }, [setIsInitialSelectedState]);
 
-    useLayoutEffect(() => {
+    useEffect(() => {
```
laviomri, line 98: When using `requestAnimationFrame`, long-pressing the arrows didn't work properly with the focus/blur handlers.

src/components/Menu/MenuGridItem/useMenuGridItemNavContext.jsx
```diff
@@ -0,0 +1,49 @@
...
+   * Currently, the two mechanisms work a bit differently.
+   * In the future, when the Menu component will use a GridKeyboardNavigationContext, this adapter shouldn't be needed anymore.
+   */
+  const innerKeyboardContext = useGridKeyboardNavigationContext([], wrapperRef);
+  const keyboardContext = useMemo(
+    () => ({
+      onOutboundNavigation: (elementRef, direction) => {
+        innerKeyboardContext.onOutboundNavigation(elementRef, direction);
+
+        switch (direction) {
```
orrgottlieb, line 25: no right because of sub menus ? - i guess we don't allow sub menu and a MenuItemGrid

src/components/index.js
```diff
@@ -75,3 +75,4 @@ export { default as ColorUtils } from "../utils/colors-utils";
 export { default as Slider } from "./Slider/Slider";
 export { default as IconButton } from "./IconButton/IconButton";
 export { default as Flex } from "./Flex/Flex";
+export { default as MenuGridItem } from "./Menu/MenuGridItem/MenuGridItem";
```
orrgottlieb, line 78: please add to publish components as well

orrgottlieb (approved):

src/components/Menu/MenuGridItem/useFocusGridItemByActiveStatus.jsx
```diff
@@ -0,0 +1,16 @@
+import { useMemo, useEffect } from "react";
+import { focusElementWithDirection } from "../../GridKeyboardNavigationContext/helper";
+import { useLastNavigationDirection } from "../Menu/hooks/useLastNavigationDirection";
+
+export const useFocusGridItemByActiveStatus = ({ wrapperRef, childRef, index, activeItemIndex }) => {
+  const { lastNavigationDirectionRef } = useLastNavigationDirection();
+  const isActive = useMemo(() => index === activeItemIndex, [activeItemIndex, index]);
```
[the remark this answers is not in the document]
laviomri: this is intentional - the calculations of `isActive` may run a few times without changing the result (e.g. when `index = 2` and the `activeItemIndex` changes from 3 to 4, and then to 0. `isActive` will remain `false` all the time). I want the content of the `useEffect` to run ONLY when the `isActive` actually changes, not when the calculation is performed.

src/components/Menu/MenuGridItem/MenuGridItem.jsx
```diff
@@ -0,0 +1,117 @@
...
+      setActiveItemIndex,
+      getPreviousSelectableIndex,
+      getNextSelectableIndex,
+      activeItemIndex,
+      isUnderSubMenu,
+      closeMenu
+    });
+
+    return (
+      <div
```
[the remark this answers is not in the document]
laviomri: great idea 👍

src/components/Menu/MenuGridItem/useMenuGridItemNavContext.jsx
laviomri: if we would have supported submenus inside a `MenuGridItem`, then we should have added a handler for outbound right navigation... however, I don't think such behavior makes any sense (I mean, it will be a grid that opens a sub menu when navigating to the right of it. Sounds odd, no? 🤔 ).
In any case, if we'll migrate the Menu implementation to use the grid keyboard navigation, it will be easier to support such (odd) behaviors 😄

src/components/index.js
laviomri: I'm not sure I understand the usage of each file correctly..
Also, should I add `useGridKeyboardNavigation` to the end of `published-components` as well?

src/components/Menu/Menu/Menu.jsx
laviomri: theoretically, it may cause a very very short delay until the `focus` is actually called ([since `useEffect` is async and `useLayoutEffect` is sync, basically](https://kentcdodds.com/blog/useeffect-vs-uselayouteffect)).
In practice I've played with it quite a bit and didn't see any issue with it.. Do you have specific concerns?

PR #512, opened 2022-02-02
Disabling activeIndex after a non-keyboard interaction

laviomri:
[Task](https://monday.monday.com/boards/245345663/pulses/2213226987)

(Evgeniy's request)
Previously, the keyboard's activeIndex feedback appeared even after selecting items with the mouse. This caused "noise", since mouse-users didn't actually need to see the feedback.
This change shouldn't have any functional effect.

### DIFF
--- a/src/hooks/useGridKeyboardNavigation/useGridKeyboardNavigation.js
+++ b/src/hooks/useGridKeyboardNavigation/useGridKeyboardNavigation.js
@@ -107,7 +107,7 @@ export default function useGridKeyboardNavigation({
 
   useEffect(() => {
     if (activeIndex > -1) {
-      ref?.current?.focus();
+      ref.current?.focus();
     }
   }, [activeIndex, ref]);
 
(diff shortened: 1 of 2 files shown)


src/hooks/useGridKeyboardNavigation/useGridKeyboardNavigation.js
```diff
@@ -108,15 +105,27 @@ export default function useGridKeyboardNavigation({
   useEventListener({ eventName: "focus", callback: onFocus, ref });
   useEventListener({ eventName: "blur", callback: onBlur, ref });
 
+  useEffect(() => {
+    if (activeIndex > -1) {
+      ref?.current?.focus();
```
MosheZemah, line 110: ref always exists

MosheZemah (approved):

src/hooks/useGridKeyboardNavigation/useGridKeyboardNavigation.js
laviomri: correct :) fixed

PR #513, opened 2022-02-03
Allow non-monday colors picker

laviomri:
[Task](https://monday.monday.com/boards/245345663/pulses/2123616491)
Allows non-monday colors to be rendered in the color picker. 
Why? Since sometimes we need to color picker for existing colors, that aren't part of our "official" palette.  (e.g. on the workspace color picker).
image
I want to discourage the usage of this flag, so I named it in accordingly (the word "force.." always makes you think twice 😄 ). 
Also, I intentionally didn't add a story, since I don't want it to be too familiar.

### DIFF
--- a/src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
+++ b/src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
@@ -1,8 +1,8 @@
-import React, { useRef, useCallback, useEffect } from "react";
+import React, { useRef, useCallback, useEffect, useMemo } from "react";
 import PropTypes from "prop-types";
 import cx from "classnames";
 import NOOP from "lodash/noop";
-import { COLOR_STYLES } from "../../../../general-stories/colors/colors-vars-map";
+import { COLOR_STYLES, contentColors } from "../../../../general-stories/colors/colors-vars-map";
 import ColorUtils from "../../../../utils/colors-utils";
 import "./ColorPickerItemComponent.scss";
 import Icon from "../../../Icon/Icon";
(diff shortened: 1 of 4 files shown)


src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
```diff
@@ -1,8 +1,8 @@
-import React, { useRef, useCallback, useEffect } from "react";
+import React, { useRef, useCallback, useEffect, useMemo } from "react";
 import PropTypes from "prop-types";
 import cx from "classnames";
 import NOOP from "lodash/noop";
-import { COLOR_STYLES } from "../../../../general-stories/colors/colors-vars-map";
+import { COLOR_STYLES, contentColors } from "../../../../general-stories/colors/colors-vars-map";
```
orrgottlieb, line 5: can we move these to something not related to stories ?

orrgottlieb (approved):

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
laviomri: I'm not sure I understand you correctly - right now it's not related to storybook. what are you suggesting?

hadasfa (approved):

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
hadasfa: I think it's about to move this constants to different location which not related to stories, maybe the colors util?

laviomri: oh I see it now! done

PR #514, opened 2022-02-03
fix support on checkbox click while shift is pressed on firefox browser

hadasfa:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


src/components/Checkbox/hooks/useFirefoxShiftLabelSupport.jsx
```diff
@@ -0,0 +1,16 @@
+import { useCallback } from "react";
+
+export function useFirefoxShiftLabelSupport({ ref }) {
+  const isFirefox = navigator.userAgent.indexOf("Firefox") > -1;
```
orrgottlieb, line 4: can you please install an npm for that, please memoize it

src/components/Checkbox/Checkbox.js
```diff
@@ -62,6 +64,7 @@ export const Checkbox = ({
       className={cx(BASE_CLASS_NAME, overrideClassName, { [`${BASE_CLASS_NAME}__disabled`]: disabled })}
       onMouseUp={onMouseUpCallback}
       htmlFor={id}
+      onClick={onLabelClick}
```
orrgottlieb, line 67: you can `onClickCapture` and prevent default and stop bubbling if needed

src/components/Checkbox/hooks/useFirefoxShiftLabelSupport.jsx
orrgottlieb, line 4: please add a test, fake the navigator useragent

orrgottlieb (changes requested):

src/components/Checkbox/hooks/useFirefoxShiftLabelSupport.jsx
```diff
@@ -0,0 +1,23 @@
...
+export function useFirefoxShiftLabelSupport({ inputRef }) {
+  const isFirefox = useMemo(() => {
+    const browser = detect();
+    console.log(browser);
+    return browser && browser.name && browser.name === BROWSER_NAME;
+  }, []);
+  const onClickCapture = useCallback(
+    e => {
+      if (e.shiftKey && isFirefox && inputRef?.current?.click) {
+        e.preventDefault();
```
hadasfa, line 15: stop prorogation does not block default checkbox behavior, only prevent default 
this is what I found about the subject: https://stackoverflow.com/questions/67612404/why-does-event-stoppropagation-not-prevent-a-label-from-checking-its-checkbo
in addition, i am not sure we want to block other on click events that we will want to add in the future (we will want the same behavior for firefox as other browsers no?)
so the question is if i should add here call for stop prorogation in addition to the prevent default call

src/components/Checkbox/hooks/useFirefoxShiftLabelSupport.jsx
```diff
@@ -0,0 +1,23 @@
+import { useCallback, useMemo } from "react";
+import { detect } from "detect-browser";
+
+const BROWSER_NAME = "firefox";
+
+export function useFirefoxShiftLabelSupport({ inputRef }) {
+  const isFirefox = useMemo(() => {
+    const browser = detect();
+    console.log(browser);
```
orrgottlieb, line 9: remove

src/components/Checkbox/hooks/useFirefoxShiftLabelSupport.jsx
```diff
@@ -0,0 +1,23 @@
+import { useCallback, useMemo } from "react";
+import { detect } from "detect-browser";
+
+const BROWSER_NAME = "firefox";
+
+export function useFirefoxShiftLabelSupport({ inputRef }) {
+  const isFirefox = useMemo(() => {
+    const browser = detect();
```
orrgottlieb, line 8: this can be done in the scope level - i don't think that the browser will change ?

src/components/Checkbox/hooks/useFirefoxShiftLabelSupport.jsx
orrgottlieb: cool!

src/components/Checkbox/hooks/useFirefoxShiftLabelSupport.jsx
```diff
@@ -0,0 +1,23 @@
...
+    return browser && browser.name && browser.name === BROWSER_NAME;
+  }, []);
+  const onClickCapture = useCallback(
+    e => {
+      if (e.shiftKey && isFirefox && inputRef?.current?.click) {
+        e.preventDefault();
+        inputRef.current.click();
+      }
+    },
+    [isFirefox]
```
orrgottlieb, line 19: please add missing dependencies

orrgottlieb (changes requested):

src/components/Checkbox/Checkbox.js
```diff
@@ -29,6 +30,18 @@ export const Checkbox = ({
 }) => {
   const iconContainerRef = useRef(null);
   const inputRef = useRef(null);
+  const isFirefox = getBrowserName() === "firefox";
```
orrgottlieb, line 33: please move to - isFirefox function in user-agent-utils?

orrgottlieb: no need for it to be a dependency

src/utils/user-agent-utils.js
```diff
@@ -0,0 +1,9 @@
+import { detect } from "detect-browser";
+
+const browser = detect();
+
+export function getBrowserName() {
```
orrgottlieb, line 5: export isFirefox function here

orrgottlieb (approved):

PR #535, opened 2022-02-15
Fix combobox drawer width

hadasfa:
change drawer appearance in combo box according to evigany notes
צילום מסך 2022-02-15 ב-13 36 35


laviomri (approved):
cool! any tests/stories that should be updated?

laviomri (approved):

PR #537, opened 2022-02-15
Fix tabs list bug - visible focus displayed when clicking out side and then clicking on one tab

hadasfa:


src/components/Tabs/Tab/Tab.scss
```diff
@@ -57,7 +57,7 @@
   }
 
   &.tab-focus-visible-inset {
-    @include focus-style-inset(3px, 3px);
+    @include focus-style-css-inset(3px, 3px);
```
hadasfa, line 60: the other mixin depend on using focus-visible class name which had different purpose (we inject it when we do real focus for old browsers support)

src/components/Tabs/TabList/TabList.jsx
```diff
@@ -42,68 +41,42 @@ const TabList = forwardRef(({ className, id, onTabChange, activeTabId, tabType,
...
-  const { focusWithinProps } = useFocusWithin({
-    onFocusWithin: () => {
-      setFocusTab(activeTabState);
-    },
-
-    onBlurWithin: () => {
-      setFocusTab(-1);
-    }
+  const getItemByIndex = useCallback(index => children[index], [children]);
+  const disabledIndexes = useMemo(() => Array.from(disabledTabIds), [disabledTabIds]);
```
laviomri, line 54: I'm probably missing something, but shouldn't `disabledTabIds` just be called `disabledTabIndexes`? what's the difference here between an id and an index?

laviomri (approved):

src/components/Tabs/TabList/TabList.jsx
hadasfa: the ids are index but saving in a set object which not support by the hook.

src/hooks/useGridKeyboardNavigation/useGridKeyboardNavigation.js
```diff
@@ -44,12 +44,12 @@ export default function useGridKeyboardNavigation({
...
   const skippedInitialActiveIndexChange = useRef(false);
   const [activeIndex, setActiveIndex] = useState(isInitialActiveState ? focusItemIndexOnMount : NO_ACTIVE_INDEX);
-  const [isUsingKeyboardNav, setIsUsingKeyboardNav] = useState(focusOnMount);
+  const isUsingKeyboardNav = useRef(true);
 
   const keyboardContext = useContext(GridKeyboardNavigationContext);
 
   const onArrowNavigation = direction => {
-    setIsUsingKeyboardNav(true);
+    isUsingKeyboardNav.current = true;
```
hadasfa, line 52: at first I change it to ref because I wanted to add some events and I was afraid from changing react state and renders side effects, but then I saw that there no need to add events.
I keep it as ref because it save some not needed renders but I can undo this change

src/hooks/useGridKeyboardNavigation/useGridKeyboardNavigation.js
```diff
@@ -82,13 +82,8 @@ export default function useGridKeyboardNavigation({
 
   const onFocus = useCallback(
     e => {
-      if (document.activeElement === ref.current) {
-        // since we use custom events (to add additional data like the navigation direction), it's possible that the element is already focused
-        return;
-      }
       const direction = e.detail?.keyboardDirection;
       if (direction) {
-        setIsUsingKeyboardNav(true);
```
laviomri, line 91: why was this removed?

laviomri (approved):

laviomri (approved):

PR #545, opened 2022-02-17
Better support for useDocumentEventListeners on Menu components

laviomri:
[Task](https://monday.monday.com/boards/245345663/pulses/2307148964)
I've added better handling for `useDocumentEventListeners`. On other Menu-related components, most of the handlers are being disabled when `useDocumentEventListeners = true`. I've matched the critical paths' behavior, to avoid stealing the focus when not intended.
NOTE: `MenuGridItem` will not work properly when `useDocumentEventListeners = true`, since most of the handlers will now be turned off. I assume here that there's no such use case yet, since `MenuGridItem` is very new, and `useDocumentEventListeners` is rarely used.

### DIFF
--- a/src/components/Menu/MenuItem/MenuItem.jsx
+++ b/src/components/Menu/MenuItem/MenuItem.jsx
@@ -130,10 +130,11 @@ const MenuItem = forwardRef(
     }, [shouldShowSubMenu, childElement, useDocumentEventListeners]);
 
     useEffect(() => {
+      if (useDocumentEventListeners) return;
       if (isActive) {
         referenceElement?.focus();
       }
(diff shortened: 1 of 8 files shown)


src/components/Menu/MenuItem/MenuItem.jsx
```diff
@@ -130,10 +130,11 @@ const MenuItem = forwardRef(
     }, [shouldShowSubMenu, childElement, useDocumentEventListeners]);
 
     useEffect(() => {
+      if (useDocumentEventListeners) return;
```
laviomri, line 133: This is the most important part

orrgottlieb (approved):

PR #551, opened 2022-02-21
support passing tooltip content without disabled mode

etaylib:
tooltip is shown only for disabled and disabled reasons. Adding support for plain tooltip content.

image

### DIFF
--- a/src/components/Menu/MenuItem/MenuItem.jsx
+++ b/src/components/Menu/MenuItem/MenuItem.jsx
@@ -299,7 +299,7 @@ MenuItem.defaultProps = {
   setSubMenuIsOpenByIndex: undefined,
   resetOpenSubMenuIndex: undefined,
   useDocumentEventListeners: false,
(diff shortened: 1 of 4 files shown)


src/components/Menu/MenuItem/MenuItem.jsx
```diff
@@ -209,8 +210,13 @@ const MenuItem = forwardRef(
       };
     }, [children, hasOpenSubMenu]);
 
-    const shouldShowTooltip = isTitleHoveredAndOverflowing || disabled;
-    const tooltipContent = disabled ? disableReason : title;
+    const shouldShowTooltip = isTitleHoveredAndOverflowing || disabled || tooltipContent;
+
+    const getTooltipContent = () => {
```
hadasfa, line 215: no need to put this logic inside a function which will create for each render.
you can set this logic as regular code inside the component.
If the logic were complex I would suggest use memo but it's only few conditions

PR #555, opened 2022-02-23
Fix stories

hadasfa:


src/storybook/functions/create-component-story.js
```diff
@@ -26,7 +26,14 @@ export function createStoryMetaSettings({ component, enumPropNamesArray, iconPro
 
   // set enum allowed values inside argsTypes object
   enumPropNamesArray?.forEach(propName => {
```
laviomri, line 28: consider renaming the arg name to `prop` so it won't confuse (since it can now be an object..)
```suggestion
  enumPropNamesArray?.forEach(prop => {
```

src/components/SplitButton/SplitButton.jsx
```diff
@@ -71,6 +71,7 @@ const SplitButton = ({
 
   // Refs //
   const ref = useRef(null);
+  console.log(ref.current);
```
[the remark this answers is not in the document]
hadasfa: will be removed :)

src/storybook/functions/create-component-story.js
hadasfa: will be fix:)

laviomri (approved):

PR #565, opened 2022-02-28
Editable heading callback on blur ignore

MosheZemah:
Callback for editable heading on blur ignore

### DIFF
--- a/src/components/EditableHeading/EditableHeading.jsx
+++ b/src/components/EditableHeading/EditableHeading.jsx
@@ -64,7 +64,7 @@ const EditableHeading = props => {
 
   const onIgnoreBlurEventCallback = useCallback(
     value => {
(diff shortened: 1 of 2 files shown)


src/components/EditableHeading/EditableHeading.jsx
```diff
@@ -61,6 +62,13 @@ const EditableHeading = props => {
     [onCancelEditing, setIsEditing]
   );
 
+  const onIgnoreBlurEventCallback = useCallback(
+    value => {
+      onIgnoreBlurEvent && onIgnoreBlurEvent(value);
```
laviomri, line 67:
```suggestion
      onIgnoreBlurEvent?.(value);
```

laviomri (approved):
LGTM though I'm not sure I fully get the use case. 
also, please add PropTypes. (should the storybook be updated too?)

PR #582, opened 2022-03-10
Interaction tests

MosheZemah:
Add interactions tests


src/components/Dropdown/__stories__/interactions.js
```diff
@@ -0,0 +1,26 @@
...
+  const dropdownElement = await getByRole(canvas, "textbox");
+  // Open the dropdown
+  await clickElement(dropdownElement);
+  // Filter it
+  await typeText(dropdownElement, "Option");
+  // Select the option
+  const optionToSelect = getByText(canvas, "Option 1");
+  await clickElement(optionToSelect);
+  // click the clear button
+  const clearButton = getByClassName("clear-indicator");
```
laviomri, line 21: is there a different locator we can use instead of classname? perhaps role, text or type?

src/components/Dropdown/__stories__/interactions.js
```diff
@@ -0,0 +1,26 @@
...
+  await clickElement(clearButton);
+  // Validate we see the placeholder again
+  const placeholderText = getByText(canvas, "Placeholder text here");
+  expect(placeholderText.innerText).toBe("Placeholder text here");
```
laviomri, line 25: something here seems redundant to me - if the text isn't what you expect, the `getByText` will fail before the `expect` (and if the element is found, then the `expect` is redundant)... what do you think?

src/__tests__/interactions-helper.js
```diff
@@ -0,0 +1,63 @@
...
+export const typeText = async (element, text, waitForDebounce = true) => {
+  let promise = userEvent.type(element, text, {
+    delay: 50
+  });
+  if (waitForDebounce) {
+    await promise;
+    return delay(250);
+  }
+  return promise;
+};
```
laviomri, line 57: WDYT about letting the user decide how much to wait?
(Also, I changed the code below to always return the promise's result, like you did on other places)
```suggestion
export const typeText = async (element, text, waitForDebounceMs = 250) => {
  let promise = userEvent.type(element, text, {
    delay: 50
  });
    const result = await promise;
    await delay(waitForDebounceMs);
    return result;
};
```

laviomri (approved):
looking very cool! 
@MosheZemah did you happen to see if chromatic runs the interaction tests? I think it does not...

PR #587, opened 2022-03-15
Fix radiobutton check issue

MosheZemah:
Fix issue with radio button needed to be clicked twice

### DIFF
--- a/src/components/RadioButton/RadioButton.jsx
+++ b/src/components/RadioButton/RadioButton.jsx
@@ -37,13 +37,6 @@ const RadioButton = forwardRef(
       }
     }, [onSelect, inputRef, disabled]);
 
-    const checkedProps = useMemo(() => {
-      if (checked !== undefined) {
-        return { checked };
-      }
-      return { defaultChecked };
-    }, [checked, defaultChecked]);
-
     return (
       <label className={cx(baseClassName, overrideClassName, { disabled })}>
         <span className={`${baseClassName}__radio-input-container`}>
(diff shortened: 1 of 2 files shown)


src/components/RadioButton/RadioButton.jsx
```diff
@@ -53,7 +46,7 @@ const RadioButton = forwardRef(
             value={value}
             name={name}
             disabled={disabled}
-            {...checkedProps}
+            defaultChecked={checked || defaultChecked}
```
hadasfa, line 49: but checked prop is for controlled state. it's means that we want that when clicking on the radiobutton it's state will not changed (only if we changed it by our prop) - and not changed if we click once / twice.

src/components/RadioButton/__stories__/radioButton.stories.mdx
```diff
@@ -61,62 +75,94 @@ Radio buttons have 3 states: regular, selected, and disabled.
...
       <RadioButton text="All updates" name="radio-buttons-group-4" />
     </div>
   </Story>
 </Canvas>
 
+Controled externally.
+
+<Canvas>
+  <Story name="Controlled Radio buttons" play={controlRadioButton}>
+    {controlledRadioButton.bind({})}
```
hadasfa, line 162: Developer will not be able to see the code this way. The best thing to do will be to put this code as part of the mdx file (example in combobox stories ). if it's not possible can you please use ControlledRadioButton as the template? (for developers be able to see the code when click on "show code")

hadasfa (approved):

PR #590, opened 2022-03-16
TS button POC

sahariko:
A quick POC on TS-ifying a component (`<Button>`).

Prop table in Storybook looks great as well:
Screen Shot 2022-03-16 at 18 26 38


src/components/Button/Button.tsx
```diff
@@ -181,11 +246,16 @@ const Button = forwardRef(
...
         "aria-haspopup": ariaHasPopup,
         "aria-expanded": ariaExpanded,
         "aria-controls": ariaControls
       };
+
+      if (loading) {
+        props["aria-busy"] = "true";
+      }
+
+      return props;
```
laviomri, line 258: interesting - why did you need to put the `loading` outside of the object?

src/components/Button/ButtonConstants.ts
```diff
@@ -0,0 +1,34 @@
...
+  NEGATIVE = "negative",
+  ON_PRIMARY_COLOR = "on-primary-color",
+  ON_INVERTED_BACKGROUND = "on-inverted-background"
+}
+
+const OLD_BUTTON_SIZES: Record<string, string> = {
+  sm: SIZES.SMALL,
+  md: SIZES.MEDIUM,
+  lg: SIZES.LARGE
+};
```
laviomri, line 21: I think `as const` will give us better typing in this case
```suggestion
const OLD_BUTTON_SIZES = {
  sm: SIZES.SMALL,
  md: SIZES.MEDIUM,
  lg: SIZES.LARGE
} as const;
```

src/components/Button/Button.tsx
sahariko: The explicit `undefined` messed TS's type inference when spreading onto the element later on.

src/components/Button/Button.tsx
```diff
@@ -8,61 +8,126 @@ import useMergeRefs from "hooks/useMergeRefs";
 import { NOOP } from "utils/function-utils";
 import Icon from "components/Icon/Icon";
 import Loader from "components/Loader/Loader";
-import { BUTTON_COLORS, BUTTON_INPUT_TYPE, BUTTON_TYPES, getActualSize } from "./ButtonConstants";
+import { ButtonColor, ButtonInputType, ButtonType, getActualSize } from "./ButtonConstants";
 import { getParentBackgroundColorNotTransparent, TRANSPARENT_COLOR } from "./helper/dom-helpers";
 import "./Button.scss";
+import { Rect } from "@popperjs/core";
```
laviomri, line 14: unused?

src/components/Button/Button.tsx
```diff
@@ -232,6 +302,7 @@ const Button = forwardRef(
       return (
         <button {...buttonProps}>
           <span className="monday-style-button__loader">
+            {/** @ts-ignore */}
```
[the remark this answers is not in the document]
sahariko: It's not `forwardRef`, it's that these components are not TS, so TS doesn't know what props they can receive.

src/components/Button/Button.tsx
```diff
@@ -8,61 +8,126 @@ import useMergeRefs from "hooks/useMergeRefs";
...
 
+// @ts-ignore
 const isIE11 = !!window.MSInputMethodContext && !!document.documentMode;
 
 // min button width
 const MIN_BUTTON_HEIGHT_PX = isIE11 ? 32 : 6;
 const UPDATE_CSS_VARIABLES_DEBOUNCE = 200;
 
-const Button = forwardRef(
+export interface ButtonProps {
```
laviomri, line 23: I think publisher-components need an update as well, no?

sahariko: True, but this is just a POC, so I kept only the really relevant files 🙂

VerHext:
Are contributions to this POC welcome?

PR #595, opened 2022-03-21
Add aria-label to Flex

Naama-Weber-Monday:
Added aria label to Flex.

### DIFF
--- a/src/components/Flex/Flex.jsx
+++ b/src/components/Flex/Flex.jsx
@@ -9,7 +9,7 @@ import classes from "./Flex.module.scss";
 
 const Flex = forwardRef(
   (
-    { className, id, elementType, direction, wrap, children, justify, align, gap, onClick, style, ariaLabelledby },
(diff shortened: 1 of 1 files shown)


src/components/Flex/Flex.jsx
```diff
@@ -93,7 +95,8 @@ Flex.defaultProps = {
   direction: Flex.directions.ROW,
   justify: Flex.justify.START,
   align: Flex.align.CENTER,
-  gap: Flex.gaps.NONE
+  gap: Flex.gaps.NONE,
+  'aria-label': ""
```
orrgottlieb, line 99: ariaLabel - and it should be undefined

orrgottlieb (changes requested):

src/components/Flex/Flex.jsx
Naama-Weber-Monday: Done

orrgottlieb (approved):

PR #607, opened 2022-03-29
Color picker interaction tests

laviomri:


src/__tests__/interactions-helper.js
```diff
@@ -108,6 +108,11 @@ export const waitForElementVisible = getterFunc => {
   });
 };
 
+export const keyboardMultipleTimes = async (text, times, options = { delay: 70 }) => {
```
hadasfa, line 111: is the word time is the correct word? it's not about time it's about amount (can be confiusing)

src/components/ColorPicker/components/ColorPickerContent/ColorPickerColorsGrid.jsx
```diff
@@ -55,6 +55,7 @@ export const ColorPickerColorsGrid = React.forwardRef(
               colorSize={colorSize}
               tooltipContent={tooltipContentByColor[color]}
               colorShape={colorShape}
+              data-testid={`color-picker-item-${color}`}
```
hadasfa, line 58: I saw we create get test id function, should we use it? it create a data test id which depend on element type enum and also the id

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.jsx
```diff
@@ -62,6 +63,7 @@ const ColorPickerItemComponent = ({
           active: isActive,
           circle: colorShape === COLOR_SHAPES.CIRCLE
         })}
+        data-testid={dataTestId}
```
hadasfa, line 66: same question about data test id

src/__tests__/interactions-helper.js
laviomri: will update to "count" 👍

hadasfa (approved):

PR #611, opened 2022-03-29
test(Slider): add interaction tests for ranged mode 

niksa-monday:
- add interaction tests for ranged mode
 - minor fixes


src/components/Slider/__tests__/Slider.interactions.js
```diff
@@ -9,24 +9,56 @@ const changeSliderValueByClickingOnTrackTest = async canvas => {
   // prepare: take sizes of slider and waiting for render
   const elRail = canvas.getByTestId("monday-slider-show-value-s__rail");
   const rect = elRail.getBoundingClientRect();
-  const elThumb = await waitForElementVisible(() => within(elRail).getByRole("slider"));
+  const elThumb = await within(elRail).findByRole("slider");
```
niksa-monday, line 12: waitForElementVisible - add failed expectations at log :(
waitForElementVisible replaced by correct findByRole

waitForElementVisible

laviomri: good thing you noticed it. (behind the scenes `waitFor` is attempting to find the element multiple times, and each failure adds an error).
would you say that we should avoid `waitForElementVisible` to prevent such errors in the log?

laviomri (approved):
awesome

src/components/Slider/__tests__/Slider.interactions.js
niksa-monday: I think yes ...

PR #617, opened 2022-03-31
feat(accordion): add data-testid, improve ids, add interaction tests

niksa-monday:
Add data-testid (reformat component)

Minor code conventions (lint) refactoring (remove redundant const)
 
 Improve AccordionItem/ExpandCollapse ids
 - ExpandCollapse component have to get prop id
   for correct A11y (aria-controls)
 - AccordionItem now getting id on base of Accordion-Id and Index
 - If Accordion-Id not specified we are taking default COMPONENT_ID

Add interaction tests for single-active mode
 - add interactions tests
 - connect tests to storybook
 - minor fix: remove unused import

### DIFF
--- a/src/components/Accordion/Accordion/Accordion.jsx
+++ b/src/components/Accordion/Accordion/Accordion.jsx
@@ -4,6 +4,18 @@ import cx from "classnames";
 import useMergeRefs from "../../../hooks/useMergeRefs";
 import "./Accordion.scss";
 
+const COMPONENT_ID = "monday-accordion";
+
+function defineChildId(index, props, accordionId) {
+  if (props.id) {
+    return props.id;
+  }
+  if (accordionId) {
+    return `${accordionId}--item-${index}`;
+  }
+  return `${COMPONENT_ID}--item-${index}`;
+}
+
 const Accordion = forwardRef(
   ({ children: originalChildren, allowMultiple, "data-testid": dataTestId, defaultIndex, className, id }, ref) => {
     const componentRef = useRef(null);
@@ -43,15 +55,18 @@ const Accordion = forwardRef(
 
     const renderChildElements = useMemo(() => {
       return React.Children.map(children, (child, itemIndex) => {
+        const originalProps = { ...child?.props };
+        const childId = defineChildId(itemIndex, originalProps, id);
         return React.cloneElement(child, {
-          ...child?.props,
+          ...originalProps,
+          id: childId,
           onClickAccordionCallback: () => {
             onChildClick(itemIndex);
           },
           open: isChildExpanded(itemIndex)
         });
       });
-    }, [isChildExpanded, onChildClick, children]);
+    }, [children, id, isChildExpanded, onChildClick]);
 
     return (
       <div ref={mergedRef} className={cx("accordion", className)} data-testid={dataTestId} id={id}>
@@ -93,7 +108,7 @@ Accordion.defaultProps = {
   id: undefined,
   allowMultiple: false,
   children: null,
-  "data-testid": "monday-accordion",
+  "data-testid": COMPONENT_ID,
   defaultIndex: []
 };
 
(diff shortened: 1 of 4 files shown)


src/components/Accordion/Accordion/Accordion.jsx
```diff
@@ -4,62 +4,77 @@ import cx from "classnames";
 import useMergeRefs from "../../../hooks/useMergeRefs";
 import "./Accordion.scss";
 
-const Accordion = forwardRef(({ children: originalChildren, allowMultiple, defaultIndex, className, id }, ref) => {
-  const componentRef = useRef(null);
-  const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+const COMPONENT_ID = "monday-accordion";
```
laviomri, line 7: consider using `ELEMENT_TYPES` along with `getTestId` (see [here](https://github.com/mondaycom/monday-ui-react-core/blob/master/src/utils/test-utils.js))

src/components/Accordion/Accordion/Accordion.jsx
```diff
@@ -4,62 +4,77 @@ import cx from "classnames";
...
-      return React.cloneElement(child, {
-        ...child?.props,
-        onClickAccordionCallback: () => {
-          onChildClick(itemIndex);
-        },
-        open: isChildExpanded(itemIndex)
+    const renderChildElements = useMemo(() => {
+      return React.Children.map(children, (child, itemIndex) => {
+        const originalProps = { ...child?.props };
+        const childId = defineChildId(itemIndex, originalProps, id);
```
laviomri, line 59: I think it's quite ok to call `getTestId` with just the itemIndex, without the accordion id.
this MAY cause multiple accordion items with the same test ids, but that's ok. testIds aren't like ids - they don't have to be unique. Furthermore, if there will be a case where a test needs to have two accordions and need to get a specific accordion item within one of the accordions, we can do something like:
`within(getByTestId(ACCORDION_ID)).getByTestId(ACCORDION_ITEM_ID)`
(I hope it makes sense).

I think it will make things a bit simpler regarding to the test ids logic :)

src/components/Accordion/Accordion/__tests__/accordion.interactions.js
```diff
@@ -0,0 +1,54 @@
...
+  await expect(elHeading.getAttribute("aria-expanded")).toBe("true");
+  await expect(elHeading.getAttribute("aria-controls")).toBe(elPanel.id);
+  await delay(CHANGES_DELAY);
+
+  // select back second (1) AccordionItem
+  elHeading = getAccordionHeadingBtText(canvas, "Setting");
+  userEvent.click(elHeading);
+  elPanel = canvas.getByRole("region");
+  await expect(elHeading.getAttribute("aria-expanded")).toBe("true");
+  await expect(elHeading.getAttribute("aria-controls")).toBe(elPanel.id);
```
laviomri, line 45: would you like to check here as well that `aria-expanded` of the first item is false?

laviomri (approved):

src/components/Accordion/Accordion/Accordion.jsx
niksa-monday: It's about id attribute NOT data-testid

laviomri: oh I see it now, scratch that :)

src/components/Accordion/Accordion/__tests__/accordion.interactions.js
niksa-monday: added:
https://github.com/mondaycom/monday-ui-react-core/pull/617/commits/86990168c387433e2dd28b902c12bba8cc86485d

src/components/Accordion/Accordion/Accordion.jsx
niksa-monday: I've opened some discussion about data-testid.
So for a while I will not perform changes, after we'll make decision I'll fix it in separate PR.

PR #620, opened 2022-04-03
docs(accordion): improve storybook (sub-items, props, etc.)

niksa-monday:
- add ability to change props directly at Storybook
 - minor fixes of props description
 
improve-accordion-01

 - add AccordionItem to ArgsTable:
mprove-accordion-02

- ads description to Sandbox Controls:
mprove-accordion-03

I think as future step we also can add AccordionItem to Controls/Playground using one of this approaches:
- https://storybook.js.org/docs/react/writing-stories/stories-for-multiple-components
- https://storybook.js.org/docs/react/writing-docs/doc-block-argstable#grouping

### DIFF
--- a/src/components/Accordion/Accordion/__stories__/accordion.stories.mdx
+++ b/src/components/Accordion/Accordion/__stories__/accordion.stories.mdx
@@ -9,6 +9,17 @@ import { accordionSingleActivePlaySuite, accordionMultiActivePlaySuite } from ".
 <Meta
   title="Data display/Accordion"
   component={ Accordion }
+  subcomponents = {{AccordionItem}}
+  argTypes={{
+    children: { control: false },
+    defaultIndex: { control: false }
+  }}
+  parameters={{
+    controls: {
+      expanded: true,
+      sort: 'requiredFirst'
+    }
+  }}
 />
 
 <!--- Component template -->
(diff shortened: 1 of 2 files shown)


src/components/Accordion/Accordion/__stories__/accordion.stories.mdx
```diff
@@ -9,6 +9,17 @@ import { accordionSingleActivePlaySuite, accordionMultiActivePlaySuite } from ".
...
   component={ Accordion }
+  subcomponents = {{AccordionItem}}
+  argTypes={{
+    children: { control: false },
+    defaultIndex: { control: false }
+  }}
+  parameters={{
+    controls: {
+      expanded: true,
+      sort: 'requiredFirst'
```
laviomri, line 20: awesome, wouldn't we like to apply this by default to all stories? maybe with storybook's global config?

niksa-monday: Yes we want. This is just an example on one of Stories. We can apply it to all stories according to:
https://storybook.js.org/docs/react/essentials/controls#show-full-documentation-for-each-property

.storybook/preview.js
```diff
@@ -19,6 +19,10 @@ import {
 } from "../src/storybook/components";
 
 addParameters({
+  controls: {
+    expanded: true,
```
hadasfa, line 23: what this will do ?

hadasfa (approved):

.storybook/preview.js
niksa-monday: Last image at PR desciption

PR #625, opened 2022-04-05
add support on virtualized list items inside list component

hadasfa:
צילום מסך 2022-04-10 ב-9 47 15


src/components/Combobox/Combobox.jsx
```diff
@@ -301,8 +306,9 @@ Combobox.defaultProps = {
   maxOptionsWithoutScroll: undefined,
   /**
    * Callback that called after clicking on the add new combo box button.
+   * @param {string} _filterValue
    */
-  onAddNew: undefined,
+  onAddNew: _filterValue => {},
```
hadasfa, line 311: for better documentation on storybook

src/components/List/List.jsx
```diff
@@ -3,23 +3,51 @@ import PropTypes from "prop-types";
...
-    >
-      {children}
-    </Component>
-  );
-});
+const List = forwardRef(
+  ({ className, id, component, children, dense, ariaLabel, ariaDescribedBy, renderOnlyVisibleItems, style }, ref) => {
+    const componentRef = useRef(null);
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    const Component = useMemo(() => component, [component]);
```
MosheZemah, line 13:
```suggestion
    const Component = Component
```

src/components/List/List.jsx
```diff
@@ -3,23 +3,51 @@ import PropTypes from "prop-types";
...
+    const overrideChildren = useMemo(() => {
+      let override = children;
+      if (Array.isArray(override) && override.length > 0) {
```
MosheZemah, line 16:
```suggestion
      if (Array.isArray(children) && children.length > 0) {
```

src/components/List/List.jsx
```diff
@@ -3,23 +3,51 @@ import PropTypes from "prop-types";
...
+        const originalFirstChildren = override[0];
```
MosheZemah, line 17:
```suggestion
        const originalFirstChild = override[0];
```

src/components/List/List.jsx
```diff
@@ -3,23 +3,51 @@ import PropTypes from "prop-types";
...
+        if (originalFirstChildren.type === ListTitle) {
+          const firstChild = React.cloneElement(originalFirstChildren, {
+            ...originalFirstChildren?.props,
+            className: `${originalFirstChildren.className || ""} monday-style-list_category--first`
+          });
+          override = [firstChild, ...override.slice(1)];
```
MosheZemah, line 23:
```suggestion
          override = [firstChild, ...children.slice(1)];
```

src/components/List/List.jsx
```diff
@@ -46,7 +74,12 @@ List.propTypes = {
    * ARIA described by string to reference an id to describe by
    */
   ariaDescribedBy: PropTypes.string,
-  children: PropTypes.any
+  children: PropTypes.any,
```
MosheZemah, line 77: I think children should not be defaultProp as it is not a regular prop

src/components/List/List.jsx
```diff
@@ -3,23 +3,51 @@ import PropTypes from "prop-types";
...
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    const Component = useMemo(() => component, [component]);
+    const overrideChildren = useMemo(() => {
+      let override = children;
+      if (Array.isArray(override) && override.length > 0) {
+        const originalFirstChildren = override[0];
+        if (originalFirstChildren.type === ListTitle) {
+          const firstChild = React.cloneElement(originalFirstChildren, {
+            ...originalFirstChildren?.props,
+            className: `${originalFirstChildren.className || ""} monday-style-list_category--first`
```
MosheZemah, line 21: all this just to add a class "first" - why not use a selector ":first" (also check it is a ListTitle)

src/components/List/List.scss
```diff
@@ -7,4 +7,8 @@
   padding: 0;
   overflow: auto;
   @include scroller();
-}
+
+  &_category--first {
```
MosheZemah, line 11: try use a ":first-child" selector (just for ListTitle if it has a className)

src/components/List/VirtualizedListItems/VirtualizedListItems.jsx
```diff
@@ -0,0 +1,19 @@
+import cx from "classnames";
+import { VirtualizedList } from "components";
+import {
+  useChildrenAsDataObjects,
+  useListItemsRenderer
+} from "components/List/VirtualizedListItems/VirtualizedListItemsHooks";
```
MosheZemah, line 6: I'm not sure why this split is needed - why not having all the hooks in this file? I think it is more convenient

src/components/List/VirtualizedListItems/VirtualizedListItemsHooks.js
```diff
@@ -0,0 +1,61 @@
...
+
+const ITEM_CHILDREN_TYPES = {
+  TITLE: "title",
+  ITEM: "item"
+};
+
+const LIST_TITLE_SPACING = 24;
+const LIST_ITEM_HEIGHT = 32;
+
+export function useChildrenAsDataObjects(children) {
```
MosheZemah, line 12: this hook should useMemo

src/components/List/VirtualizedListItems/VirtualizedListItemsHooks.js
```diff
@@ -0,0 +1,61 @@
...
+          ...otherProps,
+          // avoid add spacing to the first category on the list
+          height: isFirstCategory ? LIST_ITEM_HEIGHT : LIST_ITEM_HEIGHT + LIST_TITLE_SPACING
+        };
+      } else if (child.type === ListItem) {
+        const { id, ...otherProps } = child.props;
+        return {
+          type: ITEM_CHILDREN_TYPES.ITEM,
+          id: id || `list-item-${index}`,
+          ...otherProps,
```
MosheZemah, line 30: props should have the original id since it is used in ListItem/ListTitle

src/components/List/VirtualizedListItems/VirtualizedListItemsHooks.js
```diff
@@ -0,0 +1,61 @@
...
+export function useChildrenAsDataObjects(children) {
+  return children
+    .map((child, index) => {
+      if (child.type === ListTitle) {
+        const isFirstCategory = index === 0;
+        const { id, ...otherProps } = child.props;
+        return {
+          type: ITEM_CHILDREN_TYPES.TITLE,
+          id: id || `list-title-${index}`,
+          ...otherProps,
```
MosheZemah, line 21:
```suggestion
          id: `list-title-${index}`,
          props: child.props,
```

src/components/List/VirtualizedListItems/VirtualizedListItemsHooks.js
MosheZemah, line 30:
```suggestion
          id: `list-title-${index}`,
          props: child.props,
```

src/components/List/VirtualizedListItems/VirtualizedListItemsHooks.js
```diff
@@ -0,0 +1,61 @@
...
+      } else {
+        return undefined;
+      }
+    })
+    .filter(item => item !== undefined);
+}
+
+export function useListItemsRenderer() {
+  return useCallback((item, index, style) => {
+    const { type, ...otherProps } = item;
```
MosheZemah, line 42:
```suggestion
    const { type, props } = item;
```

src/components/List/VirtualizedListItems/VirtualizedListItemsHooks.js
```diff
@@ -0,0 +1,61 @@
...
+    let element;
+    switch (type) {
+      case ITEM_CHILDREN_TYPES.TITLE: {
+        element = <ListTitle {...otherProps} />;
```
MosheZemah, line 46:
```suggestion
        element = <ListTitle {...props} />;
```

MosheZemah (changes requested):

src/components/List/List.jsx
hadasfa: We do document children as one of the props in the prop types and default props because it is our library standard.
I did change the children prop type thought to be more accurate .

src/components/List/List.scss
hadasfa: Its problematic because its take the first list title in each list title batch (which means it will be relevant to all the list title in the page)

src/components/List/VirtualizedListItems/VirtualizedListItems.jsx
hadasfa: I prefer to keep separating the hooks to different file from the component itself.
Maybe the hooks itself are not needed? I do prefer to move big stand alone logic outside to hooks when possible so I prefer to keep it like this, but I can remove the hooks if you think it is better.

src/components/List/VirtualizedListItems/VirtualizedListItemsHooks.js
hadasfa: id is one of the component props ^^"

src/components/List/VirtualizedListItems/VirtualizedListItems.jsx
MosheZemah: I don't see the login seperating the logic here, since it is not for using it somewhere else. It is very coupled to this component, and it is more convenient to have the code in one place if possible. It also not that long

src/components/List/VirtualizedListItems/VirtualizedListItems.jsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const VirtualizedListItems = ({ children }) => {
+  const items = useChildrenAsDataObjects(children);
+  const itemRenderer = useListItemsRenderer();
+
+  return (
+    <VirtualizedList
+      scrollableClassName={cx("monday-style-list", "monday-style-list--virtualized")}
```
MosheZemah, line 14: I think it is odd we'll have monday-style-list inside monday-style-list

MosheZemah (changes requested):

src/components/List/List.jsx
hadasfa: thats the standard we implement in all our comps but I did change the prop type to be more accurate :)

MosheZemah (approved):

PR #632, opened 2022-04-11
Support on disabled tabIndex

NofarRG:
Breadcrumbs accessibility fixes:
- Add an optional disabled prop to `BreadcrumbContent` according to `disabled || isDisabled` in `BreadcrumbItem`.
- If `BreadcrumbItem` is disabled, sets the tabindex attribute to `-1`.

### DIFF
--- a/src/components/BreadcrumbsBar/BreadcrumbItem/BreadcrumbItem.jsx
+++ b/src/components/BreadcrumbsBar/BreadcrumbItem/BreadcrumbItem.jsx
@@ -57,6 +57,7 @@ const BreadcrumbItem = ({
           text={text}
           icon={icon}
           isCurrent={isCurrent}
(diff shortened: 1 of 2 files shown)


src/components/BreadcrumbsBar/BreadcrumbItem/BreadcrumbItem.jsx
```diff
@@ -57,6 +57,7 @@ const BreadcrumbItem = ({
           text={text}
           icon={icon}
           isCurrent={isCurrent}
+          disabled={isDisabled || disabled}
```
hadasfa, line 60: please replace with overrideDisabled which is purpose is to support disabled prop and also backward compatibility for deprecate isDisabled

PR #642, opened 2022-04-18
support scroll trigger in dialog

roniCohen123:


src/components/Dialog/Dialog.jsx
```diff
@@ -207,6 +208,10 @@ export default class Dialog extends PureComponent {
     this.handleEvent("mouseleave", e.target, e);
   }
 
+  onScroll(e) {
```
hadasfa, line 211: please check why snapshot tests fail and fix where needed

hadasfa (changes requested):
see note :)

PR #643, opened 2022-04-18
remove set style hardcoded for icon button (not needed and not used u…

hadasfa:
…ntil I fixed typo before)

#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


src/components/IconButton/IconButton.jsx
```diff
@@ -49,7 +49,6 @@ const IconButton = forwardRef(
 
     const overrideStyle = useMemo(() => {
       let style = {
-        display: "inline-flex",
```
hadasfa, line 52: defined inside button and until now was not active because include typo in inline-flex

PR #653, opened 2022-04-26
Fix usage on useMergedRef in checkbox

hadasfa:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests

### DIFF
--- a/src/components/Checkbox/Checkbox.js
+++ b/src/components/Checkbox/Checkbox.js
@@ -39,7 +39,7 @@ export const Checkbox = forwardRef(
     const mergedInputRef = useMergeRefs({ refs: [ref, inputRef] });
     const overrideClassName = backwardCompatibilityForProperties([className, componentClassName]);
     const onMouseUpCallback = useCallback(() => {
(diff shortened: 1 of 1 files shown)


src/components/Checkbox/Checkbox.js
```diff
@@ -39,15 +39,15 @@ export const Checkbox = forwardRef(
     const mergedInputRef = useMergeRefs({ refs: [ref, inputRef] });
     const overrideClassName = backwardCompatibilityForProperties([className, componentClassName]);
     const onMouseUpCallback = useCallback(() => {
-      const input = mergedInputRef.current;
+      const input = inputRef.current;
```
hadasfa, line 42: The way I used mergedInputRef was wrong, now the mergedInputRef is passed to the input as ref but everywhere I am keep using in the ref which I know that it have a value.
Using mergedInputRef as ref cause bug on the checkbox component

hadasfa: I will add documentation about this mistake in storybook/ the hook code on a separate PR

PR #657, opened 2022-05-03
Storybook: usePrevios hook

SergeyRoyt:
UsePrevious hook added to storybook

#### Basic
- [ ] PR has description.
#### Style
#### Storybook
- [X] Stories include all flows of using the component.
#### Tests


src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+Hook for keeping previous state value.
+
+<Canvas>
+  <Story name="Overview">
+    {() => {
+      const [count, setCount] = useState(1);
+      const prevCount = usePrevious(count);
+      const incrementButtonOnClick = () => {
+        setCount(prevCount => prevCount + 1);
```
orrgottlieb, line 28: please change the name of the argument - as you already have a `prevCount` variable it might be confusing for someone who will look at the code

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+          <Button onClick={incrementButtonOnClick}>Increment</Button>
+        </div>
+      );
+    }}
+  </Story>
+</Canvas>
+
+## Arguments
+
+<ArgsTable of={UsePrevious} />
```
orrgottlieb, line 49: please see other hooks stories for consistency of hooks tables (`useClickOutside.stories.mdx` for example)

orrgottlieb (changes requested):
Very cool and very useful hook!

Great PR!

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
SergeyRoyt: We've been discussing it with @hadasfa. The storybook for hooks is not very consistent right now, so we've decided that we'll make it consistent and that it'll look like this: title - arguments (cause hook has arguments, not props), content - ArgsTable (cause it's more readable than FunctionArguments). What do you think?

hadasfa: as far as I know we still using the props table for arguments with the "fake component" for the props api
and using the args table for the return value. @orrgottlieb 
I also think it is more readable.
Am I wrong?

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+<Canvas>
+  <Story name="Overview">
+    {() => {
+      const [count, setCount] = useState(1);
+      const prevCount = usePrevious(count);
+      const incrementButtonOnClick = () => {
+        setCount(prevValue => prevValue + 1);
+      };
+      return (
+        <div>
```
hadasfa, line 31: We can remove the outside div because it does not have any style / classes inside and therefore not needed

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+          <Flex className="use-previous-counter-container" direction={Flex.directions.COLUMN}>
```
hadasfa, line 32: let's use css modules for every new class because we planing to migrate everything to css modules is this q or the next one.

hadasfa: also let's talk about how to name classes when using css modules :)

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,49 @@
...
+            <div className="use-previous-counter-label">Current</div>
+            <Counter count={count} ariaLabeledBy="currentCounterLabel" />
```
hadasfa, line 34: Why you add this ariaLabeledBy prop here?

hadasfa (changes requested):
Awesome PR, wrote few notes. Let's talk :)

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
SergeyRoyt: I've used div parent as with this` return ()` syntax we have to return single parent element always. We could use `return [<>, <>]` syntax though, but I believe it's a bad option in most cases
Replaced div with Flex for more clarity and flexibility. Is that ok?

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
SergeyRoyt: Removed dummy component and replaced `ArgsTable` with `FunctionArguments`

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
hadasfa: I saw you removed it and the flex become the single parent element (which it was there before so its ok)
if you don't need flex and just looking for wrapping your return react code, you can return react fragment like this: `<><div>1</div><div>2</div></>`

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,51 @@
+import { useState } from "react";
+import { Story, Canvas, Meta } from "@storybook/addon-docs";
+import usePrevious from "../index";
+import Button from "components/Button/Button";
+import { Counter } from "components";
+import Flex from "components/Flex/Flex";
+import styles from "./usePrevious.stories.module.scss";
```
hadasfa, line 7: import styles / scss / css file will always be last :)

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,51 @@
...
+import { FunctionArgument, FunctionArguments } from "../../../storybook";
+
+<Meta title="Hooks/usePrevious" />
+
+# usePrevious
+
+- [Overview](#overview)
+- [Arguments](#arguments)
```
hadasfa, line 15: no return value?

src/hooks/usePrevious/__stories__/usePrevious.stories.mdx
```diff
@@ -0,0 +1,51 @@
...
+  <Story name="Overview">
+    {() => {
+      const [count, setCount] = useState(1);
+      const prevCount = usePrevious(count);
+      const incrementButtonOnClick = () => {
+        setCount(prevValue => prevValue + 1);
+      };
+      return (
+        <Flex direction={Flex.directions.COLUMN}>
+          <Flex className={styles.counterContainer} direction={Flex.directions.COLUMN}>
```
hadasfa, line 32: great !

hadasfa (approved):
yay!

PR #677, opened 2022-05-19
fix: fix issue with ButtonIcon inline issue

orrgottlieb:
Fix an issue with the wrapper of the icon button doesn't calculate the size of the content correctly


src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.scss
```diff
@@ -1,13 +1,13 @@
...
 
 .monday-style-color-item-wrapper {
   --border-radius-multiplier: 1;
 
   &.circle {
     --border-radius-multiplier: 99999;
   }
-
+  /* stylelint-disable scss/operator-no-unspaced */
   @mixin absolute-border($border-width, $border-color-var-name, $inset: -$border-width, $border-radius: 7px) {
```
laviomri, line 11: why did you need the disable - was there a problem with `$inset: -$border-width`? 
if so, should we fix instead of disabling?

laviomri (approved):

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.scss
orrgottlieb: @laviomri will fix it after merge,
this branch should have no "potential" change

PR #681, opened 2022-05-23
fix: Remove Tip box

abaum-augu:
The link referred on the Tip box is broken. Suggestion to remove it unless there is a new Canvas component with a different name. 
Sorry for putting my nose in 😅   just find out the bug and Orr told me this is open source.


orrgottlieb (approved):

orrgottlieb:
Thank you!!

PR #689, opened 2022-05-29
Feature/sergeyro/hooks plop improve

SergeyRoyt:
https://monday.monday.com/boards/245345663/pulses/2723949171
Hook plop improvements:
- Creation of the hook file itself, story and story-styles in corresponding directories
- Auto-exposure: automatically add new hook to published-components and hooks/index
- Removal of dummy components


plop/hooks/hook-story.txt
```diff
@@ -20,16 +19,28 @@ Please write here the component description
...
   </Story>
 </Canvas>
 
-## Props
-<ArgsTable of={ {{properCase componentName}} } />
+## Arguments
 
+<FunctionArguments>
+  <FunctionArgument name="options" type="Object">
+    <FunctionArgument name="value" type="any" description={<>Example of required argument.</>} required />
```
hadasfa, line 36: I am not sure <> is needed if the value of the description is text only. it does not working with string?

hadasfa: and maybe let's add here another fields example such as string or int - inside the options?

webpack/published-components.js
```diff
@@ -85,7 +85,8 @@ const publishedComponents = {
   Avatar: "components/Avatar/Avatar.jsx",
   IconButton: "components/IconButton/IconButton.jsx",
   Slider: "components/Slider/Slider.jsx",
-  // hooks
+  // Don't remove next line
```
hadasfa, line 88: aren't we need to add the same comment and plop maker to path: "src/hooks/index.js" too? why this one is working without it?

hadasfa (approved):

plop/hooks/hook-story.txt
SergeyRoyt: It doesn't work with a string

webpack/published-components.js
SergeyRoyt: It's all about the pattern. In hooks/index.js we just replace last \n with a new string and in published-components.js we should put the line in the middle of the file, so we are using plop marker - it's just the easiest way to write this regex
```
      {
        type: "append",
        path: "src/hooks/index.js",
        pattern: /(\n$)/gm,
        separator: "",
        template: `export { default as {{camelCase hookName}} } from "./{{camelCase hookName}}";\n`
      },
      {
        type: "append",
        path: "webpack/published-components.js",
        pattern: /(\/\/ plop_marker:published-hooks)/g,
        template: `  {{camelCase hookName}}: "hooks/{{camelCase hookName}}",`
      }
```

PR #694, opened 2022-06-12
Feature/sergeyro/avatar group

SergeyRoyt:
#### Basic
- [X] Used plop (`npm run plop`) to create a new component.
- [X] PR has description.
- [X] Component defines [`PropTypes`](https://reactjs.org/docs/typechecking-with-proptypes.html).
#### Style
- [X] Styles are added to `NewComponent.modules.scss` file inside of the `NewComponent` folder.
- [X] Component uses CSS Modules.
- [X] Design is compatible with [Monday Design System](https://design.monday.com/).
#### Storybook
- [] Stories include all flows of using the component.
#### Tests


src/components/Avatar/Avatar.jsx
```diff
@@ -49,6 +54,8 @@ const Avatar = ({
     return customSize ? { height: `${customSize}px`, width: `${customSize}px` } : {};
   }, [customSize]);
 
+  ariaLabel = ariaLabel || (typeof tooltipProps?.content === "string" ? tooltipProps.content : undefined);
```
hadasfa, line 57: If someone pass tooltipProps.content is should "win" the aria label, because the aria label is the default (only if there is no other configuration)

src/components/Avatar/Avatar.jsx
```diff
@@ -95,26 +102,62 @@ const Avatar = ({
     return badges.length > 0 ? <div className={cx(bemHelper({ element: "badges" }))}>{badges}</div> : null;
   }, [size, topLeftBadgeProps, topRightBadgeProps, bottomLeftBadgeProps, bottomRightBadgeProps]);
 
+  const TooltipContainer = ({ children }) => {
+    if (!tooltipProps) {
+      return <>{children}</>;
```
hadasfa, line 107: you can just write `return children;` instead `return <>{children}</>;`

src/components/Avatar/Avatar.jsx
```diff
@@ -95,26 +102,62 @@ const Avatar = ({
     return badges.length > 0 ? <div className={cx(bemHelper({ element: "badges" }))}>{badges}</div> : null;
   }, [size, topLeftBadgeProps, topRightBadgeProps, bottomLeftBadgeProps, bottomRightBadgeProps]);
 
+  const TooltipContainer = ({ children }) => {
```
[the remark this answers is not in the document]
hadasfa: if the issue is that you have 2 conditions it will be still clearer and more efficient to write to conditions instead creating 2 anonymous components in the render function :)

src/components/Avatar/Avatar.jsx
```diff
@@ -95,26 +102,62 @@ const Avatar = ({
...
+        showTrigger={[Dialog.hideShowTriggers.FOCUS, Dialog.hideShowTriggers.MOUSE_ENTER]}
+        hideTrigger={[Dialog.hideShowTriggers.BLUR, Dialog.hideShowTriggers.MOUSE_LEAVE]}
+        {...tooltipProps}
+      >
+        {children}
+      </Tooltip>
+    );
+  };
+
+  const ClickableContainer = ({ children }) => {
```
hadasfa, line 121: same comment as in the tooltip container

src/components/Avatar/__stories__/avatar.stories.mdx
```diff
@@ -120,92 +143,131 @@ Use for non-person avatars, such as a workspace or team.
...
-        size={Avatar.sizes.LARGE} 
-        type={Avatar.types.IMG} 
-        src={person1} 
-        bottomLeftBadgeProps={{src: home}} 
+    <Avatar size={Avatar.sizes.LARGE} type={Avatar.types.IMG} src={person1} bottomLeftBadgeProps={{ src: home }} />
+    <Avatar size={Avatar.sizes.LARGE} type={Avatar.types.IMG} src={person1} bottomLeftBadgeProps={{ src: minus }} />
+  </Story>
+</Canvas>
+
+### Avatar with tooltip
```
hadasfa, line 222: please add here another example with aria label as tooltip content

and write explanation above each avatar with tooltip:
Avatar with textual representation as tooltip content
and Avatar with tooltip content which is different from it's textual representation

and please add aria label to both examples (because we do want developers always add aria label to their avatars)

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+import styles from "./AvatarGroup.module.scss";
+
+const AvatarGroup = ({
+  className,
+  id,
+  children,
+  size,
+  type,
+  max,
+  counterProps,
```
hadasfa, line 17: I don't think we want to add support on this customization - its different from tooltip because we do want all counter will look the same (in tooltip we maybe want to allow to open it from different places or allow different triggers for close it).
it also can allow people to create weird appearance for example change the color of the counter without change the wrapper.

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+  counterTooltipCustomProps,
```
hadasfa, line 18: why counterTooltipCustomProps and not counterTooltipProps, the 'custom' is redundant

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+  counterTooltipIsVirtualizedList,
```
hadasfa, line 19: right now we don't support it no? so can be removed. also when we will support it we need to find more simple name

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+  counterTooltipTheme
+}) => {
+  const {
+    color: counterColor = Counter.colors.LIGHT,
+    count: counterValue = children?.length && children.length - max,
+    prefix: counterPrefix = "+",
+    maxDigits: counterMaxDigits = 3
+  } = counterProps || {};
+
+  const getCounterContainerSizeStyle = useCallback(() => {
```
hadasfa, line 29: this pattern is not standard, let's talk about it?

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+    const counterSize = size || (children?.length && children[0]?.props?.size);
```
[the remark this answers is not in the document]
hadasfa: and then all the avatar will be in the same size

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+    if (counterSize && Avatar.sizes[`${counterSize.toString().toUpperCase()}`]) {
+      return styles[`${counterSize}`];
+    }
+    return styles.medium;
+  }, [children, size]);
+  const counterContainerSizeStyle = getCounterContainerSizeStyle();
+
+  const getCounterContainerColorStyle = useCallback(() => {
+    return styles[`${counterColor}`];
```
hadasfa, line 39: this is not a standard way to implement separate logic. let's talk about it.

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+  }, [counterColor]);
+  const counterContainerColorStyle = getCounterContainerColorStyle();
+
+  if (!children) {
+    return null;
+  }
+
+  if (!Array.isArray(children) && typeof children === "object") {
+    children = [children];
```
hadasfa, line 48: I am sure that there is a way to check if it's react element (optional, not must)

hadasfa: this can be helpful: React.Children.toArray

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  return (
+    <div className={cx(styles.container, className)} id={id}>
+      {children.map((avatar, index) => {
+        if (index < max) {
+          return (
+            <Avatar
+              {...avatar.props}
```
hadasfa, line 57: it's not very standard. u can use:  React.cloneElement for prevent weird cases when passing a component which wrap an avatar component or something like this.

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+              size={size || avatar.props?.size}
+              type={type || avatar.props?.type}
+              // TODO normal key
+              key={`${id}-${index}`}
+              className={cx(styles.avatarContainer, avatar.props?.className)}
```
hadasfa, line 62: why not add a avatarClassName props?

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+              withoutBorder
+            />
+          );
+        }
+      })}
+      {(children.length > max || !!counterValue) && (
+        <AvatarGroupCounterTooltipContainer
```
hadasfa, line 69: Let's create AvatarCounter component and move all the counter logic and ui to there? it will be more readable

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,144 @@
...
+          avatars={children}
```
hadasfa, line 70: it will be more readable and simple if the counter will receive only the avatars that should be present in the counter

src/components/AvatarGroup/AvatarGroup.module.scss
```diff
@@ -0,0 +1,66 @@
+@import "src/components/Avatar/Avatar.scss";
+@import "../../styles/themes";
+
+.container {
```
hadasfa, line 4: would change the class name to avatarGroup

src/components/AvatarGroup/AvatarGroup.module.scss
```diff
@@ -0,0 +1,66 @@
...
+  display: flex;
+  flex-direction: row;
+  flex-wrap: wrap;
+  position: relative;
+}
+
+.avatarContainer {
+  margin-left: -8px;
```
hadasfa, line 13: I think we have now variables for negative spacing. can you please check it up with Yonatan ben ari?

src/components/AvatarGroup/AvatarGroup.module.scss
```diff
@@ -0,0 +1,66 @@
...
+
+.counterContainer {
+  background: var(--primary-background-hover-color);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-left: -8px;
+  z-index: 1;
+
+  &:focus {
```
hadasfa, line 30: I think u looking for focus-visible. because if the focus is not visible we don't want to display the ring. please see hoe we implement this in other components and I will explain it to you when we will talk (there is a reason why using a class name + :focus-visible)

src/components/AvatarGroup/AvatarGroup.module.scss
```diff
@@ -0,0 +1,66 @@
...
+    outline: none;
+    @include focus-style(50%);
+  }
+
+  &.small {
+    height: $avatar-size-small;
```
hadasfa, line 36: let's create a css mixin for it and use it here (for prevent a copy paste of the same code)

src/components/AvatarGroup/AvatarGroup.module.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  &.large {
+    height: $avatar-size-large;
+    width: $avatar-size-large;
+    border-radius: $avatar-size-large / 2;
+  }
+
+  &.primary {
+    @include theme-prop(color, text-color-on-primary);
```
hadasfa, line 54: please convert to use var(--text-color-on-primary), we are not using theme-prop anymore

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
```diff
@@ -0,0 +1,88 @@
...
+  avatars,
+  max,
+  type,
+  className,
+  counterTooltipCustomProps,
+  counterTooltipIsVirtualizedList,
+  counterTooltipTheme
+}) => {
+  const counterTooltipContentComponent = useMemo(() => {
+    return AvatarGroupCounterTooltipContent({
```
hadasfa, line 19: that's not really standard way to write a component in jsx files. you can write the component with regular `<Component/>` syntax and no need to use useMemo here

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
```diff
@@ -0,0 +1,88 @@
...
+      children: avatars,
+      max,
+      type,
+      className,
+      isVirtualizedList: counterTooltipIsVirtualizedList
+    });
+  }, [avatars, className, max, counterTooltipIsVirtualizedList, type]);
+
+  if (!counterTooltipContentComponent && !counterTooltipCustomProps) {
+    return <>{children}</>;
```
hadasfa: `<>` is unnecessary, you can return children;

src/components/AvatarGroup/AvatarGroup.jsx
SergeyRoyt: Are you talking about `counterTooltipIsVirtualizedList`? We do support this, but only for auto-generated tooltip. I've added example in the storybook
We can discuss a better name

src/components/Avatar/__stories__/avatar.stories.mdx
SergeyRoyt: Not 100% clear, let's discuss

src/components/AvatarGroup/AvatarGroup.jsx
SergeyRoyt: With `counterProps` I've replaced 4 props: `counterValue`, `counterColor`, `counterPrefix`, `counterMaxDigits`. I thought it would be better this way due to similarity with `counterTooltipCustomProps` and fewer amount of props
But for now only these 4 props are actually passing into Counter, the rest is not used (though it might be a bit misguiding)

Didn't get the example with counter color change without wrapper change, let's discuss

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
SergeyRoyt: I'm using such syntax in order to obtain `null` from the `AvatarGroupCounterTooltipContent` (u can't get null from `<Component/>`), specifically in case when there is no avatars, so tooltip content couldn't be rendered. The other solution would be to just check avatars length at the `AvatarGroupCounterTooltipContainer, but I'm not sure that it'll be better there - what do you think?

src/components/AvatarGroup/AvatarGroup.module.scss
SergeyRoyt: We don't have them, Yonatan told me that you and Orr had some concerns about it
Does this look ok `margin-left: calc(-1 * var(--spacing-small));` ?

src/components/Avatar/Avatar.jsx
```diff
@@ -21,8 +24,10 @@ const Avatar = ({
   src,
   icon,
   text,
-  role,
+  tooltipProps,
   ariaLabel,
+  withoutTooltip,
```
hadasfa, line 29: yay!

src/components/Avatar/Avatar.jsx
```diff
@@ -38,7 +43,8 @@ const Avatar = ({
   bottomRightBadgeProps,
   withoutBorder,
   customSize,
-  customBackgroundColor
+  customBackgroundColor,
```
hadasfa, line 46: this prop should not be here, if it for making all the avatar be in the same color let's talk about it and find other solution

src/components/Avatar/Avatar.jsx
hadasfa: we talk yesterday about ariaLabel not need to take their value from the tooltip content in any case

src/components/Avatar/Avatar.jsx
```diff
@@ -47,9 +53,16 @@ const Avatar = ({
...
   const sizeStyle = useMemo(() => {
-    return customSize ? { height: `${customSize}px`, width: `${customSize}px` } : {};
+    return customSize ? { height: customSize, width: customSize } : {};
   }, [customSize]);
 
+  ariaLabel = ariaLabel || (typeof tooltipProps?.content === "string" ? tooltipProps.content : undefined);
+  if (withoutTooltip) {
+    tooltipProps?.content && delete tooltipProps.content;
+  } else if (tooltipProps?.content === undefined && ariaLabel) {
+    tooltipProps = { ...tooltipProps, content: ariaLabel };
```
hadasfa, line 63: what if tooltip props is undefined?

src/components/Avatar/Avatar.jsx
```diff
@@ -47,9 +53,16 @@ const Avatar = ({
     return src ? {} : { backgroundColor: getElementColor(backgroundColor) };
   }, [src, backgroundColor, customBackgroundColor]);
   const sizeStyle = useMemo(() => {
-    return customSize ? { height: `${customSize}px`, width: `${customSize}px` } : {};
+    return customSize ? { height: customSize, width: customSize } : {};
   }, [customSize]);
 
+  ariaLabel = ariaLabel || (typeof tooltipProps?.content === "string" ? tooltipProps.content : undefined);
+  if (withoutTooltip) {
+    tooltipProps?.content && delete tooltipProps.content;
```
hadasfa, line 61: I understand why u did it but it is a bit weird,  I think you can just do something like this:
```suggestion
    const overrideTooltipProps = useMemo(() => {
    if (withoutTooltip) return undefined;
    else if (tooltipProps) {
     return {content: ariaLabel, ...tooltipProps};
   }
   else {
     return {content: ariaLabel};
   }
    }, [ariaLabel, tooltipProps]);
```

hadasfa: put it in use memo for to be more organized and because we creating a new object... but can also be outside of use memo

src/components/Avatar/__stories__/avatar.stories.mdx
```diff
@@ -130,92 +178,141 @@ Use for non-person avatars, such as a workspace or team.
...
+
+### Clickable avatar
+
+Use to manage callback on onClick Avatar event.
+
+<Canvas>
+  <Story name="Clickable avatar">
+    {() => {
+      const [count, setCount] = useState(0);
+      const incrementCount = e => {
```
hadasfa, line 297: please use here in use callback

src/components/Avatar/__stories__/avatar.stories.mdx
```diff
@@ -130,92 +178,141 @@ Use for non-person avatars, such as a workspace or team.
...
+        </Flex>
+      );
+    }}
+  </Story>
+</Canvas>
 
 ## Related components
-<RelatedComponents componentsNames={[TOOLTIP, LABEL, CHIP]}/>
+
+<RelatedComponents componentsNames={[AVATAR_GROUP, TOOLTIP, LABEL]} />
```
hadasfa, line 318: yay!

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,88 @@
...
+const AvatarGroup = ({
+  className,
+  avatarClassName,
+  id,
+  children,
+  size,
+  type,
+  max,
+  counterProps,
+  counterTooltipCustomProps
```
hadasfa, line 19: counterTooltipCustomProps => counterTooltiProps ?

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,88 @@
...
+}) => {
+  if (!children) {
+    return null;
+  }
+
+  children = React.Children.toArray(children);
```
hadasfa, line 25: it's weird that esline don't give u error for this. it's not standard to override prop value under the same name.
```suggestion
  const overrideChildren = React.Children.toArray(children);
```

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -0,0 +1,88 @@
...
+  const displayAvatars = children.slice(0, max);
```
hadasfa, line 26: this for example need to be in use memo because you want to iterate all the children for slicing only if there is a change on the children

src/components/Avatar/Avatar.jsx
SergeyRoyt: I haven't add this prop, not sure where it comes from, probably from merge with master

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -0,0 +1,73 @@
+import React from "react";
+import Tooltip from "../Tooltip/Tooltip";
+import Avatar from "../Avatar/Avatar";
+import PropTypes from "prop-types";
+import AvatarGroupCounterTooltipContainer from "./AvatarGroupCounterTooltipContainer";
+import cx from "classnames";
```
hadasfa, line 6: imports for package always before local imports

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -0,0 +1,73 @@
...
+import Counter from "../Counter/Counter";
+import styles from "./AvatarGroupCounter.module.scss";
+
+const AvatarGroupCounter = ({
+  counterTooltipAvatars,
+  counterProps,
+  counterTooltipCustomProps,
+  avatarSize,
+  avatarType
```
hadasfa, line 15: why we need type in the level of the counter tooltip?

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -0,0 +1,73 @@
...
+}) => {
+  const {
+    color: counterColor = Counter.colors.LIGHT,
+    count: counterValue = counterTooltipAvatars.length,
```
hadasfa, line 19: its a bit confusing and I am not sure we need it.... count is a data in counter props I am not understand...

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -0,0 +1,73 @@
...
+    prefix: counterPrefix = "+",
+    maxDigits: counterMaxDigits = 3
+  } = counterProps || {};
+
+  const counterSizeStyle = styles[avatarSize?.toString()];
+  const counterColorStyle = styles[counterColor];
+
+  if (!counterTooltipAvatars.length && !counterValue) {
```
hadasfa, line 27:
```suggestion
  if (counterTooltipAvatars === undefined || counterTooltipAvatars.length === 0) {
```

counterValue is just the length of counterTooltipAvatars or I did not understand?

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+  return (
+    <AvatarGroupCounterTooltipContainer
+      avatars={counterTooltipAvatars}
+      counterTooltipCustomProps={counterTooltipCustomProps}
+      type={avatarType}
+    >
+      {/* eslint-disable jsx-a11y/no-noninteractive-tabindex */}
+      <div tabIndex={0} className={cx(styles.counterContainer, counterSizeStyle, counterColorStyle)}>
+        {/* eslint-enable jsx-a11y/no-noninteractive-tabindex */}
```
hadasfa, line 39: I think we can delete this comment no?

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+AvatarGroupCounter.propTypes = {
+  /**
+   * Array of Avatar elements
+   */
+  counterTooltipAvatars: PropTypes.arrayOf(PropTypes.element),
+  /**
+   * Counter.propTypes: props for counter
+   */
+  counterProps: PropTypes.shape({ ...Counter.propTypes }),
```
hadasfa, line 60:
```suggestion
  counterProps: PropTypes.shape(Counter.propTypes),
```

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -0,0 +1,73 @@
...
+  counterTooltipCustomProps: PropTypes.shape({ ...Tooltip.propTypes }),
```
hadasfa, line 61:
```suggestion
  counterTooltipCustomProps: PropTypes.shape(Tooltip.propTypes),
```

src/components/AvatarGroup/AvatarGroupCounter.module.scss
```diff
@@ -0,0 +1,48 @@
...
+  background: var(--primary-background-hover-color);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  margin-left: calc(-1 * var(--spacing-small));
+  z-index: 1;
+
+  &:focus-visible,
+  &.focus-visible {
+    outline: none;
```
hadasfa, line 13: yay

src/components/AvatarGroup/AvatarGroup.jsx
SergeyRoyt: I just want people to understand that there are default and custom tooltip content. And when I add counterTooltipIsVirtulizedList back we will need to explain the scope of each prop. If I'll remove "Custom" from the  counterTooltipCustomProps, then I would add "Default" to virtualized prop and that's probably not a good idea :)

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
```diff
@@ -0,0 +1,66 @@
...
+import PropTypes from "prop-types";
+
+const AvatarGroupCounterTooltipContainer = ({ children, avatars, type, className, counterTooltipCustomProps }) => {
+  const counterTooltipContentExists = !!avatars?.length;
+  const counterTooltipContentComponent = (
+    <AvatarGroupCounterTooltipContent avatars={avatars} type={type} className={className} />
+  );
+
+  if (!counterTooltipContentExists && !counterTooltipCustomProps) {
+    return children;
```
hadasfa, line 15: this check is not needed no? if we implement what I wrote on top of the page and...  if we don'y have counterTooltipCustomProps we still want to present the tooltip no?

src/components/AvatarGroup/AvatarGroup.jsx
SergeyRoyt: Should displayAvatars and counterTooltipAvatars also be memoized?

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  if (counterTooltipCustomProps) {
```
hadasfa, line 18: no need for if else, you can just write in the tooltip props like this:
{...counterTooltipCustomProps} and set default as {}
and regarding content do something like this: 
content={counterTooltipCustomProps?.content || counterTooltipContentComponent}

src/components/AvatarGroup/AvatarGroupCounter.jsx
SergeyRoyt: Need the comment to prevent eslint to complain about tabIndex on noninteractive element

src/components/AvatarGroup/AvatarGroupCounter.jsx
SergeyRoyt: count belongs to Counter props - it's the prop for setting counter value. We give ability to set custom counter value, Evgeniy asked for this. It's standard prop for such components, see `total` in MUI https://mui.com/material-ui/api/avatar-group/
About the name - I also don't like "count" too much, but that's how the prop is called in Counter, so don't want to confuse users with renaming it

src/components/AvatarGroup/AvatarGroupCounter.jsx
SergeyRoyt: Cause we are passing afterwards to avatars in tooltip, like this 
<Avatar {...avatarProps} ... type={type || avatarProps?.type} />
And we do the same for displaying avatars in AvatarGroup
This could be used to set type for all avatars at once - at AvatarGroup prop, not at every single avatar you are passing

But we can get rid of it, if u think its irrelevant

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
```diff
@@ -0,0 +1,66 @@
...
+      {children}
+    </Tooltip>
+  );
+};
+
+AvatarGroupCounterTooltipContainer.propTypes = {
+  className: PropTypes.string,
+  type: PropTypes.oneOf([Avatar.types.TEXT, Avatar.types.ICON, Avatar.types.IMG]),
+  /**
+   * Counter element
```
hadasfa, line 49:
```suggestion
   * Avatar Counter element
```

because it is not a regular counter element right?

src/components/AvatarGroup/AvatarGroupCounter.jsx
SergeyRoyt: counterValue = count || counterTooltipAvatars.length
where `count` is prop passed by user - value to be displayed in `Counter`

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
```diff
@@ -0,0 +1,66 @@
...
+   */
+  children: PropTypes.element,
+  /**
+   * Array of Avatar elements
+   */
+  avatars: PropTypes.arrayOf(PropTypes.element),
+  counterTooltipCustomProps: PropTypes.shape({ ...Tooltip.propTypes })
```
hadasfa, line 56: counterTooltipCustomProps => counterTooltipProps ?

hadasfa:
```suggestion
  counterTooltipCustomProps: PropTypes.shape(Tooltip.propTypes)
```

src/components/AvatarGroup/AvatarGroupCounterTooltipContent.jsx
```diff
@@ -0,0 +1,102 @@
...
+import PropTypes from "prop-types";
+import styles from "./AvatarGroupCounterTooltipContent.module.scss";
+
+const AvatarGroupCounterTooltipContent = ({ avatars, type, className }) => {
+  const getTooltipContent = avatarProps => {
+    return avatarProps?.tooltipProps?.content || avatarProps?.ariaLabel;
+  };
+
+  avatars = avatars.map(avatar => ({ value: avatar.props }));
+  const displayAsGrid = !avatars.some(item => getTooltipContent(item.value));
```
hadasfa, line 15: why to display as grid?
some logic should be under use memo

src/components/AvatarGroup/AvatarGroupCounterTooltipContent.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const avatarRenderer = (item, index, style) => {
+    const avatarProps = item.value;
```
hadasfa, line 18: why another avatar renderer? there is one in the avatar component. let's talk about it?

hadasfa: also this is a problematic pattern

src/components/AvatarGroup/__stories__/AvatarGroup.stories.scss
```diff
@@ -0,0 +1,8 @@
+@import "../../../styles/themes.scss";
+
+.monday-storybook-avatar-group {
```
hadasfa, line 3: please convert to css modules

src/components/AvatarGroup/AvatarGroupCounterTooltipContainer.jsx
SergeyRoyt: I've refactored it, but in a different way, check it out
If counterTooltipCustomProps have been passed even without content inside we don't want to display counterTooltipContentComponent and don't want it to be rendered

src/components/AvatarGroup/AvatarGroupCounterTooltipContent.jsx
SergeyRoyt: If there is no tooltip text we display avatars as grid - Evgeniy asked for it
image
I'll use memo here

src/components/AvatarGroup/AvatarGroupCounterTooltipContent.jsx
SergeyRoyt: It renders not avatar only avatar itself, but kinda avatarRowItem
Let's talk

src/components/Dialog/Dialog.jsx
```diff
@@ -4,13 +4,13 @@ import { DIALOG_ANIMATION_TYPES } from "constants/AnimationTypes";
 import React, { PureComponent } from "react";
 import { createPortal } from "react-dom";
 import PropTypes from "prop-types";
-import { Manager, Reference, Popper } from "react-popper";
+import { Manager, Popper, Reference } from "react-popper";
```
hadasfa, line 7: please revert

src/components/Counter/Counter.scss
```diff
@@ -81,6 +81,10 @@
             @include theme-prop(color, text-color-on-primary);
             @include theme-prop(background-color, negative-color);
           }
+          &-light {
+            @include theme-prop(color, text-color);
```
hadasfa, line 85: replace with var(--variable-name). because this is the new convention.

src/components/Tooltip/Tooltip.jsx
```diff
@@ -128,7 +140,10 @@ export default class Tooltip extends React.PureComponent {
       animationType: "expand",
       onDialogDidHide: this.onTooltipHide,
       onDialogDidShow: this.onTooltipShow,
-      getDynamicShowDelay: this.getShowDelay
+      getDynamicShowDelay: this.getShowDelay,
+      showTrigger,
```
hadasfa, line 144: why?

src/components/Tooltip/Tooltip.jsx
```diff
@@ -138,26 +153,6 @@ Tooltip.positions = DialogPositions;
 Tooltip.animationTypes = DIALOG_ANIMATION_TYPES;
 Tooltip.themes = TOOLTIP_THEMES;
 Tooltip.justifyTypes = TOOLTIP_JUSTIFY_TYPES;
-Tooltip.defaultProps = {
```
hadasfa, line 141: why?

SergeyRoyt: I think I've just change places of propTypes and defaultProps, cause they've been reversed in comparison to every other component

hadasfa: I did not found it in the file in a different place, if it's exist please ignore this comment

src/components/AvatarGroup/AvatarGroupCounterTooltipContent.jsx
hadasfa: let's separate AvatarGroupCounterTooltipContent to 2 components for make it more clear:
AvatarGroupCounterTooltipContent and AvatarGroupCounterTooltipGridContent. (there is enough differences between these 2 components).
In the parent component use useMemo that go over all the avatars on the counter and return 2 
thins: isGrid (if some avatar does not include value) and array of all getTooltipContent
We are using use memo for not iterate all the avatars, only when the array instance reference changed.

hadasfa (approved):
I approved but wrote a comment about how to fix the last thing we talked about in our zoom talk. Let's fix it and make sure the tests still pass before we merge ? I think that after all the refactor u did the pr is much more clearer, great work :)

PR #696, opened 2022-06-13
Modal

eran-cohen:
Adding an accessible modal component to Vibe:
See mockups here: [https://www.figma.com/file/wEHsnyvBoWtOJ60J5jArWC/modal-Component](https://www.figma.com/file/wEHsnyvBoWtOJ60J5jArWC/modal-Component)

based on [a11y-dialog](https://a11y-dialog.netlify.app/)

Stroybook pages are not fully ready yet, will be glad for some guidence here


src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,219 @@
...
+  /**
+   * Show/hide the Dialog
+   */
+  show: PropTypes.bool.isRequired,
+  /**
+   * Title for the modal, mandatory when ModalHeader isn't provided in children
+   */
+  title: function (props, propName, componentName) {
+    const { children, title } = props;
+    const hasHeaderComponent = React.Children.toArray(children).some(isModalHeader);
```
niksa-monday, line 145: Like it!

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,48 @@
...
+}
+
+.dialog {
+  z-index: 2;
+  display: flex;
+  position: relative;
+  flex-direction: column;
+  padding: 0 30px;
+  padding-block-end: 24px;
+  border-radius: 16px;
```
niksa-monday, line 35: Probably we should use design-tokens here:
https://github.com/mondaycom/monday-ui-style/blob/master/src/config/tokens/core/_core-radius.scss

src/components/Modal/__tests__/modal-snapshot-tests.jest.js
```diff
@@ -0,0 +1,58 @@
...
+        <Modal {...baseProps}>
+          {content}
+          <ModalFooter>{footerContent}</ModalFooter>
+        </Modal>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it("with all parts", () => {
```
niksa-monday, line 46: I think we need to check all/most props permutations:
 - classNames
 - title
 - descriptions
 - etc. - all props
 I really recommend to use snapshotDiff for decrease snapshot size and improve readability of snapshots.

src/components/Modal/__tests__/modal-tests.jest.js
```diff
@@ -0,0 +1,175 @@
...
+  describe("rendering", () => {
+    it("should render a title if the title prop is provided", () => {
+      const TEST_TITLE = "test title";
+      const component = renderComponent({ openOnStart: true, title: TEST_TITLE });
+      const title = component.queryByText(TEST_TITLE);
+      expect(title).toBeInTheDocument();
+    });
+
+    it("should render ModalHeader if passed as children", () => {
+      const TEST_TITLE = "test title";
```
niksa-monday, line 156: I'm not sure - but It's looks like duplication for Snapshot tests ...

src/components/Modal/a11YDialog.jsx
```diff
@@ -0,0 +1,62 @@
+import React from "react";
```
niksa-monday, line 1: Name of file should be: `a11yDialog.jsx` instad of `a11YDialog.jsx`

src/components/ModalContent/__stories__/ModalContent.stories.mdx
```diff
@@ -0,0 +1,144 @@
...
+import { createComponentTemplate, createStoryMetaSettings } from "../../../storybook/functions/create-component-story";
+
+export const metaSettings = createStoryMetaSettings({
+  component: ModalContent,
+  enumPropNamesArray: [], // List enum props here
+  iconPropNamesArray: [], // List props that are typed as icons here
+  actionPropsArray: [], // List the component's actions here
+});
+
+<Meta
```
niksa-monday, line 12: I'm not sure we need separate Stories for each SubConponent.
You can check [story of Accordion](https://62037c72e0e4d6004a9a450f-iobjldxeex.chromatic.com/?path=/docs/data-display-accordion--overview) - where you have example of Tabs for sub-item props in single story:

Screen Shot 2022-06-26 at 12 34 22

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+            <div>Modal content goes here</div>
+            <ModalFooter>{modalButtons}</ModalFooter>
+          </Modal>
+        </div>
+      );
+    }}
+  </Story>
+</Canvas>
+
+## Related components
```
niksa-monday, line 306: I think we can add:
Popover, Tooltip, Alert as related components.

And what about Do's and Don'ts ?

src/components/ModalFooter/__tests__/modalFooter-snapshot-tests.jest.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import ModalFooter from "../ModalFooter";
+import { Button } from "components";
+
+describe("ModalFooter renders correctly", () => {
+  it("with empty props", () => {
+    const tree = renderer
```
niksa-monday, line 8: I think here we also need props permutation check (i.e. with className)

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+## Variants
+
+### Alert Modal (Modal)
+
+This Modal will NOT be closed when you click outside or click ESC key. Only way to close it is to use the close button.
+
+<UsageGuidelines guidelines={["set `isAlertDialog` to `true` in order to turn a Modal to modal."]} />
+
+<Canvas>
+  <Story name="Alert Modal">
```
niksa-monday, line 87: Story stop working for me when I open modal - close with Esc - and then try to open once again - Modal not opening.

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+### Modal with vertical scroll
+
+The height of Modal with long content will be limited to 90% of screen height, and a scroll bar will be added
+
+<Canvas>
+  <Story name="Modal with vertical scroll">
+    {() => {
+      const { openModalButton, modalButtons, show, closeModal, ref } = useModalHelper({
+        openButtonTitle: "Default"
```
niksa-monday, line 216: It's better to change name of button "Default" - to something more related to story: "Vertical scroll"

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+
+### Modal with custom Header
+
+Header with description and icon
+
+<Canvas>
+  <Story name="Modal with custom Header">
+    {() => {
+      const { openModalButton, modalButtons, show, closeModal, ref } = useModalHelper({
+        openButtonTitle: "Default"
```
niksa-monday, line 283: It's better to change name of button "Default" - to something more related to story: "Custom Header"

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+});
+
+<Meta title="Feedback/Modal" component={Modal} argTypes={metaSettings.argTypes} decorators={metaSettings.decorators} />
+
+<!--- Component documentation -->
+
+# Modal
+
+- [Overview](#overview)
+- [Props](#props)
```
niksa-monday, line 23: I think we have an issue with controls settings:
Screen Shot 2022-06-26 at 12 34 22

Please check other stories for setups it - in you case I think Accordion will be good example since it also have sub-components as Modal.

src/components/Modal/Modal.module.scss
SergeyRoyt: `var(--spacing-large)` would be even better

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+- [Use cases and examples](#use-cases-and-examples)
+- [Related components](#related-components)
+- [Feedback](#feedback)
+
+## Overview
+
+An easy to use accessible Modal component
+
+<Canvas>
+  <Story name="Overview">
```
SergeyRoyt, line 36: This specific modal button doesn't reopen modal once it was closed
Steps to reproduce: 
1) Focus button via Tab
2) Press Space to open Modal
3) Press Esc to close Modal
4) Result: modal doesn't opens again

src/components/Modal/__stories__/helpers.js
```diff
@@ -0,0 +1,39 @@
...
+import React, { useRef, useState } from "react";
+
+// internal custom hook to help with writing tests and stories.
+export const useModalHelper = ({ openButtonTitle = "Open", openOnStart = false, ...props } = {}) => {
+  const [show, setShow] = useState(openOnStart);
+  const buttonRef = useRef(null);
+  const closeModal = () => setShow(false);
+
+  const openModalButton = (
+    <Button onClick={() => setShow(true)} ref={buttonRef}>
```
SergeyRoyt, line 11: I would have added `ariaHasPopup` here for voice over a11y

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+            size={Modal.Size.DEFAULT}
+          >
+            <p>Normal modal</p>
+            <ModalFooter>{modalButtons}</ModalFooter>
+          </Modal>
+        </div>
+      );
+    }}
+  </Story>
+  <Story name="Size variants - Full">
```
SergeyRoyt, line 180: Usually such story is united and called "Size variants"

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,308 @@
...
+            title={"Modal title"}
+            triggerElement={ref.current}
+            show={show}
+            onClose={closeModal}
+            isAlertDialog
+            closeButtonAriaLabel={"close"}
+          >
+            <ModalContent>
+              <form>
+                <TextField title="Your Name" autoFocus size={TextField.sizes.MEDIUM} placeholder="John Doe" />
```
SergeyRoyt, line 138: I would duplicate `inputAriaLabel` from `title` or from `title` + `placeholder`.
In fact I've seen that in TextField `ariaLabel = inputAriaLabel || placeholder`.  I think it's a strange implementation, cause voice over will always pronounce placeholder even when it was cleared and never pronounce the title, which might give more understanding. Maybe `title` should be somewhere there too, what do you think? Smth like  `ariaLabel = inputAriaLabel || title` or `ariaLabel = inputAriaLabel || "${titile}, ${placeholder}"`?

src/components/ModalHeader/ModalHeader.module.scss
```diff
@@ -0,0 +1,33 @@
+@import "../../styles/themes";
+
+.mondayStyleModalHeader {
+  display: flex;
+  align-items: start;
+  flex-direction: column;
+  margin-block-end: 24px;
```
SergeyRoyt, line 7: `var(--spacing-large)`

src/components/ModalHeader/ModalHeader.jsx
```diff
@@ -0,0 +1,102 @@
...
+  icon,
+  iconSize,
+  iconClassName,
+  hideCloseButton,
+  closeButtonAriaLabel,
+  attr = {}
+}) => {
+  return (
+    <div className={cx(styles.mondayStyleModalHeader, className)}>
+      <p {...attr.title} className={cx(titleClassName, styles.title)}>
```
SergeyRoyt, line 24: What is `attr`? Why it's not in the propTypes? 
Why title comes from there and not from `title`?

src/components/Modal/__stories__/modal.stories.mdx
eran-cohen: fixed, thanks!

src/components/ModalHeader/ModalHeader.jsx
eran-cohen: attr is an object with various aria related attributes a11ydialog provides. attr.title are the attributes it provides for title. But I checked again and only the id is actually dynamic so I refactored it and ModalHeader only get an id

package.json
```diff
@@ -50,7 +50,9 @@
     "@popperjs/core": "2.11.5",
     "@react-hook/merged-ref": "1.3.2",
     "@react-stately/toggle": "3.2.7",
+    "a11y-dialog": "^7.5.0",
     "autosize": "^5.0.1",
+    "body-scroll-lock": "^4.0.0-beta.0",
```
laviomri, line 55: are we ok with using the beta version?

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+  const closeDialogIfNeeded = useCallback(() => {
+    if (!isAlertDialog) {
+      onClose();
+    }
+  }, [isAlertDialog, onClose]);
+
+  const getAnimationProps = useAnimationProps(triggerElement, instance);
+
+  // clear all scroll locks on unmount (just to be safe)
+  useEffect(() => {
```
laviomri, line 60: it might be a matter of preference, but consider extracting these 3 `useEffect`s to a separate hook (something like `useLockedBodyScroll` or whatever)

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+      } else {
+        if (animate) {
+          const animation = instance?.$el.childNodes[1]?.animate(...getAnimationProps(true));
+          animation?.finished.then(() => instance?.hide());
+        } else {
+          instance?.hide();
+        }
+      }
+    }
+  }, [show, instance, getAnimationProps]);
```
laviomri, line 97: I think you can improve readability a bit with the same functionality, WDYT?

```suggestion
  if(!instance) {
    return;
  }
  const animate = instance.$el.childNodes[1].animate;
  if (show) {
    instance?.show();
    animate?.(...getAnimationProps());
  } else if (animate) {
    const animation = animate(...getAnimationProps(true));
    animation?.finished.then(() => instance?.hide());
  } else {
    instance.hide();
  }
  }, [show, instance, getAnimationProps]);
```

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+
+  // const hideModal = useCallback(() => {
```
laviomri, line 99: remove?

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Show/hide the Dialog
+   */
+  show: PropTypes.bool.isRequired,
+  /**
+   * Title for the modal, mandatory when ModalHeader isn't provided in children
+   */
+  title: function (props, propName, componentName) {
+    const { children, title } = props;
+    const hasHeaderComponent = React.Children.toArray(children).some(isModalHeader);
```
laviomri, line 180: nice!
is it worth it to maybe make this use the same logic as the [`header` const](https://github.com/mondaycom/monday-ui-react-core/pull/696/files#diff-1a35dbb21807dc16cc9af88fe302ff530a4426014b076ecbb8868e2b14bb773fR105)?

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,47 @@
+@import "../../styles/themes";
+
+.container,
+.overlay {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
```
laviomri, line 9: can you use [`inset`](https://css-tricks.com/almanac/properties/i/inset/)?
```suggestion
  inset: 0;
```

src/components/Modal/Modal.module.scss
```diff
@@ -0,0 +1,47 @@
...
+  justify-content: center;
+  align-items: center;
+}
+
+.container[aria-hidden="true"] {
+  display: none;
+}
+
+.overlay {
+  background-color: rgba(41, 47, 76, 0.7);
```
laviomri, line 24: any variable you can use?

package.json
hadasfa: +1

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+  DEFAULT: "default",
+  FULL_WIDTH: "full_width"
+};
+
+const isModalHeader = child => child.type === ModalHeader;
+const isModalContent = child => child.type === ModalContent;
+const isModalFooter = child => child.type === ModalFooter;
+
+const Modal = ({
+  classNames,
```
hadasfa, line 22: we are a bit strict about maintaining uniformity between the api of all the props of all the components.
Please rename classNames to className.

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+  id,
+  show,
+  title,
+  description,
+  onClose,
+  isAlertDialog,
```
hadasfa, line 28: We are a bit strict about maintaining uniformity between the api of all the props of all the components. We do not used in isPrefix while naming props (because of React shorthand support for boolean props).
please change isAlertDialog to alertDialog.

src/components/Modal/Modal.jsx
hadasfa: I agree. I really like to use custom hooks for separate different logic part which are not related or depends on each-other for readability. 
But it's not a blocker ofcourse.

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+      instance?.off("hide");
+    };
+  }, [instance]);
+
+  useEffect(() => {
+    instance?.on("hide", () => onClose());
+    return () => {
+      instance?.off("hide");
+    };
+  }, [instance]);
```
hadasfa, line 79: please fix linter error :)

src/components/Modal/Modal.jsx
hadasfa: +1

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+      <ModalHeader
+        title={title}
+        description={description}
+        closeModal={onClose}
+        id={id}
+        hideCloseButton={hideCloseButton}
+        closeButtonAriaLabel={closeButtonAriaLabel}
+      />
+    );
+  }, [children, attr, title, description, hideCloseButton, closeButtonAriaLabel]);
```
hadasfa, line 120: please fix linter error :)

src/components/Modal/Modal.jsx
```diff
@@ -0,0 +1,254 @@
...
+      >
+        {header}
+        {content}
+        {footer}
+      </div>
+    </div>,
+    document.body
+  );
+
+  return <>{dialog}</>;
```
hadasfa, line 163:
```suggestion
  return dialog;
```

hadasfa: I think we can remove this <> </> no?

src/components/Modal/Modal.module.scss
hadasfa: Why we did not use a css variable here? because we needed to add opacity?

src/components/Modal/__stories__/helpers.js
```diff
@@ -0,0 +1,39 @@
+import { Button, Flex } from "components";
+import React, { useRef, useState } from "react";
+
+// internal custom hook to help with writing tests and stories.
+export const useModalHelper = ({ openButtonTitle = "Open", openOnStart = false, ...props } = {}) => {
```
hadasfa, line 5: The problem with code inside helpers in stories is that the people who read the story can't see its implementation (and therefore they usually not understand what is this helper and where they can find it).
In my opinion even this code is reusable, the readability is more important in these cases (stories).  So it will be better to copy past this code to each story (and it will be helpful for developers which trying to understand who to use the modal)

src/components/Modal/__stories__/helpers.js
```diff
@@ -0,0 +1,39 @@
...
+    show,
+    onClose: closeModal,
+    triggerElement: buttonRef.current
+  };
+
+  return { modalProps, openModalButton, modalButtons, show, closeModal, ref: buttonRef };
+};
+
+export const ModalButtons = ({ closeModal }) => {
+  return (
```
hadasfa, line 29: I guess we usualy will prefer to locate the buttons of every modal in monday in the same location above the modal dialog and with the same gap (for consistency)
Why not create ModalButtonsContainer component for this? (will receive buttons (from vibe) as childrens

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -0,0 +1,384 @@
...
+- [Related components](#related-components)
+- [Feedback](#feedback)
+
+## Overview
+
+An easy to use accessible Modal component
+
+<Canvas>
+  <Story name="Overview">
+    {() => {
```
hadasfa, line 49: The reason that the controls are not working in this story is because you don't use the component template syntax. Please see how we implement the overview story in other components (for example button) as a reference

src/components/Modal/__stories__/modal.stories.mdx
hadasfa: I agree. Not sure we have vibe Popover component, but we do have dialog container component.

hadasfa: my suggestion: TOOLTIP, DIALOG_CONTAINER, Tipseen.

src/components/Modal/__tests__/modal-tests.jest.js
```diff
@@ -0,0 +1,175 @@
...
+};
+
+describe("Modal tests", () => {
+  afterEach(() => {
+    cleanup();
+  });
+
+  describe("open", () => {
+    it("should open when triggered", async () => {
+      const component = renderComponent({ openOnStart: false });
```
hadasfa, line 51: Maybe it is just me but it,s weird to me that we test the modal component with so many wrapper. this way things can work or not work only because of one of the 2 wrappers logic or defaults. For example I think it would more reliable to test if pass show to the modal component trigger showing the modal instead of if openOnStart trigger showing the modal. Because in this case if other developer create a bug he can fix the ModalManager/hook which both are not exported outside instead of the component.

src/components/Modal/__tests__/modal-tests.jest.js
```diff
@@ -0,0 +1,175 @@
...
+      const openButton = getOpenButton(component);
+      fireEvent.click(openButton);
+      expect(await findModal(component)).toBeInTheDocument();
+    });
+
+    it("should open on start if show=true", async () => {
+      const component = renderComponent({ openOnStart: true });
```
hadasfa, line 58: Maybe it is just me but it's weird to me that we test the modal component with so many wrapper. this way things can work or not work only because of one of the 2 wrappers logic or defaults. For example I think it would more reliable to test if pass show to the modal component trigger showing the modal instead of if openOnStart trigger showing the modal. Because in this case if other developer create a bug he can fix the ModalManager/hook which both are not exported outside instead of the component.

hadasfa: In cases which we are not testing the actual api  (for example "should close when clicking on close button") I think it's not an issue

src/components/Modal/__tests__/modal-tests.jest.js
```diff
@@ -0,0 +1,175 @@
...
+    it("should NOT close on Esc or overlay click if isAlertDialog = true", async () => {
+      const component = renderComponent({ openOnStart: true, isAlertDialog: true });
+      fireEvent.keyDown(queryAlertModal(component), { key: "Escape", code: "Escape" });
+      fireEvent.click(await findOverlay(component));
+
+      expect(queryAlertModal(component)).toBeInTheDocument();
+    });
+  });
+
+  describe("accessibility", () => {
```
hadasfa, line 92: Awesome!

src/components/Modal/__tests__/modal-tests.jest.js
hadasfa: +1

src/components/ModalFooter/__stories__/ModalFooter.stories.mdx
```diff
@@ -0,0 +1,144 @@
...
+- [Use cases and examples](#use-cases-and-examples)
+- [Related components](#related-components)
+- [Feedback](#feedback)
+
+## Overview
+Please write here the component description
+
+<Canvas>
+  <Story name="Overview"
+         args={ {prop1: "prop's default value", prop2: "prop's default value"} }>
```
hadasfa, line 40: I think that keeping the example args can be confusing and it will be better to delete these

src/components/ModalFooter/__stories__/ModalFooter.stories.mdx
```diff
@@ -0,0 +1,144 @@
...
+    { modalFooterTemplate.bind({}) }
+  </Story>
+</Canvas>
+
+## Props
+<ArgsTable of={ ModalFooter } />
+
+## Usage
+<UsageGuidelines guidelines={[
```
hadasfa, line 49: We want to create  story for each modal building block? If we do let's remove all plop template examples (such as empty stories and generic "Usage guideline 1")
If not let's delete it

src/components/Modal/a11yDialog.jsx
```diff
@@ -0,0 +1,60 @@
...
+const useA11yDialogInstance = () => {
+  const [instance, setInstance] = React.useState(null);
+  const container = React.useCallback(node => {
+    if (node !== null) setInstance(new A11yDialogLib(node));
+  }, []);
+
+  return [instance, container];
+};
+
+export const useA11yDialog = props => {
```
hadasfa, line 14: Sorry for the pettiness - is it a component or a hook? if it is a hook we usually do no use propTypes and we use  this structure for attributes ^^" 
```
export const useA11yDialog = {attribute1, attribute2} => {
```
But it is super cool! And we can also use it in other dialogs when we will do a refactor

src/components/ModalContent/ModalContent.module.scss
```diff
@@ -0,0 +1,10 @@
+@import "../../styles/themes";
+@import "../../styles/states";
+
+.mondayStyleModalContent {
```
hadasfa, line 4: we usually do not add monday prefix when using css modules as standard

src/components/ModalContent/__stories__/ModalContent.stories.mdx
```diff
@@ -0,0 +1,144 @@
...
+    }
+  ]}
+/>
+
+
+## Use cases and examples
+
+### Story title 3
+Description of story 3
+<Canvas>
```
hadasfa, line 127: Is this story needed? or we can delete it?

src/components/ModalHeader/ModalHeader.module.scss
```diff
@@ -0,0 +1,33 @@
+@import "../../styles/themes";
+
+.mondayStyleModalHeader {
```
hadasfa, line 3: same comment about monday prefix

src/components/index.js
```diff
@@ -77,3 +77,7 @@ export { default as IconButton } from "./IconButton/IconButton";
 export { default as Flex } from "./Flex/Flex";
 export { default as MenuGridItem } from "./Menu/MenuGridItem/MenuGridItem";
 export { default as AvatarGroup } from "./AvatarGroup/AvatarGroup";
+export { default as Modal } from "./Modal/Modal";
```
hadasfa, line 80: We need to add the components also to the published components file otherwise developers will not be able to access them.

package.json
eran-cohen: yes, that's the current version, that's what you get when you do `npm i body-scroll-lock`.

src/components/Modal/Modal.jsx
eran-cohen: done

src/components/Modal/Modal.jsx
eran-cohen: done

src/components/Modal/Modal.module.scss
eran-cohen: 👍

src/components/Modal/Modal.jsx
eran-cohen: 👍

src/components/Modal/Modal.jsx
eran-cohen: 👍

src/components/Modal/Modal.jsx
eran-cohen: 👍

src/components/Modal/Modal.jsx
eran-cohen: @hadasfa I thought so too, but when I remove it storybook stops showing the component props on the storybook page

src/components/Modal/a11yDialog.jsx
eran-cohen: 👍

src/components/ModalContent/__stories__/ModalContent.stories.mdx
eran-cohen: 👍

src/components/ModalHeader/ModalHeader.module.scss
eran-cohen: 👍

src/components/index.js
eran-cohen: 👍

src/components/ModalContent/ModalContent.module.scss
eran-cohen: 👍

src/components/ModalFooter/__stories__/ModalFooter.stories.mdx
eran-cohen: 👍

src/components/ModalFooter/__stories__/ModalFooter.stories.mdx
eran-cohen: 👍

src/components/Modal/Modal.jsx
eran-cohen: @laviomri  it's almost the same, only difference is I'm using `some` instead of `find` because we need a boolean here instead of the actual bject

PR #705, opened 2022-06-19
fix button tooltip hidden when stop hovering + tipseen useless wrapper

hadasfa:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


src/components/Tipseen/Tipseen.jsx
```diff
@@ -47,6 +47,7 @@ const Tipseen = forwardRef(
     const componentRef = useRef(null);
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
     const overrideCloseAriaLabel = closeAriaLabel || TIPSEEN_CLOSE_BUTTON_ARIA_LABEL;
+    const TipseenWrapper = ref || id ? "div" : Fragment;
```
hadasfa, line 50: I understand this wrapper is useless in general because the ref and the id should be on the tipseen and not on the content linked to it but I can't delete it because backwards compatibility

hadasfa: So I convert it to be conditional

orrgottlieb: @hadasfa cool, add the ability to pass an element type

orrgottlieb (approved):

PR #706, opened 2022-06-20
Feature/sergeyro/avatar group a11y

SergeyRoyt:
https://monday.monday.com/boards/245345663/pulses/2829818165
Accessibility improvements for AvatarGroup

#### Basic
- [X] Used plop (`npm run plop`) to create a new component.
- [X] PR has description.
- [X] Component defines [`PropTypes`](https://reactjs.org/docs/typechecking-with-proptypes.html).
#### Style
- [X] Styles are added to `NewComponent.modules.scss` file inside of the `NewComponent` folder.
- [X] Component uses CSS Modules.
- [X] Design is compatible with [Monday Design System](https://design.monday.com/).
#### Storybook
- [X] Stories include all flows of using the component.
#### Tests
- [X] Tests are compliant with [TESTING_README.md](TESTING_README.md) instructions.


src/components/AvatarGroup/AvatarGroupCounterTooltipContentVirtualizedList.jsx
```diff
@@ -0,0 +1,63 @@
...
+import styles from "./AvatarGroupCounterTooltipContentVirtualizedList.module.scss";
+
+const AvatarGroupCounterTooltipContentVirtualizedList = ({
+  avatarItems,
+  avatarRenderer,
+  type,
+  tooltipContainerAriaLabel,
+  tooltipContentContainerRef
+}) => {
+  const maxOptionsWithoutScroll = 10;
```
orrgottlieb, line 14: all of the consts please outside of the function

src/components/AvatarGroup/AvatarGroupCounterTooltipContentVirtualizedList.jsx
```diff
@@ -0,0 +1,63 @@
...
+  const optionLineHeight = 34;
+  const optionLineWidth = 150;
+  const virtualizedItems = useMemo(
+    () => avatarItems.map(item => ({ ...item, height: optionLineHeight })),
+    [avatarItems]
+  );
+
+  let virtualizedListStyle;
```
orrgottlieb, line 22: memo ? (for reference equality)

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
```diff
@@ -0,0 +1,117 @@
...
+  counterContainerRef = undefined,
+  tooltipContentContainerRef,
+  focusPrevPlaceholderRef,
+  focusNextPlaceholderRef,
+  setShouldUpdate,
+  setIsTooltipVisible
+}) {
+  const hideTooltip = () => {
+    // Tricky way to close the tooltip
+    setTimeout(() => {
```
orrgottlieb, line 18: request animation frame? or use lodash defer please

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
```diff
@@ -0,0 +1,117 @@
...
+      setIsTooltipVisible(true);
+    });
+  };
+
+  // For Counter
+  useKeyEvent({
+    keys: ["Tab"],
+    withoutAnyModifier: true,
+    ref: counterContainerRef,
+    callback: e => {
```
orrgottlieb, line 29: extract callback to use callback

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
```diff
@@ -0,0 +1,117 @@
...
+    // Tricky way to close the tooltip
+    setTimeout(() => {
+      setIsTooltipVisible(false);
+      setIsTooltipVisible(true);
+    });
+  };
+
+  // For Counter
+  useKeyEvent({
+    keys: ["Tab"],
```
orrgottlieb, line 26: save in a const outside of the component

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
```diff
@@ -0,0 +1,117 @@
...
+      }
+    }
+  });
+}
+
+export const avatarRenderer = (item, index, style = {}, type, displayAsGrid) => {
+  const avatarProps = item.value;
+  const overrideStyle = { ...style, width: displayAsGrid ? undefined : "100%" };
+
+  const ClickableWrapper = ({ children }) => {
```
orrgottlieb, line 82: extract component outside of the avatarRenderer

src/components/Menu/MenuItem/MenuItem.jsx
```diff
@@ -181,6 +182,14 @@ const MenuItem = forwardRef(
     }, [iconBackgroundColor, disabled]);
 
     const renderMenuItemIconIfNeeded = () => {
+      if (iconComponent) {
```
orrgottlieb, line 185: not a fan of it = this is a custom renderer for the menu item, if you need one, please create a `CustomMenuItem`

orrgottlieb (changes requested):

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -38,13 +43,15 @@ const AvatarGroup = ({
           ...avatar?.props,
           size: size || avatar?.props?.size,
           type: type || avatar?.props?.type,
-          className: cx(styles.avatarContainer, avatarClassName)
+          className: cx(styles.avatarContainer, avatarClassName),
+          onClick: () => avatarOnClick(avatar)
```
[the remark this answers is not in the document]
SergeyRoyt: Not sure we should pass back all the props, but we should pass some props so user would be able to comfortably determine which avatar was clicked and work with it afterwards. Maybe we should send back only `id`? Perhaps `ariaLabel`, `text` for some cases? And maybe `index`? What do you think?
I know we can send back just the event, but for MenuItem it will be a different type of event, which user would have to handle specifically, which I guess we don't want to - I want to pass exactly same arguments whether Avatar was clicked, avatar from the AvatarGroup or MenuItem with avatar inside

src/styles/states.scss
```diff
@@ -100,9 +100,10 @@
   }
 }
 
-@mixin scroller($width: 4px, $color: var(--ui-border-color)) {
+@mixin scroller($width: 4px, $height: 6px, $color: var(--ui-border-color)) {
```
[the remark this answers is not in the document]
SergeyRoyt: In monday-ui-react-core there are no broken usages of scroller, storybook looks fine. 
Am I correct that this change can't affect anyone out of monday-ui-react-core?

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -18,26 +28,75 @@ const AvatarGroupCounter = ({ counterTooltipAvatars, counterProps, counterToolti
...
+  };
+
+  const areAvatarsClickable = counterTooltipAvatars.some(a => a.props?.onClick);
+  if (areAvatarsClickable) {
+    return (
+      <MenuButton
+        component={counterComponent}
+        zIndex={1}
+        componentClassName={cx(styles.counterContainer, counterSizeStyle, counterColorStyle)}
+        ariaLabel={`${counterValue} additional items`}
```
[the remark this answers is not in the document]
SergeyRoyt: Let's just maybe add then an `ariaLabel` property to the `counterProps`? If none passed - default value like now

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -30,6 +31,10 @@ const AvatarGroup = ({
     return null;
   }
 
+  const avatarOnClick = avatar => {
```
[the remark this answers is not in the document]
SergeyRoyt: Outside of component declaration you mean? Or inline?

src/components/Menu/MenuItem/MenuItem.jsx
SergeyRoyt: Maybe let's not allow all custom components, but make only Avatar component possible here as well as Icon? Like add `avatarIconProps` to the MenuItem? And we have full control to limit avatarProps in case
 I've seen examples of Avatar usage in order to just display a picture - so we'll allow pictures in addition to icons. Seems reasonable to me. What do you think?

src/components/AvatarGroup/AvatarGroup.jsx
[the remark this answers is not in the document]
orrgottlieb: it should be the avatarId

src/components/AvatarGroup/AvatarGroupCounter.jsx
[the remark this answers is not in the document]
orrgottlieb: aria label is managed by the user, which means they need to calculate the number of the counter - which should be internal logic

src/components/AvatarGroup/AvatarGroup.jsx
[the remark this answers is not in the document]
orrgottlieb: outside of the function component

src/components/Menu/MenuItem/MenuItem.jsx
orrgottlieb: people are abusing the icon already to pass whatever renderer that they want, if you will allow another one we can't upgrade, change the component without the risk of breaking our users.

The MenuIcon should not be aware of the Avatar
If you want to be on the safe side you could add a AvatarMenuItem and use it with specific props

src/styles/states.scss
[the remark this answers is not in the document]
orrgottlieb: yes

src/components/AvatarGroup/AvatarGroup.jsx
```diff
@@ -7,6 +7,10 @@ import Tooltip from "../Tooltip/Tooltip";
 import AvatarGroupCounter from "./AvatarGroupCounter";
 import styles from "./AvatarGroup.module.scss";
 
+const avatarOnClick = avatarProps => {
```
orrgottlieb, line 10: why do we need a generator function?

src/components/AvatarGroup/AvatarGroupCounter.jsx
```diff
@@ -1,43 +1,103 @@
...
+        component={counterComponent}
+        zIndex={1}
+        componentClassName={cx(styles.counterContainer, counterSizeStyle, counterColorStyle)}
+        ariaLabel={`${counterValue} additional ${counterAriaLabelItemsName}`}
+      >
+        <Menu id="menu" size={Menu.sizes.MEDIUM} className={styles.menu} focusItemIndexOnMount={0}>
+          {counterTooltipAvatars.map((avatar, index) => {
+            return (
+              <AvatarMenuItem
+                key={index}
```
orrgottlieb, line 69: `key` should not be index, if you have id please use it

src/components/AvatarGroup/AvatarGroupCounterTooltipContentVirtualizedList.jsx
orrgottlieb: please define the FINAL variables before the component defenition

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
```diff
@@ -0,0 +1,130 @@
...
+    ref: counterContainerRef,
+    callback: useCallback(
+      e => {
+        if (e.target === counterContainerRef.current) {
+          e.preventDefault();
+          tooltipContentContainerRef.current && tooltipContentContainerRef.current.focus();
+          setShouldUpdate(prev => !prev);
+        }
+      },
+      [counterContainerRef.current, setShouldUpdate, tooltipContentContainerRef]
```
orrgottlieb, line 39: see the lint comment

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
```diff
@@ -0,0 +1,130 @@
...
+    )
+  });
+
+  // For Counter
+  useKeyEvent({
+    keys: KEYS,
+    modifier: useKeyEvent.modifiers.SHIFT,
+    ref: counterContainerRef,
+    callback: useCallback(
```
orrgottlieb, line 48: all of the `useKeyEvents` are almost the same, can you please extract to a "customHook" ?

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
```diff
@@ -0,0 +1,130 @@
...
+          e.preventDefault();
+          counterContainerRef.current && counterContainerRef.current.focus();
+        }
+      },
+      [counterContainerRef, tooltipContentContainerRef.current]
+    )
+  });
+}
+
+const ClickableWrapper = ({ children, avatarProps }) => {
```
orrgottlieb, line 92: this could be a nice addition to `Clickable` ?

src/components/Menu/MenuItem/AvatarMenuItem.jsx
```diff
@@ -0,0 +1,22 @@
...
+import PropTypes from "prop-types";
+
+// MenuItem with Avatar instead of icon
+const AvatarMenuItem = forwardRef((props, ref) => {
+  const renderAvatar = useCallback(() => <Avatar {...props.avatarProps} />, [props.avatarProps]);
+  return <MenuItem {...props} icon={renderAvatar} ref={ref} />;
+});
+
+AvatarMenuItem.propTypes = {
+  props: PropTypes.shape({ ...MenuItem.propTypes, avatarProps: PropTypes.shape(Avatar.propTypes) })
```
orrgottlieb, line 13: props is not a prop of `AvatarMenuItem`

src/components/Menu/MenuItem/AvatarMenuItem.jsx
```diff
@@ -0,0 +1,22 @@
+import React, { forwardRef, useCallback } from "react";
+import MenuItem from "./MenuItem";
+import Avatar from "../../Avatar/Avatar";
+import PropTypes from "prop-types";
+
+// MenuItem with Avatar instead of icon
+const AvatarMenuItem = forwardRef((props, ref) => {
```
orrgottlieb, line 7: please deconstruct with `{ avatarProps, ...rest }`

src/components/Menu/MenuItem/AvatarMenuItem.jsx
```diff
@@ -0,0 +1,22 @@
...
+  const renderAvatar = useCallback(() => <Avatar {...props.avatarProps} />, [props.avatarProps]);
+  return <MenuItem {...props} icon={renderAvatar} ref={ref} />;
+});
+
+AvatarMenuItem.propTypes = {
+  props: PropTypes.shape({ ...MenuItem.propTypes, avatarProps: PropTypes.shape(Avatar.propTypes) })
+};
+AvatarMenuItem.defaultProps = {
```
orrgottlieb, line 15: remove if not needed

orrgottlieb (changes requested):

src/components/AvatarGroup/AvatarGroup.jsx
SergeyRoyt: Cause in Avatar according whether onClick event was passed or not ClickableWrapper is applying. And if I'd leave it as it were `onClick={() => avatarOnClick(avatar.props)}` it would have always been "clickable" as there always some function

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
SergeyRoyt: Don't see how you can effectively avoid `useKeyEvents` usage duplication... Are you speaking about this? 
Or you mean just extracting `useTooltipContentTabNavigation` into a separate file? Cause they are already in a custom hook

src/components/Menu/MenuItem/AvatarMenuItem.jsx
SergeyRoyt: Could you please check [this commit](https://github.com/mondaycom/monday-ui-react-core/pull/706/commits/91825cd8a7f3951ea323c863064142c73daad9e8)? Does it look better? 
Code completion for MenuItem props is working fine, but for Avatar props it's not, how do I fix that?

src/components/AvatarGroup/AvatarGroupCounterTooltipHelper.js
SergeyRoyt: I've seen the tip, but it does work with `ref.current` and doesn't work with just `ref`
Not sure how to make it work some other way. Maybe some explanation would help

src/components/Menu/MenuItem/AvatarMenuItem.jsx
orrgottlieb: IMO i rather have al the props in the argument of the function and not to deconstruct it with the component

orrgottlieb (approved):

PR #716, opened 2022-06-26
Storybook/sergeyro/flexible welcome page

SergeyRoyt:
https://monday.monday.com/boards/245345663/pulses/2846323029

- Storybook welcome page, which is responsive to screen size
- Storybook a11y page, responsive principles
- Storybook responsive related components
- Flex gapRow & gapColumn props


src/components/Flex/Flex.jsx
```diff
@@ -29,7 +31,11 @@ const Flex = forwardRef(
   ) => {
     const componentRef = useRef(null);
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
-    const overrideStyle = useMemo(() => ({ ...style, gap: `${gap}px` }), [style, gap]);
+    const overrideStyle = useMemo(() => {
+      const overrideGapRow = gapRow || gap;
```
hadasfa, line 35: but the default value is gaps.none which different from undefined. it will override the gap value even if it is exist. I think that gapRow and gapColumn default value should be undefined for fixing this.

src/components/Flex/Flex.jsx
```diff
@@ -29,7 +31,11 @@ const Flex = forwardRef(
...
+      const overrideGapColumn = gapColumn || gap;
```
hadasfa, line 36: same

src/components/Flex/Flex.jsx
hadasfa: or maybe its working because none = 0 which in condition same ad false?

src/storybook/components/related-components/related-components.jsx
```diff
@@ -20,7 +21,13 @@ export const RelatedComponents = ({ componentsNames }) => {
       }),
     [componentsNames]
   );
-  return <article className="monday-storybook-related-components">{componentsDataElements}</article>;
+  return (
+    <article className="monday-storybook-related-components">
+      <Flex direction={Flex.directions.ROW} align={Flex.align.START} wrap gap={20}>
```
hadasfa, line 26: I think here maybe meirv maybe will prefer media queries behavior. please consult with her.

src/storybook/stand-alone-documentaion/accessibility/accessibility.stories.mdx
```diff
@@ -32,123 +28,158 @@ import {
...
 Web Accessibility is based on guidelines published by The World Wide Web Consortium (W3C) for accessible content called Web Content Accessibility Guidelines or WCAG.
 
 [WCAG Quick Reference](https://www.w3.org/WAI/WCAG21/quickref/?currentsidebar=%23col_customize#principle1)
 
-<Tip title="Get our Accessibility checklist"><Link href={accessibilityCheckList} target="_blank">Download</Link> our
-    accessibility checklist file.</Tip>
+<Tip title="Get our Accessibility checklist">
+  <Link href={accessibilityCheckList} target="_blank">
+    Download
+  </Link>{" "}
```
hadasfa, line 40: {""} is by purpose?

src/storybook/stand-alone-documentaion/accessibility/principles/principlesAccessibility.jsx
```diff
@@ -1,11 +1,9 @@
...
-import "../../welcome/principles/principles.scss";
 
-const BASE_CLASS = "monday-storybook-principles";
-
-export const PrinciplesAccesability = () => (
-  <div className={BASE_CLASS}>
+export const PrinciplesAccessibility = () => (
+  <Flex direction={Flex.directions.ROW} wrap gap={Flex.gaps.LARGE} align={Flex.align.START}>
     <Principle
       imgSrc={principleClarity}
```
hadasfa, line 8: I am not sure that in this case flex wrap will be the right approach.  Check it with meirav please

src/storybook/stand-alone-documentaion/welcome/contributor/contributor.scss
```diff
@@ -3,19 +3,18 @@
 .monday-storybook-welcome-contributor {
   &_visual-element {
     position: relative;
-    width: 100%;
+    width: 320px;
```
hadasfa, line 6: why?

hadasfa (changes requested):
hey! I have been looked in to this pr.
the solution of flex wrap looks good to me for the contributors as we discuss because we have multiple rows with people. but I do no think its fit for the related components and descriptions. there I guess we will always want to display everything in one row or one column and the current solution not will provide this result. Also the contributors are not centered when the rows are not full and we need to fix it. Please do design review with Meriav and fix according to her notes. If something is hard to implement (for example centering the contributors) we can fix it together, I scheduled a meeting for us on Wednesday. Since this pr is not urgent, let's not merge it tomorrow? Let's wait until the day after, I also have a few questions about some css changes.

PR #723, opened 2022-06-28
Add css module support for chips component

SergeyRoyt:
https://monday.monday.com/boards/245345663/pulses/2867072856

Module styles to Chips in order to avoid css overrides like in case of Etay Liberman (https://monday.slack.com/archives/C02S6F817JQ/p1656349368074379)
Breaking changes, cause originally css styles naming was not unique (`.label`) and there is no way to support backward compatibility. But probably need to change also known usages of Chips where styles get overriding


hadasfa (approved):
perfect

PR #727, opened 2022-06-29
Butotn inset focus style

orrgottlieb:
Add the ability to change the focus-visable look and feel of button and iconButton


src/components/Button/Button.jsx
```diff
@@ -145,7 +146,8 @@ const Button = forwardRef(
           "monday-style-button--left-flat": leftFlat,
           "monday-style-button--prevent-click-animation": preventClickAnimation,
           "monday-style-button--no-side-padding": noSidePadding,
-          "monday-style-button--disabled": disabled
+          "monday-style-button--disabled": disabled,
+          "inset-focus-style": insetFocus
         }
       );
     }, [
```
laviomri, line 153: he's right! :) 
image

laviomri (approved):
left a comment. 
Also assuming that the snapshot changes are OK

src/components/Button/Button.jsx
orrgottlieb: +1

PR #734, opened 2022-07-04
combox auto select first option

mentaman:
We want the user to be able to just search and enter, without pressing key down

### DIFF
--- a/src/components/Combobox/Combobox.jsx
+++ b/src/components/Combobox/Combobox.jsx
@@ -24,7 +24,7 @@ const Combobox = forwardRef(
       id,
       placeholder,
       size,
-      keepOptionSelected,
+      defaultVisualFocusFirstItem,
       optionLineHeight,
       optionsListHeight,
       autoFocus,
@@ -124,7 +124,7 @@ const Combobox = forwardRef(
       visualFocusItemId,
       onItemClickCallback: onOptionClick
     } = useActiveDescendantListFocus({
-      keepOptionSelected,
+      defaultVisualFocusFirstItem,
       focusedElementRef: inputRef,
       containerElementRef: resultsContainerRef,
       focusedElementRole: useActiveDescendantListFocus.roles.COMBOBOX,
@@ -291,7 +291,7 @@ Combobox.propTypes = {
   noResultsRenderer: PropTypes.func,
   stickyCategories: PropTypes.bool,
   /** By default have an option selected, when focusing selecting the first option, or when changing items */
-  keepOptionSelected: PropTypes.bool,
+  defaultVisualFocusFirstItem: PropTypes.bool,
   /** Clear the filter/search on selection (click or enter) */
   clearFilterOnSelection: PropTypes.bool,
   /**
@@ -337,7 +337,7 @@ Combobox.defaultProps = {
   stickyCategories: false,
   optionRenderer: null,
   clearFilterOnSelection: false,
-  keepOptionSelected: false,
+  defaultVisualFocusFirstItem: false,
   renderOnlyVisibleOptions: false,
   onClick: _optionData => {}
 };
--- a/src/components/Combobox/Combobox.jsx
+++ b/src/components/Combobox/Combobox.jsx
@@ -24,7 +24,7 @@ const Combobox = forwardRef(
       id,
       placeholder,
       size,
-      chooseFirstOnFocus,
+      keepOptionSelected,
       optionLineHeight,
       optionsListHeight,
       autoFocus,
@@ -124,7 +124,7 @@ const Combobox = forwardRef(
       visualFocusItemId,
       onItemClickCallback: onOptionClick
     } = useActiveDescendantListFocus({
-      chooseFirstOnFocus,
+      keepOptionSelected,
       focusedElementRef: inputRef,
       containerElementRef: resultsContainerRef,
       focusedElementRole: useActiveDescendantListFocus.roles.COMBOBOX,
@@ -289,7 +289,8 @@ Combobox.propTypes = {
   shouldScrollToSelectedItem: PropTypes.bool,
   noResultsRenderer: PropTypes.func,
   stickyCategories: PropTypes.bool,
-  chooseFirstOnFocus: PropTypes.bool,
+  /** By default have an option selected, when focusing selecting the first option, or when changing items */
+  keepOptionSelected: PropTypes.bool,
   /** Clear the filter/search on selection (click or enter) */
   clearFilterOnSelection: PropTypes.bool,
   /**
@@ -335,7 +336,7 @@ Combobox.defaultProps = {
   stickyCategories: false,
   optionRenderer: null,
   clearFilterOnSelection: false,
-  chooseFirstOnFocus: false,
+  keepOptionSelected: false,
   renderOnlyVisibleOptions: false,
   onClick: _optionData => {}
 };
--- a/src/components/Combobox/Combobox.jsx
+++ b/src/components/Combobox/Combobox.jsx
@@ -24,6 +24,7 @@ const Combobox = forwardRef(
       id,
       placeholder,
       size,
+      chooseFirstOnFocus,
       optionLineHeight,
       optionsListHeight,
       autoFocus,
@@ -123,12 +124,13 @@ const Combobox = forwardRef(
       visualFocusItemId,
       onItemClickCallback: onOptionClick
     } = useActiveDescendantListFocus({
-      focusedElementRef: inputRef,
-      containerElementRef: resultsContainerRef,
-      focusedElementRole: useActiveDescendantListFocus.roles.COMBOBOX,
-      itemsIds: filteredOptionsIds,
-      onItemClick: overrideOnClick,
-      isItemSelectable: isChildSelectable
+        chooseFirstOnFocus,
+        focusedElementRef: inputRef,
+        containerElementRef: resultsContainerRef,
+        focusedElementRole: useActiveDescendantListFocus.roles.COMBOBOX,
+        itemsIds: filteredOptionsIds,
+        onItemClick: overrideOnClick,
+        isItemSelectable: isChildSelectable
     });
 
     const hasResults = filteredOptions.length > 0;
@@ -287,6 +289,7 @@ Combobox.propTypes = {
   shouldScrollToSelectedItem: PropTypes.bool,
   noResultsRenderer: PropTypes.func,
   stickyCategories: PropTypes.bool,
+  chooseFirstOnFocus: PropTypes.bool,
   /** Clear the filter/search on selection (click or enter) */
   clearFilterOnSelection: PropTypes.bool,
   /**
@@ -332,6 +335,7 @@ Combobox.defaultProps = {
   stickyCategories: false,
   optionRenderer: null,
   clearFilterOnSelection: false,
+  chooseFirstOnFocus: false,
   renderOnlyVisibleOptions: false,
   onClick: _optionData => {}
 };
--- a/src/components/Combobox/Combobox.jsx
+++ b/src/components/Combobox/Combobox.jsx
@@ -124,13 +124,13 @@ const Combobox = forwardRef(
       visualFocusItemId,
       onItemClickCallback: onOptionClick
     } = useActiveDescendantListFocus({
-        chooseFirstOnFocus,
-        focusedElementRef: inputRef,
-        containerElementRef: resultsContainerRef,
-        focusedElementRole: useActiveDescendantListFocus.roles.COMBOBOX,
-        itemsIds: filteredOptionsIds,
-        onItemClick: overrideOnClick,
-        isItemSelectable: isChildSelectable
+      chooseFirstOnFocus,
+      focusedElementRef: inputRef,
+      containerElementRef: resultsContainerRef,
+      focusedElementRole: useActiveDescendantListFocus.roles.COMBOBOX,
+      itemsIds: filteredOptionsIds,
+      onItemClick: overrideOnClick,
+      isItemSelectable: isChildSelectable
     });
 
     const hasResults = filteredOptions.length > 0;
--- a/src/components/Combobox/Combobox.jsx
+++ b/src/components/Combobox/Combobox.jsx
@@ -335,7 +335,7 @@ Combobox.defaultProps = {
   stickyCategories: false,
   optionRenderer: null,
   clearFilterOnSelection: false,
-  chooseFirstOnFocus: true,
+  chooseFirstOnFocus: false,
   renderOnlyVisibleOptions: false,
   onClick: _optionData => {}
 };
--- a/src/components/Combobox/Combobox.jsx
+++ b/src/components/Combobox/Combobox.jsx
@@ -335,7 +335,7 @@ Combobox.defaultProps = {
   stickyCategories: false,
   optionRenderer: null,
   clearFilterOnSelection: false,
-  chooseFirstOnFocus: false,
+  chooseFirstOnFocus: true,
   renderOnlyVisibleOptions: false,
   onClick: _optionData => {}
 };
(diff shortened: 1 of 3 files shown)


src/components/Combobox/Combobox.jsx
```diff
@@ -332,6 +335,7 @@ Combobox.defaultProps = {
   stickyCategories: false,
   optionRenderer: null,
   clearFilterOnSelection: false,
+  chooseFirstOnFocus: false,
```
RonNachmany, line 338: Naming + add docs comment

RonNachmany (changes requested):

RonNachmany (approved):

hadasfa (changes requested):
nice! please add tests for check this functionality and also run all previous tests with this flag on (the previous tests should run once with this flag off and once with this flag on unless if one of them is not relevant with the new behavior)

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
+    focusedElementRef: {
+      current: element
+    },
+    keepOptionSelected,
+    itemsIds: ITEM_IDS,
+    isItemSelectable: isItemSelectable,
+    onItemClick,
+    isHorizontalList: isHorizontal
+  };
+  return renderHook(argprops => useActiveDescendantListFocus({ ...props, ...argprops }));
```
hadasfa, line 33: where argprops come from :o

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
 }
 
-function runListUnitTest(isHorizontal) {
+function runListUnitTest({ isHorizontal, keepOptionSelected }) {
   const moveForwardKey = isHorizontal ? "{arrowRight}" : "{arrowDown}";
   const oppositeMoveForwardKey = !isHorizontal ? "{arrowRight}" : "{arrowDown}";
+  const naturalIndex = keepOptionSelected ? 0 : undefined;
```
hadasfa, line 39: its not natural focus because the natural index is always on the container. visualFocusIndex?

hadasfa: and it's not good that the visual focus index before navigate to the combobox will be 0, it's affect the screen reader. it must be undefined :(

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
+    });
+
+    const beforeIndex = result.current.visualFocusItemIndex;
+    const beforeId = result.current.visualFocusItemId;
+
+    const removeItems = 2;
+    const itemsAfterRemove = ITEM_IDS.slice(removeItems);
+    rerender({ itemsIds: itemsAfterRemove });
+
+    if (!keepOptionSelected) {
```
hadasfa, line 67: if this test should run only for !keepOptionSelected let's separate it to a different suite instead run tests conditionally . it's hard to understand like this what is the behavior that we need to expect in each state.

hadasfa: so split it to to tests and find a way for writing the shared logic (for example a function or something)

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
 }
 
-function runListUnitTest(isHorizontal) {
+function runListUnitTest({ isHorizontal, keepOptionSelected }) {
   const moveForwardKey = isHorizontal ? "{arrowRight}" : "{arrowDown}";
   const oppositeMoveForwardKey = !isHorizontal ? "{arrowRight}" : "{arrowDown}";
+  const naturalIndex = keepOptionSelected ? 0 : undefined;
+  const naturalIndexAfterInitial = keepOptionSelected ? 0 : -1;
+
+  it("should focus same item after item changed", async () => {
```
hadasfa, line 42: please split to 2 tests for the 2 different states:
"should visually focus on the item with the same index after items list changed"
"should  visually focus on the same item(with the same id) after items list changed"

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
+      expect(result.current.visualFocusItemId).toEqual(itemsAfterRemove[beforeIndex]);
+      expect(result.current.visualFocusItemIndex).toEqual(beforeIndex);
+    } else {
+      expect(result.current.visualFocusItemId).toEqual(beforeId);
+      expect(result.current.visualFocusItemIndex).toEqual(beforeIndex - removeItems);
+    }
+  });
+
+  it("should focus correct item on focus", async () => {
+    const onItemClick = jest.fn();
```
hadasfa, line 77: if u expect here to 2 different results for 2 different state it should be 2 different states and the name of each test should indicate what "correct item is".
For example (new names for the 2 different tests):
- "Should not visually focus on any item after initial natural focus on the list wrapper element."
- "Should visually focus on the first item in the list after initial natural focus on the list wrapper element."

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
+    act(() => {
+      // set focus on the list's element which in charge on natural focus element
+      element.focus();
+    });
+
+    expect(result.current.visualFocusItemIndex).toEqual(naturalIndex);
+  });
+
+  it("should focus correct item keyboard forward", async () => {
+    const onItemClick = jest.fn();
```
hadasfa, line 89: Thank u!

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
+
+    act(() => {
+      // set focus on the list's element which in charge on natural focus element
+      element.focus();
+    });
+
+    expect(result.current.visualFocusItemIndex).toEqual(naturalIndex);
+  });
+
+  it("should focus correct item keyboard forward", async () => {
```
hadasfa, line 88:
```suggestion
  it("Should move visual focus to the closest item when use press keyboard forward", async () => {
```

hadasfa, line 88: Thank u!

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
+    const onItemClick = jest.fn();
+    const { result } = renderHookForTest({ onItemClick, isHorizontal, keepOptionSelected });
+
+    act(() => {
+      // set focus on the list's element which in charge on natural focus element
+      element.focus();
+    });
+
+    const keyboardTimes = 2;
```
hadasfa, line 97: one press is enough

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,109 @@ import userEvent from "@testing-library/user-event";
...
+    for (let i = 0; i < keyboardTimes; i++) {
+      act(() => {
+        // move visual focus to first item
+        userEvent.keyboard(moveForwardKey);
+      });
+    }
+
+    expect(result.current.visualFocusItemIndex).toEqual(naturalIndexAfterInitial + keyboardTimes);
```
hadasfa, line 105: no need to split two tests here. just keep in variable the value of result.current.visualFocusItemIndex before pressing keyboard forward in variable + steps count and compare it to the result.current.visualFocusItemIndex after press

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -93,22 +170,57 @@ function runListUnitTest(isHorizontal) {
...
+  // all open
+  const allOpenComb = {};
+  for (let feature of features) {
+    allOpenComb[feature] = true;
+  }
+  combinations.push(allOpenComb);
+
+  // one open all other closed
+  for (const feature of features) {
+    const oneOpen = {};
```
hadasfa, line 195: this is super cool but now u will have tests that will run only for specific feature

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
hadasfa: I did not understand if it is the index before we focus on the wrapper or after, if it is before it's a problem. if it's after it is ok :)

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
mentaman: החסידה

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
mentaman: it's after focus

mentaman: added a test to make sure it starts as undefined

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
mentaman: wanted two since for one it starts with -1 and for the other 0, so wanted to make sure it's not some value by coincidence

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
mentaman: You have either all open or all closed.

and also one open each time.

Didnt want to make it expanontially o(2^n) I can do it o(n^2) though so each feature will be with another feature, but I think this way is enough and each feature will be together anyway since there is an option where all are together just not two alone

mentaman: See here
image

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
hadasfa: the steps count is not matter because it will always be previous + 1 ... but fine it's not critical

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
mentaman: From rerender

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
mentaman: Yes at the begging it wasnt previous + 2, it was just a number before the changes. 
I still prefer 2 😦

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -5,32 +5,66 @@ import userEvent from "@testing-library/user-event";
...
+  };
+  return renderHook(argprops => useActiveDescendantListFocus({ ...props, ...argprops }));
 }
 
-function runListUnitTest(isHorizontal) {
+function runListUnitTest({ isHorizontal, defaultVisualFocusFirstItem }) {
   const moveForwardKey = isHorizontal ? "{arrowRight}" : "{arrowDown}";
   const oppositeMoveForwardKey = !isHorizontal ? "{arrowRight}" : "{arrowDown}";
+
+  it("should focus correct item keyboard forward", async () => {
```
hadasfa, line 40:
```suggestion
  it("should focus index + 1 item when user press keyboard forward", async () => {
```

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -78,37 +112,153 @@ function runListUnitTest(isHorizontal) {
...
+
+      act(() => {
+        // set focus on the list's element which in charge on natural focus element
+        element.focus();
+      });
+
+      expect(result.current.visualFocusItemIndex).toEqual(0);
+    });
+
+    it("should focus first selectable one only", async () => {
```
hadasfa, line 235: it's a copy paste of the tests for is selectable that already in the general unit tests no?
I think we can delete it?

src/hooks/useActiveDescendantListFocus/__tests__/useActiveDescendantListFocus.jest.js
```diff
@@ -78,37 +112,153 @@ function runListUnitTest(isHorizontal) {
...
+    });
   });
-  describe("Horizontal list", () => {
-    const isHorizontal = true;
-    runListUnitTest(isHorizontal);
+
+  describe("no defaultVisualFocusFirstItem option", () => {
+    const defaultVisualFocusFirstItem = false;
+
+    it("should not change keyboard item on focus", async () => {
```
hadasfa, line 252:
```suggestion
    it("should not set any item as visually focus when menu is focused", async () => {
```

src/hooks/useActiveDescendantListFocus/index.js
```diff
@@ -42,6 +43,45 @@ function useActiveDescendantListFocus({
...
+  const prevVisualFocusItemId = usePrevious(itemsIds[visualFocusItemIndex]);
+
+  const getFirstSelectable = useCallback(() => {
+    for (let idx = 0; idx < itemsIds.length; idx++) {
+      if (isItemSelectable(idx)) {
+        return idx;
+      }
+    }
+    return -1;
+  }, [itemsIds]);
```
hadasfa, line 58: please fix linter error

src/hooks/useActiveDescendantListFocus/index.js
```diff
@@ -42,6 +43,45 @@ function useActiveDescendantListFocus({
...
+
+  useEffect(() => {
```
hadasfa, line 60: it's a very scary use effect for outside reader... it's take a few moments to understand what is purpose.

If I were you, I would move all the logic which related to keep the id with the same value when the list change to a different custom hook "useKeepSameItemFocusedWhenItemsListChange" including all the use previous usage and stuff... you don't need to do it, I will do it before merging.

src/hooks/useActiveDescendantListFocus/index.js
```diff
@@ -123,6 +163,13 @@ function useActiveDescendantListFocus({
     }
   }, [visualFocusItemIndex]);
 
+  const onFocusCallback = useCallback(() => {
+    if (visualFocusItemIndex === -1 && defaultVisualFocusFirstItem) {
+      triggerByKeyboard.current = true;
```
hadasfa, line 168: but what if focus is not by keyboard? it will draw you a focus when the focus trigger is a mouse... need to check it

PR #735, opened 2022-07-04
IconButton: added wrapperClassName prop

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/2889734734

Added div-wrapper and a `wrapperClassName` prop

### DIFF
--- a/src/components/IconButton/IconButton.jsx
+++ b/src/components/IconButton/IconButton.jsx
@@ -16,7 +16,7 @@ const IconButton = forwardRef(
   (
     {
       className,
-      containerClassName,
+      wrapperClassName,
       id,
       icon,
       size,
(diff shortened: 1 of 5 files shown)


src/components/IconButton/IconButton.jsx
```diff
@@ -76,32 +77,34 @@ const IconButton = forwardRef(
...
-            icon={icon}
-            iconType={Icon.type.SVG}
-            iconSize={iconSize}
-            ignoreFocusStyle
-            className="icon-button-padding"
-            clickable={false}
-          />
-        </Button>
-      </ToolTip>
+      <div className={cx(wrapperClassName, styles.wrapper)}>
```
hadasfa, line 80: maybe iconButtonWrapper will be more clear?

src/components/IconButton/IconButton.modules.scss
```diff
@@ -1,3 +1,9 @@
-.referenceContainer {
+.wrapper {
+  width: fit-content;
```
hadasfa, line 2: please ask Orr if it is ok to use this property (fit-content) because it's looks like it not supported in all browsers: https://caniuse.com/?search=fit-content

src/components/IconButton/IconButton.jsx
hadasfa: Also let's do it conditional for not break exist components which use this component already
something like: const IconButtonWrapper = wrapperClassName ? "div" : Fragment

hadasfa (approved):

PR #745, opened 2022-07-10
Set default text color for dialog container

hadasfa:


laviomri (approved):

3dyonic (commented):
looks like you took the color from the story and add it to the component.
why?
now all the instances will get this color, why do you declare color on a wrapper component?
it should inherit the body color? no?

PR #753, opened 2022-07-14
Navigate in list with keyboard.

samitc:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests

### DIFF
--- a/src/components/List/List.jsx
+++ b/src/components/List/List.jsx
@@ -13,35 +13,23 @@ const List = forwardRef(
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
     const Component = component;
     const childrenRefs = useRef([]);
-    const changeFocus = focusIndex => {
-      setFocusIndex(focusIndex);
-      childrenRefs.current[focusIndex].focus();
-    };
     const onKeyDown = useCallback(
       event => {
         const isUpKey = event.keyCode === keyCodes.UP_ARROW;
         const isDownKey = event.keyCode === keyCodes.DOWN_ARROW;
+        let overrideFocusIndex = undefined;
         if (isUpKey || isDownKey) {
           if (isDownKey && focusIndex + 1 < childrenRefs.current.length) {
-            changeFocus(focusIndex + 1);
+            overrideFocusIndex = focusIndex + 1;
           } else if (isUpKey && focusIndex > 0) {
-            changeFocus(focusIndex - 1);
+            overrideFocusIndex = focusIndex - 1;
           }
           event.preventDefault();
-        }
-      },
-      [focusIndex]
-    );
-    const childMapping = useCallback(
-      (child, index) => {
-        return {
-          ...child,
-          ref: ref => (childrenRefs.current[index] = ref),
-          props: {
-            ...child.props,
-            tabIndex: focusIndex === index ? "0" : "-1"
+          if (overrideFocusIndex !== undefined) {
+            setFocusIndex(overrideFocusIndex);
+            childrenRefs.current[overrideFocusIndex].focus();
           }
-        };
+        }
       },
       [focusIndex]
     );
@@ -50,11 +38,17 @@ const List = forwardRef(
       if (renderOnlyVisibleItems) {
         override = <VirtualizedListItems>{override}</VirtualizedListItems>;
       } else {
-        override = override.map(childMapping);
+        childrenRefs.current = childrenRefs.current.slice(0, override.length);
+        override = React.Children.map(override, (child, index) =>
+          React.cloneElement(child, {
+            ref: ref => (childrenRefs.current[index] = ref),
+            tabIndex: focusIndex === index ? "0" : "-1"
+          })
+        );
       }
 
       return override;
-    }, [childMapping, children, renderOnlyVisibleItems]);
+    }, [children, focusIndex, renderOnlyVisibleItems]);
 
     return (
       <Component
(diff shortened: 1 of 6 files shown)


src/components/List/List.jsx
```diff
@@ -1,35 +1,74 @@
...
+        };
+      },
+      [focusIndex]
+    );
     const overrideChildren = useMemo(() => {
       let override = children;
       if (renderOnlyVisibleItems) {
         override = <VirtualizedListItems>{override}</VirtualizedListItems>;
+      } else {
+        override = override.map(childMapping);
```
hadasfa, line 53: This is nice :) But not very standard. Also it will not work in case you have single element/null/undef as child (never should be the case in a list but still we do not want to throw an error.)
Please use React.Children.map for iterate the children and React.cloneElement for cloning the element (this way you also will not need to use useCallback for child mapping)

src/components/List/List.jsx
```diff
@@ -1,35 +1,74 @@
...
+import { keyCodes } from "../../constants/KeyCodes";
 
 const List = forwardRef(
   ({ className, id, component, children, dense, ariaLabel, ariaDescribedBy, renderOnlyVisibleItems, style }, ref) => {
     const componentRef = useRef(null);
+    const [focusIndex, setFocusIndex] = useState(0);
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
     const Component = component;
+    const childrenRefs = useRef([]);
+    const changeFocus = focusIndex => {
```
hadasfa, line 16: We prefer not to create anonymous functions inside our component render function if we can prevent this. I wrote suggestion for how to change the code for deleting this function.

hadasfa: (The reason for this it that writing an anonymous function in a render method will create new function instance any render)

src/components/List/List.jsx
```diff
@@ -1,35 +1,74 @@
...
+      setFocusIndex(focusIndex);
+      childrenRefs.current[focusIndex].focus();
+    };
+    const onKeyDown = useCallback(
+      event => {
+        const isUpKey = event.keyCode === keyCodes.UP_ARROW;
+        const isDownKey = event.keyCode === keyCodes.DOWN_ARROW;
+        if (isUpKey || isDownKey) {
+          if (isDownKey && focusIndex + 1 < childrenRefs.current.length) {
+            changeFocus(focusIndex + 1);
```
hadasfa, line 26: you can add to this function a overrideFocusIndex = undef and inside every condition only change it's value.
after the condition you can check if overrideFocusIndex!== undefined and only if it is you can call the set focus logic (not in a function)

src/components/ListItem/ListItem.jsx
```diff
@@ -49,7 +49,7 @@ const ListItem = forwardRef(({ className, id, onClick, selected, disabled, size,
       onClick={componentOnClick}
       onKeyDown={onKeyDown}
       role="listitem"
-      tabIndex={disabled ? "-1" : tabIndex}
+      tabIndex={tabIndex}
```
hadasfa, line 52: why? if the list item is disabled we do not want to allow focus on it

hadasfa (changes requested):
Really awesome PR. wrote 2 small notes :)

src/constants/KeyCodes.js
```diff
@@ -1,6 +1,8 @@
 export const keyCodes = {
   ENTER: "Enter",
-  SPACE: " "
+  SPACE: " ",
+  DOWN_ARROW: 40,
```
orrgottlieb, line 4: use these - as the cods are deprecated
 
  ARROW_DOWN: "ArrowDown",
  ARROW_UP: "ArrowUp"

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

src/components/List/List.jsx
```diff
@@ -1,35 +1,68 @@
...
+import { keyCodes } from "../../constants/KeyCodes";
 
 const List = forwardRef(
   ({ className, id, component, children, dense, ariaLabel, ariaDescribedBy, renderOnlyVisibleItems, style }, ref) => {
     const componentRef = useRef(null);
+    const [focusIndex, setFocusIndex] = useState(0);
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
     const Component = component;
+    const childrenRefs = useRef([]);
+    const onKeyDown = useCallback(
```
orrgottlieb, line 16: did you check what happens when there are multiple Lists rendered ? 

i suggest you should add a onFocus and onBlur to initiate the keydown listening

src/components/List/List.jsx
```diff
@@ -1,35 +1,68 @@
...
+      event => {
+        const isUpKey = event.keyCode === keyCodes.UP_ARROW;
+        const isDownKey = event.keyCode === keyCodes.DOWN_ARROW;
+        let overrideFocusIndex = undefined;
+        if (isUpKey || isDownKey) {
+          if (isDownKey && focusIndex + 1 < childrenRefs.current.length) {
```
orrgottlieb, line 22: do we want to add cyclic navigation ?

orrgottlieb (approved):

src/components/List/List.jsx
hadasfa: yes we do but I did not want to make it a blocker. but according to a11y requirements it should be

PR #776, opened 2022-07-24
add exports to package.json + build icons esm file

eran-cohen:
https://monday.monday.com/boards/245345663/views/62248892/pulses/2970462587

using [exports](https://webpack.js.org/guides/package-exports/) in package json to specify path for icons pointing to a newly created dist/icons/index.js file listing all available icons to support tree shaking.

We can now do `import { Close } from 'monday-ui-react-core/icons';` and tree shaking works fine.  
CommonJs and /dist path mapping still works for backward compatibility but I think we should add a dageroid rule discouraging linking directly into the dist folder

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,15 @@
   "name": "monday-ui-react-core",
   "version": "1.50.1",
   "description": "Official monday.com UI resources for application development in React.js",
-  "main": "dist/main.js",
-  "module": "dist/esm.js",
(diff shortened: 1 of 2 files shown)


sahariko (approved):
Amazing stuff!

LGTM but let's wait for a review from someone from the Vibe team.

orrgottlieb (approved):

PR #786, opened 2022-07-26
Use motion tokens on Label.scss

3dyonic:
https://monday.monday.com/boards/2861766393/pulses/2990443694

#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests

### DIFF
--- a/src/components/Label/Label.scss
+++ b/src/components/Label/Label.scss
@@ -25,7 +25,7 @@
   }
 
   &--with-animation {
-    animation: label-pop var(--motion-expressive-short) var(--motion-timing-enter);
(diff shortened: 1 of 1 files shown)


src/components/Label/Label.scss
```diff
@@ -126,7 +125,6 @@
     }
 
     70% {
-      transform: rotate(-5deg);
```
hadasfa, line 129: we sure we want to remove the rotate? we showed it to meirav?

SergeyRoyt: +1

PR #794, opened 2022-07-31
Convert button component to typescript & project configurations 

hadasfa:
**POC for convert first component (butto) to ts**
This pr contains the following:
1. Converting button component to typescript
2. tsconfig & Webpack & jest configurations for support handling with ts / tsx files
3. A script for creating types.d.ts file which contains all the the types of the already converted components (for support autocomplete in other repos which using vibe)
5. Update the helpers functions of our storybook infra for fixing the issue with displaying enums correctly in typescript component and js components:
before:
צילום מסך 2022-08-21 ב-21 46 26
after:
צילום מסך 2022-08-21 ב-21 51 04

6.  Update snapshots which changed after converting (because of aria-busy)
7. Removing mocks that after converting to ts cause tests to failed


src/storybook/functions/create-component-story.js
```diff
@@ -35,8 +50,19 @@ export function createStoryMetaSettings({ component, enumPropNamesArray, iconPro
...
+      // docgen is the parser we using for parsing all our component prop types, default props and
+      // other component data (it's configure under storybook main.js file)
+      const componentName = component.__docgenInfo.displayName;
       argTypes[prop] = {
-        options: Object.values(enums)
+        options: enums,
+        labels: enums,
+        table: {
+          type: {
+            summary: parseStringForEnums(componentName, enumName, enums),
```
hadasfa, line 61: for change what is displayed under the description

src/storybook/functions/create-component-story.js
```diff
@@ -35,8 +50,19 @@ export function createStoryMetaSettings({ component, enumPropNamesArray, iconPro
...
+            // For not displaying box for enumns in controls of js not converted components
+            detail: null
```
hadasfa, line 63: for undo the annoying box

src/storybook/functions/create-component-story.js
```diff
@@ -35,8 +50,19 @@ export function createStoryMetaSettings({ component, enumPropNamesArray, iconPro
     }
     const enums = component[enumName];
     if (enums && enums instanceof Object) {
+      // docgen is the parser we using for parsing all our component prop types, default props and
+      // other component data (it's configure under storybook main.js file)
+      const componentName = component.__docgenInfo.displayName;
       argTypes[prop] = {
-        options: Object.values(enums)
+        options: enums,
+        labels: enums,
```
hadasfa, line 58: for displaying keys instead of values as options labels

src/components/Button/Button.tsx
```diff
@@ -1,25 +1,97 @@
...
 import Icon from "../../components/Icon/Icon";
 import Loader from "../../components/Loader/Loader";
-import { BUTTON_COLORS, BUTTON_INPUT_TYPE, BUTTON_TYPES, getActualSize } from "./ButtonConstants";
+import { ButtonColor, ButtonInputType, ButtonType, getActualSize, Size } from "./ButtonConstants";
 import { getParentBackgroundColorNotTransparent, TRANSPARENT_COLOR } from "./helper/dom-helpers";
-import "./Button.scss";
 import { ELEMENT_TYPES, getTestId } from "../../utils/test-utils";
+import "./Button.scss";
 
+// @ts-ignore
```
hadasfa, line 14: why?

src/components/Button/Button.tsx
```diff
@@ -1,25 +1,97 @@
-/* eslint-disable react/jsx-props-no-spreading,react/button-has-type */
-import { SIZES } from "../../constants/sizes";
 import React, { forwardRef, useCallback, useEffect, useMemo, useRef, useState } from "react";
-import PropTypes from "prop-types";
 import cx from "classnames";
+import { SIZES } from "../../constants/sizes";
```
hadasfa, line 3: check why it's not breaking for js old component when we change size

tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+  "compilerOptions": {
```
hadasfa, line 2: check if we need to do changes here

tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "lib": ["esnext", "dom"],
+    "allowJs": true,
+    "target": "es6",
+    "module": "commonjs",
```
hadasfa, line 6: check why we define it like this

tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "sourceMap": false,
```
hadasfa, line 7: why not?

tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "strict": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
+    "declaration": true,
+    "skipLibCheck": true,
```
hadasfa, line 14: why?

tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "noImplicitAny": true,
+    "outDir": "dist",
+    "baseUrl": ".",
+    "allowSyntheticDefaultImports": true,
+    "paths": {
```
hadasfa, line 19: Are these paths still relevant after all the tree shaking changes?

tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+      "hooks/*": ["src/hooks/*"],
+      "utils/*": ["src/utils/*"],
+      "components/*": ["src/components/*"],
+      "constants/*": ["src/constants/*"]
+    },
+    "jsx": "react-jsx"
```
hadasfa, line 25: why this specifically?

__mocks__/fileMock.js
```diff
@@ -2,6 +2,6 @@ const path = require("path");
 
 module.exports = {
   process(src, filename) {
-    return `module.exports = ${JSON.stringify(path.basename(filename))};`;
+    return { code: `module.exports = ${JSON.stringify(path.basename(filename))};` };
```
hadasfa, line 5: jest 28 breaking change

src/components/AlertBanner/__tests__/AlertBanner-snapshots.jest.js
```diff
@@ -6,20 +6,6 @@ import AlertBannerButton from "../AlertBannerButton/AlertBannerButton";
 import AlertBannerText from "../AlertBannerText/AlertBannerText";
 import { NOOP } from "../../../utils/function-utils";
 
-jest.mock("../../Button/Button", () => {
```
hadasfa, line 9: after converting to ts this mock cause errors on running the tests so I thought it can remove in this stage and later we can figure out how to involve mocks in a way that will work with ts

src/components/Button/__tests__/button-snapshot-tests.jest.js
```diff
@@ -4,11 +4,6 @@ import NOOP from "lodash/noop";
 import Button from "../Button";
 import { WhatsNew } from "../../Icon/Icons";
 
-jest.mock("../../Icon/Icon", () => {
```
hadasfa, line 7: after converting button to ts this mock cause errors on running the tests  (about type is undefined) so I thought it can remove in this stage and later we can figure out how to involve mocks in a way that will work with ts

src/utils/icon-utils.js
```diff
@@ -0,0 +1,16 @@
+const fs = require("fs");
+
```
hadasfa, line 2: When this util was out of the src folder it cause issues while creating the dist folder

jest.config.js
```diff
@@ -63,7 +63,12 @@ module.exports = {
   // globalTeardown: undefined,
 
   // A set of global variables that need to be available in all test environments
-  // globals: {},
+  globals: {
+    extensionsToTreatAsEsm: [".ts", ".js"],
```
sahariko, line 67: `.jsx`/`.tsx`? 🤔

scripts/build-types-index.js
```diff
@@ -0,0 +1,18 @@
...
+const fs = require("fs");
+const path = require("path");
+// @ts-ignore
+// Right now we create types only for components we converted to typescript and we wrote in published-converted-components.
+// After we will convert all the source files we will change the import to webpack/published-components
+const { publishedComponents } = require("../webpack/published-converted-components");
+
+function buildComponentsTypesIndexFile() {
+  const content = Object.entries(publishedComponents)
+    .map(([name, path]) => `export { default as ${name} } from "./types/${path}";`)
```
sahariko, line 12: Does it exist for all components? If not, this could be an issue.

scripts/build-types-index.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
```
sahariko, line 1: This file should probably be merged with the `build-esm` script.

src/components/AlertBanner/__tests__/AlertBanner-snapshots.jest.js
sahariko: Which errors? Why did we need this mock in the first place?

src/components/Button/Button.tsx
```diff
@@ -234,6 +307,7 @@ const Button = forwardRef(
       return (
         <button {...buttonProps}>
           <span className="monday-style-button__loader">
+            {/** @ts-ignore */}
```
sahariko, line 310: Why? If you can, it's preferable to add a reason next to each `@ts-ignore` so that we know why it's there.

src/components/Button/Button.tsx
```diff
@@ -288,107 +365,31 @@ const Button = forwardRef(
...
-};
+Object.assign(Button, {
+  sizes: SIZES,
+  colors: ButtonColor,
+  kinds: ButtonType,
+  types: ButtonInputType,
+  inputTags: ButtonInputType
+});
 
 Button.defaultProps = {
```
sahariko, line 376: Do we still need `defaultProps`? IMO it would be better to switch to native JS default values.

src/components/Button/ButtonConstants.ts
```diff
@@ -0,0 +1,45 @@
...
+  md: SIZES.MEDIUM,
+  lg: SIZES.LARGE
+} as const;
+
+export type Size = typeof SIZES[keyof typeof SIZES] | keyof typeof OLD_BUTTON_SIZES;
+
+// Support old sizes (backward compatible)
+export const getActualSize = (size?: Size) => {
+  if (size && size in OLD_BUTTON_SIZES) {
+    // @ts-expect-error
```
sahariko, line 28: Reason.

src/components/Dropdown/__tests__/__snapshots__/dropdown-snapshot-tests.jest.js.snap
```diff
@@ -1893,6 +1893,7 @@ exports[`Dropdown renders correctly with leftAvatar 1`] = `
               Ocean
             </div>
             <button
+              aria-busy={false}
```
sahariko, line 1896: Is this change legit? I've seen it pop all over snapshot files.

tsconfig.json
sahariko: No, let's remove them.

webpack/published-converted-components.js
```diff
@@ -0,0 +1,19 @@
+const path = require("path");
+const publishedComponents = {
+  Button: "components/Button/Button",
+  AlertBannerButton: "components/AlertBanner/AlertBannerButton",
+  ToastButton: "components/Toast/ToastButton/ToastButton"
```
sahariko, line 5: These components still appear in the `published-components` file, is this intentional?

webpack/published-converted-components.js
```diff
@@ -0,0 +1,19 @@
+const path = require("path");
```
sahariko, line 1: No need for 2 separate files for published components, export 2 constants from the existing `published-components` file - `publishedComponents` and `publishedTSComponents`.

scripts/build-types-index.js
hadasfa: We do not use the published component original file until we end to migrate all components, we are using a file which called published-converted-components until the migration will be done

hadasfa: So why should it be a problem?

src/components/Button/Button.tsx
hadasfa: I think we keep these because something which related to the storybook integration. maybe you remember ? @orrgottlieb @laviomri 
Anyway I will try to remove this and see what happend

src/components/Dropdown/__tests__/__snapshots__/dropdown-snapshot-tests.jest.js.snap
hadasfa: Yes I checked it before and it's legit in terms of a11y

webpack/published-converted-components.js
hadasfa: I saw what you wrote in the other note about this problem. I will change it

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.55.0",
   "description": "Official monday.com UI resources for application development in React.js",
   "main": "./dist/main.js",
+  "types": "./dist/types.d.ts",
```
eran-cohen, line 6: @hadasfa if you're planning to merge this code before you guys finish migrating all the components (highly recommended!), don't forget to remove this line so it won't cause problems with existing MF using vibe

src/components/Button/Button.tsx
```diff
@@ -1,25 +1,97 @@
...
+  onFocus?: (event: React.FocusEvent<HTMLButtonElement>) => void;
+  /** On Button Blur callback */
+  onBlur?: (event: React.FocusEvent<HTMLButtonElement>) => void;
+  rightFlat?: boolean;
+  leftFlat?: boolean;
+  preventClickAnimation?: boolean;
+  noSidePadding?: boolean;
+  /** default color for text color in ON_PRIMARY_COLOR kind (should be any type of css color (rbg, var, hex...) */
+  defaultTextColorOnPrimaryColor?: string;
+  dataTestId: string;
```
eran-cohen, line 82: @hadasfa are you sure dataTestId (also other props like insetFocus, ect) is mandatory? please keep in mind that typescript will enforce this, so people will be forced to pass all mandatory props or their build will fails

src/components/Button/Button.tsx
eran-cohen: I agree with @sahariko, typescript will ignore those values anyway and it will be confusing if we'll continue using defaultProps

eran-cohen (changes requested):
exciting, can't wait to see Vibe fully support typescript!
great work 🔥

src/components/AlertBanner/__tests__/AlertBanner-snapshots.jest.js
hadasfa: the mock return undefined instead of the object which cause exceptions.
We need to solve it but I think it is ok to split it to a separate pr (because this is a bit big already).
I did not implement this tests but in general mock like this in snapshot are used for decrease the dependency of a component in the structure of a component inside it, so snapshot will not break when the button structure is change.
It would be better to have mock here but we have a lot of components which not using mocks for this so I think it is ok to handle this issue separately

package.json
hadasfa: why this can cause problems? can we talk about this?

src/components/Button/Button.tsx
hadasfa: it's your note so I am not sure. that was the explanation you wrote next to it:
צילום מסך 2022-08-23 ב-8 18 14

As I understand, it's because we did not convert these components to TS. Am I correct? (If so I will add a note)

src/components/Button/ButtonConstants.ts
hadasfa: This note was added in your TS POC PR. maybe you remember why?

src/components/Button/Button.tsx
hadasfa: Added a comment about this

src/components/Button/Button.tsx
hadasfa: We did not removing the default props for component which we using for fake inheritance (which is basically not inheritance but composition).
For example we have icon button which receive the same default props like the button and we want the storybook to be aware of it.
In regular component we will remove default props.

src/components/Button/Button.tsx
```diff
@@ -1,25 +1,97 @@
...
+  noSidePadding?: boolean;
+  /** default color for text color in ON_PRIMARY_COLOR kind (should be any type of css color (rbg, var, hex...) */
+  defaultTextColorOnPrimaryColor?: string;
+  dataTestId?: string;
+  /** Change the focus indicator from around the button to within it */
+  insetFocus?: boolean;
+}
+
+const Button: React.ForwardRefExoticComponent<ButtonProps & React.RefAttributes<unknown>> & {
+  sizes?: typeof SIZES;
```
laviomri, line 89: I'm curious, is this optional because of the `Object.assign` at the bottom of the file? any way to workaround this? 
If this remains as is, I think anyone that wants to reference `Button.colors` will have to do `Button.color?.my_color`, and that's annoying..

src/components/Button/Button.tsx
```diff
@@ -172,7 +244,7 @@ const Button = forwardRef(
     const mergedRef = useMergeRefs({ refs: [ref, buttonRef] });
 
     const buttonProps = useMemo(() => {
-      return {
+      const props: Record<string, any> = {
```
laviomri, line 247: will this work?
```suggestion
      const props: React.HTMLProps<HTMLButtonElement> = {
```

src/components/Button/Button.tsx
```diff
@@ -275,7 +352,8 @@ const Button = forwardRef(
         {children}
         {rightIcon ? (
           <Icon
-            iconType={Icon.type.ICON_FONT}
+            // @ts-ignore
```
laviomri, line 355: why the ts-ignore? what's the hidden error?

src/components/Button/Button.tsx
laviomri: @hadasfa I think the issue was that storybook didn't detect the TS types, and was looking for `defaultProps` to understand the types (and show them in the props table). Worth checking this out, since I agree with @eran-cohen and @sahariko

src/components/Button/Button.tsx
hadasfa: agree, I would like to remove this. There is a specific reason to keep it (why we set this at first? @sahariko )

src/components/Button/Button.tsx
hadasfa: I was sure I add here a comment will recheck

src/components/Button/Button.tsx
hadasfa: I almost sure that in regular cases it will succeed to detect the default values, but either way it will not work for button composites (icon button etc) and dialog composites (tooltip etc...)
Will be checked as an option when I will convert a regular component

src/components/Button/Button.tsx
hadasfa: I think I found a way to remove the ? but not sure this is a conventional thing to do. Would be happy to hear what you think

src/components/Button/Button.tsx
hadasfa: Cause some issues on build. because it's not effect on the external API I prefer to keep it like it was and try to change it later (prefer it will not be a blocker)

src/components/Button/Button.tsx
hadasfa: The problem was as far asI understood, that we can't define a type which contain the enums as required and not set these enums while declare the button component value. I did not found a way to do it so I kept it like this for now. If you have any idea how to solve it I can change.

src/components/Button/Button.tsx
hadasfa: Nope :(. issues with some of the values. prefer it will not be a blocker in this stage.

src/components/Button/ButtonConstants.ts
hadasfa: added

eran-cohen (approved):
great milestone!

src/components/Icon/__tests__/all-icons-tests.jest.js
```diff
@@ -1,4 +1,4 @@
-import { exposeIcons } from "../../../../webpack/published-components";
+import { exposeIcons } from "utils/icon-utils";
```
orrgottlieb, line 1: alais ?

src/components/Slider/__tests__/slider-non-ranged-tests.jest.js
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { act } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { snapshotDiff } from "../../../../jest/utils";
+import { snapshotDiff } from "utils/jest-utils";
```
orrgottlieb, line 4: alias?

orrgottlieb (approved):

eran-cohen (approved):

PR #800, opened 2022-08-01
Fix h1 style gaps in Editable heading edit mode.

3dyonic:
https://monday.monday.com/boards/2861766393/pulses/3015744283

#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests

### DIFF
--- a/src/components/EditableInput/EditableInput.jsx
+++ b/src/components/EditableInput/EditableInput.jsx
@@ -52,7 +52,8 @@ const EditableInput = forwardRef(
       onCancelEditing,
       textareaSubmitOnEnter,
       onArrowKeyDown,
-      ariaLabel
+      ariaLabel,
+      brandFont
     },
     ref
   ) => {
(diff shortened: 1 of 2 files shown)


src/components/EditableInput/EditableInput.jsx
```diff
@@ -201,7 +202,8 @@ const EditableInput = forwardRef(
         id={id}
         style={style}
         className={cx("editable-input--wrapper", className, {
-          "no-resize": autoSize
+          "no-resize": autoSize,
+          "brand-font": brandFont
```
hadasfa, line 206: I guess the reason it's working is because we defined brand-font under heading? when we convert to css modules it will stop working and we will not understand why.
Maybe we will defined the css in heading in a way it will only works there and again in this component in a way it will only work here?

hadasfa (approved):

src/components/EditableInput/EditableInput.jsx
3dyonic: as we discussed classes are injected from the a global file so the issue is not relevant per now.

PR #808, opened 2022-08-02
Fix bug: displaying visual focus on dropdown mounted by mouse

hadasfa:
---

### DIFF
--- a/src/components/Combobox/Combobox.jsx
+++ b/src/components/Combobox/Combobox.jsx
@@ -145,32 +145,29 @@ const Combobox = forwardRef(
     useEffect(() => {
       const isFirstInputRender =
         (prevInputRef === undefined || inputRef.current === undefined) && inputRef.current !== undefined;
-      console.log(
-        prevInputRef,
-        inputRef,
-        "first render",
-        isFirstInputRender,
-        autoFocus && defaultVisualFocusFirstIndex && !mountedByKeyboard
-      );
       if (isFirstInputRender && autoFocus && defaultVisualFocusFirstIndex && !mountedByKeyboard) {
-        console.log("first");
-
         setShouldHideVisualFocusManually(true);
       }
     }, [autoFocus, defaultVisualFocusFirstIndex, mountedByKeyboard, prevInputRef, inputRef, setTriggeredByKeyboard]);
 
-    console.log("is it magic");
-    const onInputFocus = useCallback(() => {
-      console.log("whyyy");
-      if (shouldHideVisualFocusManually) {
+    const isInputFocused = document.activeElement === inputRef.current;
+
+    // First focus happens right after mounting. Sometimes using listen to focus event create race condition and this is why we listen to focus
+    // changes this way
+    useEffect(() => {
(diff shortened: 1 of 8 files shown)


src/components/Combobox/Combobox.jsx
```diff
@@ -49,6 +50,7 @@ const Combobox = forwardRef(
       renderOnlyVisibleOptions,
       clearFilterOnSelection,
       maxOptionsWithoutScroll,
+      mountedByKeyboard,
```
hadasfa, line 53: Combobox  can be auto focused when mounted. in the case of supporting displaying visual focus on first item, we will prefer to show it only if the trigger was keyboard.
We can't know it without prop because the mount trigger happened before our component actually... mount. (Is mounted is ok in terms of grammar and english?)

src/components/Combobox/Combobox.jsx
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable jsx-a11y/click-events-have-key-events */
-import React, { useRef, useState, forwardRef, useMemo, useCallback } from "react";
+import React, { useRef, useState, forwardRef, useMemo, useCallback, useEffect } from "react";
```
hadasfa, line 2: right now I don't have a way to test it in combobox because we did not implement any infra for testing visual focus there yet

src/hooks/useActiveDescendantListFocus/index.js
```diff
@@ -91,27 +104,33 @@ function useActiveDescendantListFocus({
...
   // we keep it for backward compatibility
   const backwardCompatibilityCreateOnClickCallback = useCallback(
     itemIndex => event => onItemClick(event, itemIndex),
     [onItemClick]
   );
-  console.log(triggeredByKeyboard);
   return {
-    visualFocusItemIndex: triggeredByKeyboard.current ? visualFocusItemIndex : undefined,
-    visualFocusItemId: triggeredByKeyboard.current ? visualFocusItemId : undefined,
+    visualFocusItemIndex: triggeredByKeyboard ? visualFocusItemIndex : undefined,
```
hadasfa, line 122: triggeredByKeyboard changed to be state instead of ref because we needed to trigger rerender

src/components/Combobox/Combobox.jsx
```diff
@@ -134,6 +137,49 @@ const Combobox = forwardRef(
...
+        filteredOptionsIds &&
+        visualFocusItemId === undefined &&
+        defaultVisualFocusFirstIndex
+      ) {
+        const nextVisuallyFocusId = filteredOptionsIds[0];
+        setVisualFocusItemId(nextVisuallyFocusId, true);
+      }
+    }, [filteredOptionsIds, visualFocusItemId, defaultVisualFocusFirstIndex, setVisualFocusItemId]);
+
+    // Hide visual focus when use click on clear button (we turn it on manually when user start typing which cause this edge case)
```
hadasfa, line 176:
```suggestion
    // Hide visual focus when user click on clear button (we turn it on manually when user start typing which cause this edge case)
```

PR #818, opened 2022-08-14
Fix icon button size

hadasfa:
צילום מסך 2022-08-14 ב-13 45 03

### DIFF
--- a/src/components/Button/Button.jsx
+++ b/src/components/Button/Button.jsx
@@ -8,7 +8,7 @@ import useMergeRefs from "../../hooks/useMergeRefs";
 import { NOOP } from "../../utils/function-utils";
 import Icon from "../../components/Icon/Icon";
 import Loader from "../../components/Loader/Loader";
-import { BUTTON_COLORS, BUTTON_INPUT_TYPE, BUTTON_TYPES, getActualSize } from "./ButtonConstants";
+import { BUTTON_COLORS, BUTTON_INPUT_TYPE, BUTTON_TYPES, getActualSize, BUTTON_ICON_SIZE } from "./ButtonConstants";
 import { getParentBackgroundColorNotTransparent, TRANSPARENT_COLOR } from "./helper/dom-helpers";
 import "./Button.scss";
 import { ELEMENT_TYPES, getTestId } from "../../utils/test-utils";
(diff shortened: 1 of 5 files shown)


src/components/Button/Button.scss
```diff
@@ -73,11 +73,11 @@
     }
   }
   .monday-style-button--left-icon {
-    padding-right: $spacing-small;
+    margin-right: var(--spacing-small);
```
hadasfa, line 76: because padding effect the actual size of the button icon (and then "24px" is not really 24px)

SergeyRoyt (approved):

src/components/Button/Button.jsx
```diff
@@ -217,17 +217,17 @@ const Button = forwardRef(
 
     const leftIconSize = useMemo(() => {
       if (typeof leftIcon !== "function") return;
-      return "24";
+      return "20";
```
sahariko, line 220: I see this number and a couple more being used in multiple components, why not add this to a constant?

src/components/IconButton/IconButton.jsx
```diff
@@ -44,9 +44,9 @@ const IconButton = forwardRef(
     const iconSize = useMemo(() => {
       if (size === IconButton.sizes.XXS) return 16;
       if (size === IconButton.sizes.XS) return 16;
-      if (size === IconButton.sizes.SMALL) return 24;
-      if (size === IconButton.sizes.MEDIUM) return 24;
-      if (size === IconButton.sizes.LARGE) return 24;
+      if (size === IconButton.sizes.SMALL) return 20;
+      if (size === IconButton.sizes.MEDIUM) return 20;
+      if (size === IconButton.sizes.LARGE) return 20;
       return 24;
```
sahariko, line 50: Consider using a switch-case:
```js
switch (size) {
  case IconButton.sizes.XXS:
  case IconButton.sizes.XS:
    return 16;
  case IconButton.sizes.SMALL:
  case IconButton.sizes.MEDIUM:
  case IconButton.sizes.LARGE:
    return 20;
  default:
    return 24;
}
```

hadasfa: it should be just a quick fix but I guess we can refactor it a bit

PR #823, opened 2022-08-16
Feature/sergeyro/css modules migration 1

SergeyRoyt:
- [ ] 1. Tab
- [ ] 2. TabPanel
- [ ] 3. TextWithHighlight
- [ ] 4. VirtualizedGrid
- [ ] 5. VirtualizedList


src/components/Tabs/Tab/Tab.jsx
```diff
@@ -1,14 +1,31 @@
...
+      value,
+      disabled,
+      active,
+      focus,
+      onClick,
+      icon,
+      iconType,
+      iconSide,
+      children,
+      "data-testid": dataTestId
```
hadasfa, line 25: the api will be data-testid or dataTestId?

src/components/Tabs/Tab/Tab.jsx
```diff
@@ -21,7 +38,7 @@ const Tab = forwardRef(
           ariaHidden={true}
           iconType={iconType}
           icon={icon}
-          className={cx("tab-icon", iconSide)}
+          className={cx(styles.tabIcon, "tab-icon", iconSide)}
```
hadasfa, line 41: We need to decide if we want to keep the old css classes or not @orrgottlieb @SergeyRoyt

src/components/TextWithHighlight/TextWithHighlight.module.scss
```diff
@@ -0,0 +1,18 @@
...
+  -webkit-box-orient: vertical;
+}
+
+.textWithHighlightWrapper em.highlightText {
+  @include theme-prop(background-color, primary-selected-color);
+  @include theme-prop(color, primary-text-color);
+  border-radius: 2px;
+}
+
+.textWithHighlightWrapper.withEllipsis {
```
hadasfa, line 16: no & in the script? I am not sure it's a problem but it can create a code which is less to edit is some cases, no?

src/components/VirtualizedGrid/VirtualizedGrid.jsx
```diff
@@ -198,7 +205,7 @@ const VirtualizedGrid = forwardRef(
                 rowCount={calcRowCount}
                 onScroll={onScrollCB}
                 onItemsRendered={onItemsRenderedCB}
-                className="virtualized-grid-scrollable-container"
+                className={cx("virtualized-grid-scrollable-container")}
```
hadasfa, line 208: why we did no added a css modules class name here?

src/components/Tabs/Tab/Tab.jsx
SergeyRoyt: "data-testid" everywhere as the same attribute can by used on html tags

src/components/TextWithHighlight/TextWithHighlight.module.scss
SergeyRoyt: Sometimes it's creating a really messy files... But @orrgottlieb wanted css files to be fully unwrapped. 
Also this functionality is implemented by independent plugin and it will require some time investments to modify this process, if we want to. Alternative solution would be to manually edit the especially messy files
Waiting for @orrgottlieb with this one

src/components/VirtualizedGrid/VirtualizedGrid.jsx
SergeyRoyt: If modular class wasn't added it means it wasn't found in the stylesheet - this class exists just for overrides

src/components/Tabs/Tab/Tab.jsx
SergeyRoyt: Let's wait for Orr and decide

SergeyRoyt: Orr suggestion was to keep old classes everywhere for now, complete the migration and then check whether removing old classname breaks something or not - if not then remove

src/components/TextWithHighlight/TextWithHighlight.module.scss
SergeyRoyt: Orr decision was to keep everything as it is, but to manually refactor messy stylesheets

src/components/VirtualizedGrid/VirtualizedGrid.jsx
3dyonic: i think we need to add a css modules class also, so when we remove the plain selectors we will still have classes per each element.

hadasfa: I agree with @3dyonic but approved for this not be a blocker.

hadasfa (approved):

src/components/VirtualizedGrid/VirtualizedGrid.jsx
SergeyRoyt: I believe it would be more clear and easier to checl if we leave it like this for now

SergeyRoyt:

PR #824, opened 2022-08-16
Feature/sergeyro/css modules migration 2

SergeyRoyt:
- [ ] 1. TabsContext
- [ ] 2. TabPanels
- [ ] 3. TabList
- [ ] 4. Tooltip
- [ ] 5. Toast
- [ ] 6. ToastLink


src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap
```diff
@@ -56,12 +56,12 @@ Snapshot Diff:
 +           class="monday-style-dialog-content-component top edge-center has-tooltip monday-style-animation-expand-appear-done monday-style-animation-expand-enter-done"
 +         >
 +           <div
-+             class="monday-style-tooltip monday-style-tooltip-dark padding-size-undefined"
++             class="mondayStyleTooltip monday-style-tooltip mondayStyleTooltipDark monday-style-tooltip-dark paddingSizeUndefined padding-size-undefined"
```
hadasfa, line 59: Maybe it's ok but "paddingSizeUndefined" looks to me like maybe there is an issue with converting the padding size to class name?
Worth double check?

src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap
```diff
@@ -56,12 +56,12 @@ Snapshot Diff:
...
 +           >
 +             20%
 +           </div>
 +           <div
-+             class="monday-style-tooltip-arrow monday-style-arrow monday-style-arrow-dark padding-size-undefined"
++             class="monday-style-tooltip-arrow mondayStyleArrow mondayStyleArrowDark paddingSizeUndefined monday-style-arrow monday-style-arrow-dark padding-size-undefined"
```
hadasfa, line 64: I understand it is a script but I think it's important to keep the class naming according to our conventions (aka without monday prefix)

src/components/Slider/__tests__/__snapshots__/slider-ranged-tests.jest.js.snap
```diff
@@ -67,12 +67,12 @@ Snapshot Diff:
 +           class="monday-style-dialog-content-component top edge-center has-tooltip monday-style-animation-expand-appear-done monday-style-animation-expand-enter-done"
 +         >
 +           <div
-+             class="monday-style-tooltip monday-style-tooltip-dark padding-size-undefined"
++             class="mondayStyleTooltip monday-style-tooltip mondayStyleTooltipDark monday-style-tooltip-dark paddingSizeUndefined padding-size-undefined"
```
hadasfa, line 70: same note about monday prefix

src/components/Tabs/TabList/TabList.module.scss
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.tabsWrapper.disabled {
+  opacity: 0;
+  height: 0;
+  padding-bottom: 5px;
+  pointer-events: none;
+}
+
+.tabsWrapper.disabled .tabsList .tabWrapper .tabInner:hover {
```
hadasfa, line 45: looks to me like a potential issue. tabWrapper and tabInner:hover are in different component and therefore a different scope. I think this will not work. Am I missing something?

src/components/Tabs/TabList/__tests__/__snapshots__/tabList-snapshot-tests.jest.js.snap
```diff
@@ -2,11 +2,12 @@
...
 exports[`TabList renders correctly stretched tab list 1`] = `
 <div
-  className="tabs--wrapper stretched"
+  className="tabsWrapper tabs--wrapper stretched"
+  data-testid="tab-list"
   id=""
 >
   <ul
-    className="tabs-list"
+    className="tabsList tabs-list undefined"
```
hadasfa, line 10: Is this undefined is because of jest&css module integration issues or is something is not working? What happen in prod?

src/components/Toast/Toast.jsx
```diff
@@ -68,7 +72,14 @@ const Toast = ({
       : null;
   }, [actions]);
   const classNames = useMemo(
-    () => cx("monday-style-toast", `monday-style-toast--type-${type}`, className),
+    () =>
+      cx(
+        styles.mondayStyleToast,
```
hadasfa, line 77: same note about monday prefix

src/components/Toast/Toast.module.scss
```diff
@@ -0,0 +1,111 @@
+@import "../../styles/global-css-settings";
+@import "../../styles/typography";
+@import "../../styles/themes";
+
+.mondayStyleToast {
```
hadasfa, line 5: monday prefix it's a problem :) 
But the fact the the structure of the file is more flat with the script actually work here and it's readable.

src/components/Tooltip/Tooltip.jsx
```diff
@@ -54,8 +56,13 @@ export default class Tooltip extends React.PureComponent {
     return (
       <div
         style={style}
-        className={classnames(
-          `monday-style-tooltip monday-style-tooltip-${theme} padding-size-${paddingSize}`,
+        className={cx(
+          styles.mondayStyleTooltip,
+          "monday-style-tooltip",
+          styles[`${camelCase("monday-style-tooltip-" + theme)}`],
```
hadasfa, line 62: why also converting "monday-style-tooltip-" to camel case?

src/components/Tooltip/Tooltip.jsx
```diff
@@ -136,14 +145,20 @@ export default class Tooltip extends React.PureComponent {
       content,
       getContainer: getContainer || this.getContainer,
       moveBy,
-      tooltipClassName: `monday-style-arrow monday-style-arrow-${theme} padding-size-${paddingSize}`,
+      tooltipClassName: cx(
+        styles[camelCase("monday-style-arrow")],
```
hadasfa, line 149: there is no easier way to perform these conversions (without convert already known constant to camel case every time?

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -1,11 +1,9 @@
 @import "../../styles/themes";
 @import "../../styles/typography";
 @import "../../styles/global-css-settings";
-
-
 $tooltip-z-index: 9999999999999999;
 
-.monday-style-tooltip {
+.mondayStyleTooltip {
```
hadasfa, line 6: monday style prefix is problematic in my opinion/

src/components/Tooltip/__tests__/__snapshots__/tooltip-snapshot-tests.jest.js.snap
```diff
@@ -28,14 +28,14 @@ Array [
       className="monday-style-dialog-content-component top edge-center has-tooltip"
     >
       <div
-        className="monday-style-tooltip monday-style-tooltip-dark padding-size-undefined"
+        className="mondayStyleTooltip monday-style-tooltip mondayStyleTooltipDark monday-style-tooltip-dark paddingSizeUndefined padding-size-undefined"
```
hadasfa, line 31: Is this an issue with jest&css module integration or a bug in the implementation?

src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap
SergeyRoyt: It's not a trouble of css modules conversion - cause previously it was like this `padding-size-undefined`, but I see how  it can be refactored - will do

src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap
SergeyRoyt: I'll try to replace it via script

src/components/Tooltip/Tooltip.jsx
SergeyRoyt: U mean we can do it like this or what?
```styles[`mondayStyleTooltip${camelCase(theme)}`]```

File: src/components/Tooltip/Tooltip.jsx (hunk 9 above)
Comment (SergeyRoyt) [reply src/components/Tooltip/Tooltip.jsx:149]: Script does that - that's the example of human error :) Script won't replace this prop cause it's custom prop.
Actually we don't even need this at all - such class doesn't exist in the stylesheet and serves only for overrides, so I removed this line `styles[camelCase("monday-style-arrow")]`

File: src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap (hunk 1 above)
Comment (SergeyRoyt) [reply src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap]: Refactored

File: src/components/Tabs/TabList/TabList.module.scss (hunk 4 above)
Comment (hadasfa) [reply src/components/Tabs/TabList/TabList.module.scss:45]: still wating for an answer. tab inner is out of scope and will contain different class name

Comment (SergeyRoyt) [reply src/components/Tabs/TabList/TabList.module.scss]: Fixed it with`:global`

File: src/components/Tabs/TabList/__tests__/__snapshots__/tabList-snapshot-tests.jest.js.snap (hunk 5 above)
Comment (SergeyRoyt) [reply src/components/Tabs/TabList/__tests__/__snapshots__/tabList-snapshot-tests.jest.js.snap]: It was due to `size` prop, which is undefined by default. I've fixed it

File: src/components/Toast/Toast.module.scss (hunk 7 above)
Comment (SergeyRoyt) [reply src/components/Toast/Toast.module.scss]: Removed

File: src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap (hunk 2 above)
Comment (SergeyRoyt) [reply src/components/Slider/__tests__/__snapshots__/slider-non-ranged-tests.jest.js.snap]: Replaced it

**Comment by SergeyRoyt** [comment]
@hadasfa this PR is ready for review

File: src/components/Tabs/TabList/TabList.module.scss [hunk 12]
```diff
@@ -0,0 +1,47 @@
...
+
+.tabsList.lg {
+  height: 48px;
+}
+
+.stretched .tabsList {
+  width: 100%;
+}
+
+.stretched .tabsList :global(.tab--wrapper) {
```
hadasfa, line 30: global is for using css modules global classes right? so why we are using a class name which not belong to css modules? how does it work? (tab--wrapper is with bem convetion which we do not use in css modules)

src/components/Tooltip/Tooltip.jsx
```diff
@@ -54,8 +56,15 @@ export default class Tooltip extends React.PureComponent {
...
         style={style}
-        className={classnames(
-          `monday-style-tooltip monday-style-tooltip-${theme} padding-size-${paddingSize}`,
+        className={cx(
+          styles.tooltip,
+          "monday-style-tooltip",
+          styles[`${camelCase("tooltip-" + theme)}`],
+          `monday-style-tooltip-${theme}`,
+          {
+            [styles[`${camelCase("padding-size-" + paddingSize)}`]]: paddingSize,
```
hadasfa, line 65: no solution to the camel case issue? why not paddingSize + camelCase(...)

hadasfa (approved):

SergeyRoyt:

PR #826, opened 2022-08-17
Feature/sergeyro/css modules migration 4

SergeyRoyt:
- [ ] 1. List
- [ ] 2. ListTitle
- [ ] 3. ListItem
- [ ] 4. ListItemIcon
- [ ] 5. Link
- [ ] 6. Label


src/components/Icon/hooks/useIconProps.js
```diff
@@ -34,8 +35,10 @@ export default function useIconProps({
   }, []);
 
   const computedClassName = useMemo(() => {
-    return cx("icon_component", className, {
+    return cx(styles.iconComponent, "icon_component", className, {
+      [styles.iconComponentClickable]: clickable,
```
hadasfa, line 39: I think here the prefix is not needed but it's not a blocker (clickable was a bit more readable than iconComponentClickable because we already have icon component class name as wrapper)

src/components/Link/Link.jsx
```diff
@@ -50,24 +52,27 @@ const Link = forwardRef(
...
+        {getIcon(!isStart, icon, cx(styles.iconEnd, "monday-style-link--icon-end"))}
       </a>
     );
   }
 );
 
 function getIcon(shouldShow, icon, className) {
   if (!shouldShow) return;
-  return <Icon className={className} clickable={false} icon={icon} iconType={Icon.type.ICON_FONT} />;
+  return <Icon className={cx(className)} clickable={false} icon={icon} iconType={Icon.type.ICON_FONT} />;
```
hadasfa, line 75: maybe we can remove the cx wrapper when only one class is passed to the class name?

hadasfa (approved):
I think it will be best to fix the wrapping with cx function when only one parameter is passed but after this there is no need in other review.

SergeyRoyt:

PR #828, opened 2022-08-18
AttentionBox - close icon button instead of icon

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3086884014

[Chromatic link](https://62037c72e0e4d6004a9a450f-hctauvthuj.chromatic.com/?path=/docs/feedback-attentionbox--overview)

image

- Need to pay attention to Icon size. It's 20 x 20 now, not 24 x 24 as it was before. This happens due to limitations of current IconButton implementation - you can get Icon size = 24 only if `size` prop of `IconButton` equals `undefined`, but that's really not ever going to happen, cause there is a default value for `size: IconButton.sizes.MEDIUM`. So probably need to fix this `IconButton` logic as well
- Should tooltip be displayed on button focus?
- Also some refactor was made to make further css modules automation easier :)

### DIFF
--- a/src/components/AttentionBox/AttentionBox.jsx
+++ b/src/components/AttentionBox/AttentionBox.jsx
@@ -2,12 +2,15 @@ import React, { useMemo } from "react";
 import PropTypes from "prop-types";
 import cx from "classnames";
 import Icon from "../../components/Icon/Icon";
+import IconButton from "../IconButton/IconButton";
 import CloseSmall from "../../components/Icon/Icons/components/CloseSmall";
 import AlertIcon from "../../components/Icon/Icons/components/Alert";
 import { ICON_TYPES } from "../../components/Icon/IconConstants";
 import { backwardCompatibilityForProperties } from "../../helpers/backwardCompatibilityForProperties";
-import { baseClassName, closeClassName, compactClassName, ATTENTION_BOX_TYPES } from "./AttentionBoxConstants";
+import { ATTENTION_BOX_TYPES } from "./AttentionBoxConstants";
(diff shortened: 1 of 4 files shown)


src/components/AttentionBox/AttentionBox.jsx
```diff
@@ -82,22 +90,27 @@ const AttentionBox = ({
...
         {text}
       </div>
       {onClose && (
-        <Icon
-          iconType={Icon.type.SVG}
-          iconLabel="Close"
+        <IconButton
+          size={IconButton.sizes.SMALL}
+          color={IconButton.colors.ON_PRIMARY_COLOR}
+          className={cx(`${ATTENTION_BOX_CSS_BASE_CLASS}__close-icon`, "icon")}
```
hadasfa, line 107: why icon class name?

hadasfa: it's super generic class name

SergeyRoyt: `wrapper` is also super generic class name)
But is there a difference if it's nested under `close-icon` rule? It can't be used without it, no?

hadasfa: I did not see a wrapper class name but it is also bad if it's exist.
Let's not use super generic class names before converting to css modules because there is always a change it will cause a weird behavior (even if it's nested maybe other deveper use the same class name and you will get some of his defentions)

SergeyRoyt: Fixed

PR #829, opened 2022-08-18
Feature/sergeyro/dropdown item tooltip

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3107782229

[Chromatic link](https://62037c72e0e4d6004a9a450f-zwevtncmhh.chromatic.com/)

image


src/components/Dropdown/Dropdown.jsx
```diff
@@ -229,7 +229,7 @@ const Dropdown = ({
         Menu,
         ClearIndicator,
         Input,
-        ...(finalOptionRenderer && { Option }),
+        Option,
```
hadasfa, line 232: why we removed finalOptionRenderer ?

src/components/Dropdown/__stories__/dropdown.stories.mdx
```diff
@@ -145,124 +142,180 @@ The Dropdown component supports multiple options selection in two different stat
...
+        []
+      );
+      return (
+        <StoryDescription vertical>
+          <div>
             <Dropdown
-              placeholder="Multiple line multi state"
-              defaultValue={[options[0]]}
-              options={options}
+              defaultValue={[optionsAvatar[0]]}
```
hadasfa, line 238: is this change by purpose?

src/components/Dropdown/Dropdown.jsx
hadasfa, line 232: is this by purpose? why?

src/components/Dropdown/components/option/option.jsx
```diff
@@ -1,16 +1,24 @@
 /* eslint-disable react/jsx-props-no-spreading */
 import React from "react";
 import { components } from "react-select";
+import TooltipWrapper from "../../../Tooltip/TooltipWrapper";
+import Tooltip from "../../../Tooltip/Tooltip";
 import "./option.scss";
 
-const Option = ({ Renderer, ...props }) => {
-  const { data } = props;
-  if (!Renderer) return null;
```
hadasfa, line 8: why removed?

src/components/Dropdown/components/option/option.jsx
```diff
@@ -1,16 +1,24 @@
...
+    <TooltipWrapper
+      wrapWithTooltip={!!data?.tooltipProps}
+      tooltipProps={{ position: Tooltip.positions.RIGHT, ...data.tooltipProps }}
+    >
+      {Renderer ? (
+        <components.Option {...props} className="dropdown-wrapper__option--reset">
+          <Renderer {...data} />
+        </components.Option>
+      ) : (
+        <components.Option {...props} className="dropdown-wrapper__option--reset" />
```
hadasfa, line 19: before the change it was null, why now it's changed?

src/components/Tooltip/TooltipWrapper.jsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+import PropTypes from "prop-types";
+import Tooltip from "./Tooltip";
+
+const TooltipWrapper = ({ children, wrapWithTooltip, tooltipProps }) => {
+  if (!wrapWithTooltip) {
```
hadasfa, line 6: not really need a wrapper for this. you can just set the component according to condition and set all the props which related to tooltip only if this condition apply

src/components/Dropdown/Dropdown.jsx
SergeyRoyt: Here all custom element renders are specified. Previously the logic was like this: if `optionRenderer` has been passed then use this custom function to render option, otherwise use default renderer. Now it's always using custom renderer, cause it's the only option to add Tooltip wrapper for the option. And now `options.jsx` option-renderer is the entry point and it's responsible for deciding how specific option should be rendered

src/components/Tooltip/TooltipWrapper.jsx
SergeyRoyt: Does this look better?
image


I've made it in analogy with ClickableWrapper, which I've found rather useful, thought it's not necessary at all

src/components/Dropdown/components/option/option.jsx
SergeyRoyt: Responded in prev comment

src/components/Tooltip/TooltipWrapper.jsx
hadasfa: I like this approach more  :)

src/components/Dropdown/components/option/option.jsx
hadasfa: not a must but it can be more readable here by doing something like this instead of condition:
```
 <components.Option {...props} className="dropdown-wrapper__option--reset">
          {Renderer ? <Renderer {...data} /> : undefined}
</components.Option>

I did not use here Renderer && <Renderer {...data} /> because did not wanted to pass false as child
```

src/components/Dropdown/Dropdown.jsx
hadasfa: I just not find here where we pass finalOptionRenderer to the option component so I am not sure I am understand. can you show me?

src/components/Dropdown/components/option/option.jsx
SergeyRoyt: Tried exactly what u r proposing, but it resulted in this
image

And I'm not sure why, cause I thought that if u r passing `undefined` in place of `children` it should work the same as when u r not passing children at all. But here for some reason it works in a different way - maybe due to some `react-select` implementation details

src/components/Tooltip/TooltipWrapper.jsx
SergeyRoyt: Done :)

hadasfa (approved):

PR #831, opened 2022-08-21
IconButton: add showTooltip prop

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3118525928
Tried to search for such use cases to keep consistent, but haven't found anything like it - @hadasfa if u know that it has a different name somewhere, pls tell me:)

### DIFF
--- a/src/components/IconButton/IconButton.jsx
+++ b/src/components/IconButton/IconButton.jsx
@@ -23,7 +23,7 @@ const IconButton = forwardRef(
       size,
       tooltipContent,
       ariaLabel,
-      showTooltip,
+      hideTooltip,
       kind,
       active,
       disabled,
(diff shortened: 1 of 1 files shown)


src/components/IconButton/IconButton.jsx
```diff
@@ -23,6 +23,7 @@ const IconButton = forwardRef(
       size,
       tooltipContent,
       ariaLabel,
+      showTooltip,
```
hadasfa, line 26: let's change it to hideTooltip because we prefer always use negative value as default in optional props?

SergeyRoyt: Changed 👍

hadasfa (approved):

PR #837, opened 2022-08-22
Feature/sergeyro/checkbox aria disabled

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3120699386


SergeyRoyt:
@hadasfa, Talyah said that this is not needed - u were right! :) 
Closing the PR

PR #839, opened 2022-08-23
Feature/sergeyro/accordion collapse icon not clickable

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3129618502

ExpandCollapse: added `ignoreFocusStyle` prop to the Icon

### DIFF
--- a/src/components/ExpandCollapse/ExpandCollapse.jsx
+++ b/src/components/ExpandCollapse/ExpandCollapse.jsx
@@ -41,9 +41,7 @@ const ExpandCollapse = forwardRef(
               iconType={Icon.type.SVG}
               icon={DropdownChevronDown}
               iconSize={iconSize}
-              tabindex="-1"
               clickable={false}
-              ignoreFocusStyle
             />
           </button>
           {isExpanded && (
--- a/src/components/ExpandCollapse/ExpandCollapse.jsx
+++ b/src/components/ExpandCollapse/ExpandCollapse.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef, forwardRef, useCallback } from "react";
+import React, { forwardRef, useCallback, useRef, useState } from "react";
 import PropTypes from "prop-types";
 import cx from "classnames";
 import useMergeRefs from "../../hooks/useMergeRefs";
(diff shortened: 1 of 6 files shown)


src/components/Icon/Icon.jsx
```diff
@@ -133,14 +130,11 @@ Icon.propTypes = {
 };
 
 Icon.defaultProps = {
-  onClick: NOOP,
   className: "",
   icon: "",
-  clickable: true,
```
orrgottlieb, line 139: @SergeyRoyt this is breaking change - please add this to the backlog and consider it when upgrading major

orrgottlieb (changes requested):

hadasfa (approved):

hadasfa (changes requested):
I saw what orr wrote, let's split this from the clickable is false by default pr for being able to merge it

src/components/ExpandCollapse/ExpandCollapse.jsx
```diff
@@ -43,6 +43,7 @@ const ExpandCollapse = forwardRef(
               iconSize={iconSize}
               tabindex="-1"
               clickable={false}
+              ignoreFocusStyle
```
hadasfa, line 46: this is the wrong fix. I saw you need to pass it even clickable false so I understand the issue is that the accordion icon have focus even it is not a button (clickable=false).
the reason for the focus beahvior which we don't want to allow is tabindex="-1".
we must remove it because we do not want to allow focus on only decorative element (if you don't sure why this is the case let's talk)

hadasfa (changes requested):
please see note

src/components/ExpandCollapse/ExpandCollapse.jsx
SergeyRoyt: Let's talk about this one to clarify a bit

SergeyRoyt: Removed both `tabIndex` and `ignoreFocusStyle`

hadasfa (approved):

PR #847, opened 2022-08-30
Export content color

or-nuri-monday:
Exporting content colors vars names


laviomri (approved):
LGTM - @hadasfa please take a look too

PR #848, opened 2022-08-30
Support new dropdown state for displaying menu inside scrollable dialog

hadasfa:
צילום מסך 2022-08-30 ב-18 08 18

I did not had any idea for tests which relevant to add for this.

### DIFF
--- a/src/components/Dropdown/Dropdown.jsx
+++ b/src/components/Dropdown/Dropdown.jsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/require-default-props,react/forbid-prop-types */
 import { SIZES } from "../../constants/sizes";
-import React, { useCallback, useMemo, useState } from "react";
+import React, { useCallback, useMemo, useRef, useState } from "react";
 import Select, { components } from "react-select";
 import AsyncSelect from "react-select/async";
 import NOOP from "lodash/noop";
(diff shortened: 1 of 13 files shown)


src/components/Dropdown/Dropdown.jsx
```diff
@@ -129,8 +132,20 @@ const Dropdown = ({
       });
     }
 
+    mergedStyles.menu = provided => {
+      if (!insideScroll) return provided;
+
+      // If inside a scroll try to get dropdown location at the dom
```
hadasfa, line 138:
```suggestion
      // If inside a scroll, we try to get dropdown location at the dom
```

src/components/Dropdown/components/Control/Control.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import { components } from "react-select";
+
+const Control = props => {
+  const { selectProps } = props;
+  const control = <components.Control {...props} />;
+  const controlRef = selectProps?.selectProps?.controlRef;
+  if (controlRef)
+    return (
+      <div className="monday-dropdown_scrollable-wrapper" ref={controlRef}>
```
hadasfa, line 10: did not use css modules here because class name is only for readability purpose. can change and add empty css module file

PR #850, opened 2022-08-31
Feature/sergeyro/css modules migration 10

SergeyRoyt:
- [ ] 1. Accordion
- [ ] 2. AttentionBox
- [ ] 3. Avatar
- [ ] 4. AvatarBadge
- [ ] 5. AvatarContent


src/components/Accordion/Accordion/Accordion.module.scss
```diff
@@ -0,0 +1,23 @@
...
+.accordion {
+  height: fit-content;
+  border: 1px solid;
+  @include theme-prop(border-color, ui-border-color);
+  box-sizing: border-box;
+  border-radius: $border-radius-small;
+}
+
+:global(.accordion-item) :global(.expand-collapse--wrapper) :global(.expand-collapse) {
+  border-radius: 0;
```
hadasfa, line 14: it's a bit weird because expand collapse and accordion should look the same (only the behavior is different a bit)
can you show to meirav how its looks in expand-collapse vs accordion and ask if its ok to apply the changed to expand-collapse

src/components/AttentionBox/AttentionBox.jsx
```diff
@@ -44,23 +48,36 @@ const AttentionBox = ({
   return (
     <aside
       className={cx(
+        styles.box,
```
hadasfa, line 51: let's change the name to attentionBox because box is a different component and it can be confusing in debug and investigate issues

src/components/AttentionBox/AttentionBox.module.scss
```diff
@@ -0,0 +1,88 @@
+@import "../../styles/themes";
+@import "../../styles/typography";
+@import "../../styles/global-css-settings";
+
+.box {
```
hadasfa, line 5: comment about change box to attentionBox

src/components/Avatar/Avatar.jsx
```diff
@@ -139,11 +165,20 @@ const Avatar = ({
...
-              [bemHelper({ element: "circle", state: "is-square" })]: overrideSquare,
-              [bemHelper({ element: "circle", state: "without-border" })]: withoutBorder
-            })}
+            className={cx(
+              styles.circle,
+              "monday-style-avatar_circle",
+              styles[`${camelCase("circle--" + type)}`],
+              `monday-style-avatar_circle--${type}`,
+              {
+                [styles.circleIsDisabled]: overrideDisabled,
```
hadasfa, line 174: please switch naming to disabled, square, withoutBorder.... (the circle prefix not needed and create weird names such as circleIsSquare)

src/components/Avatar/AvatarContent.jsx
```diff
@@ -1,32 +1,66 @@
...
+      );
     case AVATAR_TYPES.ICON:
       return (
         <Icon
           icon={icon}
           aria-label={ariaLabel}
           role={role}
           clickable={false}
-          className={className}
+          className={cx(className)}
```
hadasfa, line 49: all these not needed cx :(. it's not a must but it will be better to remove these.

src/components/Accordion/Accordion/Accordion.module.scss
SergeyRoyt: Hmm, not sure what u mean...
We don't have any other usages of `ExpandCollapse` other than in `AccordionItem`, but `ExpandCollapse` is a published component, so it can be used by someone outside of Vibe - therefore I think we must be careful changing these overrides and transmitting them to ExpandCollapse e.g.

and without these overrides it'll look like this (the border), so I guess there is no need to consult with Meirav 
image

src/components/AttentionBox/AttentionBox.jsx
```diff
@@ -1,14 +1,16 @@
+import { camelCase } from "lodash";
```
orrgottlieb, line 1: `import camelCase from "lodash/camelCase"`

src/components/Avatar/Avatar.jsx
```diff
@@ -1,8 +1,9 @@
+import { camelCase } from "lodash";
```
orrgottlieb, line 1: same comment - lets fix it in the merged PR @SergeyRoyt

orrgottlieb (approved):

SergeyRoyt:
- [ ] TODO need to fix this when all PRs are merged https://github.com/mondaycom/monday-ui-react-core/pull/850#discussion_r973726536

PR #854, opened 2022-09-04
add support for menus inside a container which using transform function

hadasfa:
צילום מסך 2022-09-04 ב-13 38 46

צילום מסך 2022-09-04 ב-13 38 59

### DIFF
--- a/src/components/Dropdown/Dropdown.jsx
+++ b/src/components/Dropdown/Dropdown.jsx
@@ -64,7 +64,8 @@ const Dropdown = ({
   ref,
   withMandatoryDefaultOptions,
   isOptionSelected,
-  insideScroll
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.jsx
```diff
@@ -64,7 +64,8 @@ const Dropdown = ({
   ref,
   withMandatoryDefaultOptions,
   isOptionSelected,
-  insideScroll
+  insideOverflowContainer,
```
hadasfa, line 67: i know i changed the name but i guess it's ok because it was released only from Wednesday... I guess no one is start to used it already

orrgottlieb (approved):

PR #857, opened 2022-09-06
Convert icon component and related code to js

hadasfa:


src/StoryBookComponents/StoryLink/StoryLine.scss
```diff
@@ -1,3 +0,0 @@
-.wrapperClassName {
```
hadasfa, line 1: not used

src/StoryBookComponents/StoryLink/StoryLine.jsx
```diff
@@ -1,19 +0,0 @@
-import React from "react";
-import cx from "classnames";
```
hadasfa, line 2: not used

package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.61.0",
   "description": "Official monday.com UI resources for application development in React.js",
   "main": "./dist/main.js",
+  "types": "./dist/types.d.ts",
```
sahariko, line 6: Cannot be added until all components have types.

src/components/Icon/CustomSvgIcon/CustomSvgIcon.tsx
```diff
@@ -1,25 +1,39 @@
...
+  clickable?: boolean;
+  replaceToCurrentColor?: boolean;
+  customColor?: string;
+  ref?: Ref<SVGElement>;
+}
+
+const CustomSvgIcon: FunctionComponent<CustomSvgIconProps> = ({
+  className = "",
+  ref,
+  src = "",
```
sahariko, line 29: It's required, so a default value doesn't make sense.

src/components/Icon/CustomSvgIcon/CustomSvgIcon.tsx
```diff
@@ -1,25 +1,39 @@
...
+  role?: AriaRole;
+  ariaHidden?: boolean;
+  clickable?: boolean;
+  replaceToCurrentColor?: boolean;
+  customColor?: string;
+  ref?: Ref<SVGElement>;
+}
+
+const CustomSvgIcon: FunctionComponent<CustomSvgIconProps> = ({
+  className = "",
```
sahariko, line 27: No need for default value, `classnames` ignores `undefined`.

src/components/Icon/FontIcon/FontIcon.tsx
```diff
@@ -0,0 +1,55 @@
...
+      icon: Icon,
+      role = "img",
+      "aria-hidden": ariaHidden,
+      "data-testid": dataTestId
+    },
+    iconRef
+  ) => {
+    const isIconFunction = typeof Icon === "function";
+    const iconClassName = isIconFunction ? "" : Icon;
+    // @ts-ignore
```
sahariko, line 32: Why?

src/components/Icon/FontIcon/FontIcon.tsx
```diff
@@ -0,0 +1,55 @@
...
+        ref={iconRef}
+        aria-label={iconLabel}
+        tabIndex={tabIndex}
+        role={role}
+        id={id}
+        data-testid={dataTestId}
+      >
+        {isIconFunction && (
+          //Icon components are not converted to typescript in this stage
+          //@ts-ignore
```
sahariko, line 48: Why?

src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,150 @@
...
+      id = "",
+      onClick = CLICK_NOOP,
+      /**
+       * class name to be added to icon
+       */
+      className = "",
+      /**
+       * we support three types of icons - SVG, FONT and SRC (classname) so this prop is either the name of the icon or the component
+       */
+      icon = "",
```
sahariko, line 46: Makes no sense, since this is required.

src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,150 @@
...
+    {
+      /**
+       * component id
+       */
+      id = "",
+      onClick = CLICK_NOOP,
+      /**
+       * class name to be added to icon
+       */
+      className = "",
```
sahariko, line 42: No need.

src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,150 @@
...
+    if (!icon) {
+      return null;
+    }
+
+    const isFunctionType = typeof icon === "function";
+    if (iconType === IconType.SVG || isFunctionType || typeof icon === "object") {
+      const IconComponent = icon;
+      return (
+        // The icons are not converted to ts for js yet.
+        // @ts-ignore
```
sahariko, line 107: Why?

src/components/Icon/__tests__/all-icons-tests.jest.ts
```diff
@@ -1,17 +1,21 @@
...
   it("should contains all icons", () => {
     const allIconFiles = exposeIcons();
-    const missingIcons = [];
+    const missingIcons: Array<string> = [];
     Object.keys(allIconFiles).forEach(iconPath => {
       const iconName = arrayLast(iconPath.split("/"));
+      /**
+       * TS thinks allIcons is an array but its an object which contains all icons and that's why it's work when iconName is string
+       */
+      // @ts-ignore
```
sahariko, line 18: The convention is to write the reason on the same line as the `@ts-ignore` directive.

sahariko: This is strange, TS should figure out that `import * as` returns an object. What's the exact error here?

src/components/Icon/hooks/useIconProps.tsx
```diff
@@ -16,20 +16,30 @@ export default function useIconProps({
...
   const iconRef = useRef(null);
   const onEnterCallback = useCallback(
-    event => {
+    (event: KeyboardEvent) => {
       const isActive = document.activeElement === iconRef.current;
       if (!isActive) {
         return;
       }
+      // Keyboard event is different from mouse event
+      //@ts-ignore
```
sahariko, line 36: Let's fix this the TS way instead of ignoring.

src/storybook/functions/create-component-story.js
```diff
@@ -45,6 +45,7 @@ export function createStoryMetaSettings({ component, enumPropNamesArray, iconPro
     if (prop instanceof Object) {
       enumName = prop.enumName;
       prop = prop.propName;
+      debugger;
```
sahariko, line 48: 👀

sahariko (changes requested):

package.json
hadasfa: yes it was added by mistake will removed

src/components/Icon/CustomSvgIcon/CustomSvgIcon.tsx
hadasfa: I know, added for readability but I will remove

src/components/Icon/FontIcon/FontIcon.tsx
hadasfa: as I wrote in the comment:           //Icon components are not converted to typescript in this stage
this is why

src/components/Icon/Icon.tsx
hadasfa: as written in the comment:         // The icons are not converted to ts for js yet.

hadasfa: I mean the icon components which contains vibe svgs and created by the script (there is another pr for this which not merged yet)

src/components/Icon/__tests__/all-icons-tests.jest.ts
hadasfa: Yes I did not understand the error... I fixed the comment. please tell me if it is not enough.

src/components/Icon/FontIcon/FontIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+        onClick={onClick}
+        ref={iconRef}
+        aria-label={iconLabel}
+        tabIndex={tabIndex}
+        role={role}
+        id={id}
+        data-testid={dataTestId}
+      >
+        {isIconFunction && (
+          //@ts-ignore: Icon components are not converted to typescript in this stage
```
sahariko, line 46: I don't think I understand.

`Icon` ultimately comes from the prop `icon`, which is defined as `string` or `React.Component`, this shouldn't have anything to do with components conversion.

What's the original error here?

src/components/Icon/FontIcon/FontIcon.tsx
```diff
@@ -0,0 +1,53 @@
...
+      "aria-label": iconLabel,
+      tabIndex,
+      icon: Icon,
+      role = "img",
+      "aria-hidden": ariaHidden,
+      "data-testid": dataTestId
+    },
+    iconRef
+  ) => {
+    const isIconFunction = typeof Icon === "function";
```
sahariko, line 30: Are React components functions? If not, we should adjust types.

I think that React has a utility function for checking if something is a React component or not.

src/components/Icon/Icon.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+    if (!icon) {
+      return null;
+    }
+
+    const isFunctionType = typeof icon === "function";
+    if (iconType === IconType.SVG || isFunctionType || typeof icon === "object") {
+      const IconComponent = icon;
+      return (
+        // @ts-ignore: icons are not converted to ts for js yet.
```
sahariko, line 106: Same comment as the one about `<Icon/>`.

src/components/Icon/FontIcon/FontIcon.tsx
hadasfa: You were right I just used the wrong type, fixed it.
thanks!

src/components/Icon/FontIcon/FontIcon.tsx
hadasfa: It should be only a function component but i am afraid of breaking changes in weird cases (for example creating a custom renderer in a use callback hook... not sure it will count as function component)... developers really abuse this icon prop and I prefer to be safe

hadasfa (commented):
comment is not relevant anymore therefore deleted

sahariko (approved):

PR #858, opened 2022-09-06
Css modules typescript infra via 'typescript-plugin-css-modules' + eslint ts specification

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3191529916


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.59.1",
   "description": "Official monday.com UI resources for application development in React.js",
   "main": "./dist/main.js",
+  "types": "./dist/types.d.ts",
```
orrgottlieb, line 6: this was removed by design

orrgottlieb (changes requested):

hadasfa (changes requested):
What Orr comment + it's looks like the new es lint settings cause a little bit a mess.
for example: now it's ignore _ when defined not required parameter in function and ignore // eslint-disable-next-line no-unused-vars because now I guess we don't allow it by different settings? let's make sure all the eslint errors are fixed before merging.

package.json
SergeyRoyt: Removed 👍

src/components/Switch/index.js
```diff
@@ -37,7 +38,7 @@ export function useHiddenSwitch({
     isChecked: overrideDefaultChecked || checked,
     HiddenSwitch: (
       <input
-        ref={ref}
+        ref={inputRef}
```
orrgottlieb, line 41: this might be breaking change, please use mergedRef

.eslintrc.js
```diff
@@ -6,12 +63,32 @@ module.exports = {
...
+        "@typescript-eslint/no-unused-vars": ["error", { argsIgnorePattern: "^_", varsIgnorePattern: "^_" }]
+      }
     }
   ],
   env: {
     browser: true,
     es2021: true
   },
+  globals: {
+    require: true,
```
orrgottlieb, line 89: why do we need require ?

src/components/Switch/index.js
orrgottlieb: or was it a bug?

.eslintrc.js
```diff
@@ -6,12 +63,32 @@ module.exports = {
...
+    module: true
```
orrgottlieb, line 90: why do we need module as well?

src/components/Switch/index.js
SergeyRoyt: It was a bug. `ref` was used, though it was not defined

.eslintrc.js
SergeyRoyt: image
Without it I get this kind of errors

orrgottlieb: we shouldn't use them in our code, so can you please ignore them in these specific files - (these files are for scripts)

SergeyRoyt: Done 👍

orrgottlieb (approved):

PR #867, opened 2022-09-11
Convert clickable to tsx (and related)

hadasfa:


src/components/Button/__tests__/button.jest.tsx
```diff
@@ -1,18 +1,25 @@
 import React from "react";
-import { render, fireEvent, cleanup } from "@testing-library/react";
-import Button from "../Button";
+import { render, fireEvent, cleanup, RenderResult } from "@testing-library/react";
+import Button, { ButtonProps } from "../Button";
 
 const text = "Click Me!";
 const className = "test-class";
+const mockOnClick = jest.fn();
```
laviomri, line 7: this might be dangerous - the mock's state is never reset between tests. Consider calling `jest.resetAllMocks()` in an `afterEach`, or creating a new mock yourself in a `beforeEach`

src/components/Clickable/Clickable.tsx
```diff
@@ -0,0 +1,91 @@
+/* eslint-disable jsx-a11y/no-noninteractive-tabindex */
```
laviomri, line 1: how many of these errors did you have in this file? perhaps consider moving this `eslint-disable` to each relevant line (assuming there aren't too many)?

src/components/Clickable/Clickable.tsx
```diff
@@ -0,0 +1,91 @@
...
+  children?: string | ReactNode | null | [ReactNode];
+  role?: AriaRole;
+  enableTextSelection?: boolean;
+  onMouseDown?: MouseEventHandler<HTMLDivElement>;
+  ariaHidden?: boolean;
+  ariaLabel?: string;
+  tabIndex?: string | number;
+  style?: CSSProperties;
+  /**
+   *     Backward compatibility for props naming - please do not use
```
laviomri, line 27: Consider notating it - https://jsdoc.app/tags-deprecated.html
```suggestion
   * @deprecated Backward compatibility for props naming - please do not use
```

src/hooks/useKeyboardButtonPressedFunc.js
```diff
@@ -6,7 +6,7 @@ const KEYDOWN_ENTER_EVENT = 13;
 export function useKeyboardButtonPressedFunc(onClick) {
   return useCallback(
     e => {
-      if (e.keyCode === KEYDOWN_SPACE_EVENT || e.keyCode === KEYDOWN_ENTER_EVENT) {
+      if ((onClick && e.keyCode === KEYDOWN_SPACE_EVENT) || e.keyCode === KEYDOWN_ENTER_EVENT) {
         onClick(e);
```
laviomri, line 10: something seems a bit off - if `onClick === undefined`, and `e.keyCode === KEYDOWN_ENTER_EVENT`, we will call `onClick(e)` even though `onClick` is undefined. Didn't TypeScript shout on this?
in any case, I think you can do the following -
```suggestion
      if ([KEYDOWN_ENTER_EVENT, KEYDOWN_SPACE_EVENT].includes(e.keyCode)) {
        onClick?.(e);
```

src/types/VibeBaseButtonComponentProps.ts
```diff
@@ -0,0 +1,15 @@
...
+  /** Callback function to run when the button is clicked */
+  onClick?: MouseEventHandler<HTMLElementType>;
+  /** aria controls - receives id for the controlled region */
+  ariaControls?: string;
+  /** Whether the button should be disabled or not */
+  disabled?: boolean;
+  /** aria to be set if the popup is open */
+  ariaHasPopup?: React.HTMLProps<HTMLElementType>["aria-haspopup"];
+  /** aria to be set if the popup is open */
+  ariaExpanded?: boolean;
```
laviomri, line 14: making sure - do both `ariaHasPopup` and `ariaExpanded` have the same description?
If so, should they always have the same value?
And, if so, why do we need both of them at the same time?

hadasfa: I Admit I copy paste old description which was not accurate, will be fixed

src/components/Button/ButtonConstants.ts
```diff
@@ -25,10 +25,9 @@ export type Size = typeof SIZES[keyof typeof SIZES] | keyof typeof OLD_BUTTON_SI
 // Support old sizes (backward compatible)
 export const getActualSize = (size?: Size) => {
   if (size && size in OLD_BUTTON_SIZES) {
-    //  Element implicitly has an 'any' type because expression of type 'Size' can't be used to index type
-    //  '{ readonly sm: "small"; readonly md: "medium"; readonly lg: "large"; }'.
-    //   Property 'xxs' does not exist on type '{ readonly sm: "small"; readonly md: "medium"; readonly lg: "large"; }'.
-    // @ts-expect-error
+    // @ts-expect-error: Element implicitly has an 'any' type because expression of type 'Size' can't be used to index type
+    //     //  '{ readonly sm: "small"; readonly md: "medium"; readonly lg: "large"; }'.
+    //     //   Property 'xxs' does not exist on type '{ readonly sm: "small"; readonly md: "medium"; readonly lg: "large"; }'.
```
sahariko, line 30: The reason we add to `@ts-expect-error`s is an opportunity to explain in human language why we put it there, copying the TS error in its entirety kinda misses that point IMO.

src/components/Clickable/Clickable.tsx
```diff
@@ -0,0 +1,91 @@
...
+import VibeBaseButtonComponentProps from "../../types/VibeBaseButtonComponentProps";
+import VibeComponent from "../../types/VibeComponent";
+import "./Clickable.scss";
+
+const CSS_BASE_CLASS = "monday-style-clickable";
+const bemHelper = BEMClass(CSS_BASE_CLASS);
+
+export interface ClickableProps extends VibeBaseButtonComponentProps<HTMLElement> {
+  elementType?: ElementType;
+  children?: string | ReactNode | null | [ReactNode];
```
sahariko, line 18: Did you mean array of `ReactNode`?

```suggestion
  children?: string | ReactNode | null | ReactNode[];
```

What you wrote originally means "an array with a single item that is a `ReactNode`", meaning if an array of 2 or more `ReactNode`s is passed, TS would throw an error.

src/components/Clickable/Clickable.tsx
```diff
@@ -0,0 +1,91 @@
...
+      tabIndex = 0,
+      disabled,
+      style,
+      // Backward compatibility for props naming
+      dataTestId,
+      "data-testid": dataTestIdUpdated
+    },
+    ref
+  ) => {
+    const overrideTabIndex = tabIndex && +tabIndex;
```
sahariko, line 56: This line of code is super duper confusing.

PR #870, opened 2022-09-13
Feature/sergeyro/ts infra css modules

SergeyRoyt:
1) Fix jest errors which happened due to `esModuleInterop` absence
2) tsconfig: include src/* to fix TS2307: Module not found "./*.module.scss"
3) Fix `noImplicitlyAny` on modular styles string index accessors in typescript files - via helper getStyle function
4) Fix eslint errors about using node global variables in js files
5) tsconfig: remove `ts-esm.js` from included

### DIFF
--- a/src/helpers/typesciptCssModulesHelper.ts
+++ b/src/helpers/typesciptCssModulesHelper.ts
@@ -3,6 +3,6 @@
  * @param styles modular styles object
  * @param key string classname
  */
-export function getStyle<StylesType>(styles: StylesType, key: string): StylesType[keyof StylesType] | undefined {
+export function getStyle<StylesType>(styles: StylesType, key: string) {
   return styles[key as keyof typeof styles];
 }
--- a/src/helpers/typesciptCssModulesHelper.ts
+++ b/src/helpers/typesciptCssModulesHelper.ts
@@ -1,0 +1,8 @@
+/**
+ * Return style by key - solves to fix noImplicitAny errors when referencing modular styles from ts files via index accessor
+ * @param styles modular styles object
+ * @param key string classname
+ */
+export function getStyle<StylesType>(styles: StylesType, key: string): StylesType[keyof StylesType] | undefined {
+  return styles[key as keyof typeof styles];
+}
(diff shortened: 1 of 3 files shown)


SergeyRoyt:
Talked with @sahariko about TS2307 (Module or corresponding type declaration not found) error when importing .module.scss file from ts file. He supported the idea of including `src/**/*` into TSC processing - he doesn't see a reason why it shouldn't be done and told me that it's included in each tsconfig in other projects
Besides that I've fixed eslint errors about using node global variables in js files

src/helpers/typesciptCssModulesHelper.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * Return style by key - solves to fix noImplicitAny errors when referencing modular styles from ts files via index accessor
+ * @param styles modular styles object
+ * @param key string classname
+ */
+export function getStyle<StylesType>(styles: StylesType, key: string): StylesType[keyof StylesType] | undefined {
```
sahariko, line 6: TS is really good at inferring types:

```suggestion
export function getStyle<StylesType>(styles: StylesType, key: string) {
```

SergeyRoyt: I know, I've just left it for clarity) But let's remove it if u think that would be better

tsconfig.json
```diff
@@ -1,12 +1,12 @@
...
-  "include": ["./dist/ts-esm.js", "types/styles.d.ts"],
+  "include": ["src/**/*", "types/**/*"],
   "compilerOptions": {
     "rootDir": "./src",
     "lib": ["esnext", "dom"],
     "allowJs": true,
     "target": "es6",
     "module": "commonjs",
-    "sourceMap": false,
+    "sourceMap": true,
```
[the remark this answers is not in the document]
SergeyRoyt: I've seen some deprecation warning about this, but can't reproduce rn... :0 
Reverting...

orrgottlieb (approved):

PR #883, opened 2022-09-21
fix externals issue

orrgottlieb:


laviomri (approved):
LGTM

webpack.config.js
```diff
@@ -4,12 +4,15 @@ const CopyPlugin = require("copy-webpack-plugin");
 const TerserJSPlugin = require("terser-webpack-plugin");
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 const { getPublishedComponents } = require("./webpack/published-components");
+const BundleAnalyzerPlugin = require("webpack-bundle-analyzer").BundleAnalyzerPlugin;
+
+const ANALYZE_BUNDLE = false;
```
niksa-monday, line 9: [Optional]

Probably we can add additional NPM script at package json for not change false-true back and force.
Like at [monolith](https://github.com/DaPulse/dapulse/blob/master/package.json#L352), see also [here](https://github.com/DaPulse/dapulse/blob/master/webpack.config.js#L142)

niksa-monday (approved):

webpack.config.js
orrgottlieb: @niksa-monday will do in additional PR

PR #897, opened 2022-10-02
convert helpers/utils to typescript

orrgottlieb:
Convert the utils/helpers libraries to ts

i left the scripts which are running in node in JS - will move them to a different directory in a different PR


src/utils/colors-utils.ts
```diff
@@ -3,7 +3,7 @@ import { COLOR_STYLES, contentColors } from "./colors-vars-map";
 const ColorUtils = {
   modes: COLOR_STYLES,
   contentColors,
-  getMondayColorAsStyle: (color, mode = COLOR_STYLES.REGULAR, withVar = true) => {
+  getMondayColorAsStyle: (color: string, mode = COLOR_STYLES.REGULAR, withVar = true) => {
```
laviomri, line 6: is any color acceptable here? 
if not, is there a quick way to type it? (should it be the property values of `contentColorsByName`?)

src/utils/dom-event-utils.ts
```diff
@@ -12,67 +12,71 @@ const ARROW_UP_KEY_CODE = 38;
...
+  keyCode: KeyboardEvent["keyCode"] | KeyboardEvent["which"],
+  key?: KeyboardEvent["key"]
+) {
   if (!event) return false;
   return event.keyCode === keyCode || event.which === keyCode || (key && event.key === key);
-};
+}
 
-export const isEnterEvent = event => {
+export function isEnterEvent(event: KeyboardEvent) {
```
laviomri, line 24: not that it matters to me that much, but why did you change from `const` to functions?

laviomri (approved):

src/utils/dom-event-utils.ts
orrgottlieb: @laviomri i'm not a huge favor of declaring `consts` when functions are needed - less evaluation, but i think it is clearer with TS (IMO)

laviomri: @orrgottlieb I'm not opinionated to consts or functions, but if we prefer having one over the other I think we should have a linting rule for that. Otherwise we'll have some PRs moving consts to functions and other PRs moving them back. 
what do you think?

PR #919, opened 2022-10-04
Bugfix: Disabled button no longer showing a not-allowed cursor #774

### LINKED ISSUE #774: Disabled button no longer showing a not-allowed cursor
Due to `pointer-events: none;` `cursor: not-allowed` is no longer working

vczb:
- Remove pointer events none
  - Add disabled attribute
  - Update snapshot

---

## Bug reason: 

the bug is happen because is not possible use cursor styles and pointer-events:none together.

## Implementation:

I have been removed the `pointer-events` and added a `disable` html attribute to the button ( that improve the semantic HTML and avoid the click event ).

Also, I have been addeded `monday-style-button--prevent-click-animation` className when is disabled to prevent the click animation

After the changes I update the snapshots 

---

#### Basic
- [x] PR has description.
#### Style
- [x] Styles are added to `NewComponent.modules.scss` file inside of the `NewComponent` folder.
#### Storybook
#### Tests

----

Close #774

### DIFF
--- a/src/components/AlertBanner/__tests__/__snapshots__/AlertBanner-snapshots.jest.js.snap
+++ b/src/components/AlertBanner/__tests__/__snapshots__/AlertBanner-snapshots.jest.js.snap
@@ -49,11 +49,16 @@ exports[`AlertBanner should render correctly with props 1`] = `
       aria-label="close-toast"
       className="monday-alert-banner__alert-banner-close-btn monday-style-button monday-style-button--size-small monday-style-button--kind-tertiary monday-style-button--color-on-primary-color"
       data-testid="button"
+      disabled={false}
       onBlur={[Function]}
       onClick={[Function]}
       onFocus={[Function]}
       onMouseDown={[Function]}
+      onMouseEnter={[Function]}
+      onMouseLeave={[Function]}
       onMouseUp={[Function]}
+      onPointerEnter={[Function]}
+      onPointerLeave={[Function]}
       type="button"
     >
       <svg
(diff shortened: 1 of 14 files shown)


orrgottlieb (changes requested):
hello @vczb thank you so much for the detailed PR and Description
there is an issue with onMouseLeave while a button is disabled -
you can see the following [thread](https://codesandbox.io/s/gracious-galileo-qymqw?file=/src/App.js)

They have solved it with changing the `onMouseLeave` and `onMouseEnter` to `onPointerLeave` and `onPointerEnter`


The reason we added the `pointer-events: none` was to solve an issue we had with our dialog component who listens to event on the "origin" element

So, your solution will work but we will have a bug in our dialog component

vczb:
> onPointerLeave

Hi @orrgottlieb thanks for the clarification. What you think about override the `onPointerEnter` and `onPointerLeave` and check if is disabled? 

Somethink like [this example](https://codesandbox.io/s/button-disable-events-zt2noh?file=/src/App.js)

orrgottlieb:
@vczb i think you can change the `onMouseEnter` to `onPointerEnter` - we would need to check it, but i think it will do the work

vczb:
> we would need to check it, but i think it

I already update the PR, I believe that will work as expected, please test it and let me know if any change is needed

vczb:
> @vczb i think you can change the `onMouseEnter` to `onPointerEnter` - we would need to check it, but i think it will do the work

hi @orrgottlieb let me know if you need more changes if not I will only fix the conflicts

shlomitc:
This PR has been stale for a while so I'm closing it. Thanks

PR #928, opened 2022-10-11
modal css fixes

eran-cohen:
https://monday.monday.com/boards/245345663/views/62248892/pulses/3356926427

### DIFF
--- a/src/components/Modal/Modal.module.scss
+++ b/src/components/Modal/Modal.module.scss
@@ -5,7 +5,7 @@
 }
 
 .container {
-  z-index: 2;
+  z-index: 10000;
   display: flex;
   justify-content: center;
   align-items: center;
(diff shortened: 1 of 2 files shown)


src/components/Modal/Modal.module.scss
```diff
@@ -5,7 +5,7 @@
 }
 
 .container {
-  z-index: 2;
+  z-index: 10000;
```
orrgottlieb, line 8: @eran-cohen why not have a css variable that user can override ?

orrgottlieb (approved):

src/components/Modal/Modal.module.scss
eran-cohen: @orrgottlieb we will, I'll add a task for it, just want to release this asap because it's blocking someone from using the modal

orrgottlieb: sure

PR #932, opened 2022-10-12
Add docs about change publish component files for every ts migration update

hadasfa:
...

### DIFF
--- a/TYPESCRIPT_MIGRATION.md
+++ b/TYPESCRIPT_MIGRATION.md
@@ -22,3 +22,8 @@ While starting migrating components on your own, please pay attention to the fol
 1. After finish to migrate your component and related file, please also edit the component mdx story file the following way:
 2. Remove the prop "of" from the story ArgsTable and replace it with the "story" prop with the value "Overview" (To update the props table to be displaying the current values for enums. 
 3. Check the call to the function createStoryMetaSettings at the start of the mdx file. If there are props enums that are missing from there, please add them.
+
+### Publish files
+1. Please go over the published-js-components file and move all the paths of the components you converted to the published-ts-components file.
+   For example, if you converted the Avatar component from javascript to typescript, then the published-ts-components file should contain the following line: `Avatar: "components/Avatar/Avatar"`.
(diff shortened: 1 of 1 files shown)


TYPESCRIPT_MIGRATION.md
```diff
@@ -22,3 +22,8 @@ While starting migrating components on your own, please pay attention to the fol
 1. After finish to migrate your component and related file, please also edit the component mdx story file the following way:
 2. Remove the prop "of" from the story ArgsTable and replace it with the "story" prop with the value "Overview" (To update the props table to be displaying the current values for enums. 
 3. Check the call to the function createStoryMetaSettings at the start of the mdx file. If there are props enums that are missing from there, please add them.
+
+### Publish files
+1. Please go over the published-js-components file and move all the paths of the components you converted to the published-ts-components file.
+   For example, if you converted the Avatar component from javascript to typescript, then the published-ts-components file should contain the following line: `Avatar: "components/Avatar/Avatar"`.
```
laviomri, line 28: nitpicks - 
1. links would be nice :D 
2. using a little less words
```suggestion
1. Please go over the [published-js-components.js](./webpack/published-js-components.js) file and move all the paths of the converted components to the [published-ts-components.js](./webpack/published-ts-components.js) file.
   For example, when converting the `Avatar` component from JS to TS, then [published-ts-components.js](./webpack/published-ts-components.js) should contain the following line: `Avatar: "components/Avatar/Avatar"`.
```
```

File: TYPESCRIPT_MIGRATION.md [hunk 2]
```diff
@@ -22,3 +22,8 @@ While starting migrating components on your own, please pay attention to the fol
...
+2. Make sure that the path of the migrated component does not contain an extension. For example: your component path should looks like this: `components/Avatar/Avatar` and not like this: `components/Avatar/Avatar.tsx`.
```
laviomri, line 29: is there a way to enforce it? what would happen if the file extension is (wrongly) specified?

laviomri (approved):

TYPESCRIPT_MIGRATION.md
hadasfa: it's our standards regarding imports. I think there is no way to enforced it in this case (md).

PR #937, opened 2022-10-16
Ts/hadas/tooltip

hadasfa:


src/constants/sizes.ts
```diff
@@ -4,9 +4,11 @@ export const BASE_SIZES = {
   LARGE: "large"
 } as const;
 
+export const BASE_SIZES_WITH_NONE = { NONE: "none", ...BASE_SIZES };
+
 export const SIZES = { XXS: "xxs", XS: "xs", ...BASE_SIZES } as const;
 
-export const DialogPositions = {
+export const DialogPosition = {
```
hadasfa, line 11: was not sure about it because the standard we have for enums was singular naming but we did not convert it while converting other components so I did no convert it to enum...
you think it will be better to convert it to regular enum?

src/components/Tooltip/Tooltip.tsx
```diff
@@ -1,25 +1,117 @@
...
+import { BASE_SIZES_WITH_NONE, DialogPosition } from "../../constants";
+import VibeComponentProps from "../../types/VibeComponentProps";
+import { TooltipArrowPosition, TooltipTheme } from "./TooltipConstants";
+import { HideShowEvent, JustifyType, AnimationType } from "../../constants/dialog";
 import "./Tooltip.scss";
 
+export interface TooltipProps extends VibeComponentProps {
+  style?: CSSProperties;
+  children: ReactElement;
+  content: string | ReactElement;
```
orrgottlieb, line 16: @hadasfa i think that string is a ReactElement so no need to mention both

orrgottlieb (changes requested):

src/components/Tooltip/Tooltip.tsx
hadasfa: tried to removed it but its throw errors:
צילום מסך 2022-10-19 ב-13 41 00

orrgottlieb: so lets add a type for it, so we can reuse it

orrgottlieb (approved):

PR #939, opened 2022-10-18
Fix attention box migration to ts

hadasfa:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests

### DIFF
--- a/src/components/AttentionBox/AttentionBox.tsx
+++ b/src/components/AttentionBox/AttentionBox.tsx
@@ -37,15 +37,15 @@ const AttentionBox: React.FC<AttentionBoxProps> & {
   // Backward compatibility for props naming
   componentClassName,
   // Will remove when releasing version 2 as BREAKING CHANGES
-  withIconWithoutHeader,
-  type,
-  icon,
-  iconType,
+  withIconWithoutHeader = false,
+  type = AttentionBox.types.PRIMARY,
+  icon = AlertIcon,
+  iconType = Icon.type.SVG,
   title,
   text,
-  withoutIcon,
+  withoutIcon = false,
   onClose,
-  compact
+  compact = false
 }) => {
   const iconLabel = useMemo(() => {
     if (type === AttentionBoxType.DANGER) {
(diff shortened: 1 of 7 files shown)


src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -18,7 +18,7 @@ interface AttentionBoxProps extends VibeComponentProps {
   // Will remove when releasing version 2 as BREAKING CHANGES
   withIconWithoutHeader?: boolean;
   /** we support 4 types of attention boxes */
-  type?: AttentionBoxTypes;
+  type?: AttentionBoxType;
   /** Icon classname for icon font or SVG Icon Component for SVG Type */
   icon?: string | React.FunctionComponent | null;
```
SergeyRoyt, line 23: I'm making icons everywhere with this type `string | React.FC<iconSubComponentProps> | null`, which I've seen somewhere. Wouldn't it be better to use it here also?

src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -138,13 +138,13 @@ const AttentionBox: React.FC<AttentionBoxProps> & {
 };
 
 Object.assign(AttentionBox, {
-  types: AttentionBoxTypes,
+  types: AttentionBoxType,
   iconTypes: IconType
 });
 
 AttentionBox.defaultProps = {
```
SergeyRoyt, line 145: We are removing defaultProps, aren't we?

src/components/AttentionBox/__stories__/attentionBox.stories.mdx
```diff
@@ -85,7 +85,7 @@ There are four types of attention boxes: Primary, success, danger, and dark (nat
         <AttentionBox
           title="You're doing great"
           text="Lorem Ipsum is simply dummy text of the printing and typesetting industry."
-          type={AttentionBoxTypes.SUCCESS}
+          type={AttentionBoxType.SUCCESS}
```
SergeyRoyt, line 88: I guess it's better to reference `AttentionBox.types` here, no?

src/components/AttentionBox/AttentionBox.tsx
hadasfa: You are right, I will change it

PR #941, opened 2022-10-18
AlertBanner: Typescript migration

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3388459542


src/components/AlertBanner/AlertBanner.tsx
```diff
@@ -1,24 +1,52 @@
...
+import { AlertBannerTextProps } from "./AlertBannerText/AlertBannerText";
 import "./AlertBanner.scss";
 
-const NOOP = () => {};
+interface AlertBannerProps extends VibeComponentProps {
+  /**
+   * Set external styling to the progress bar.
+   */
+  className?: string;
+  backgroundColor?: ALERT_BANNER_BACKGROUND_COLORS;
```
orrgottlieb, line 19: ALERT_BANNER_BACKGROUND_COLORS should be enum - lets change the name of it to the enum to camel case structure

src/components/AlertBanner/AlertBanner.tsx
```diff
@@ -1,24 +1,52 @@
...
+  isCloseHidden?: boolean;
+  /** ARIA description for the progress bar */
+  ariaLabel?: string;
+  onClose?: () => void;
```
orrgottlieb, line 23: this is being passed to Button, onClick, so it should have the same signature

orrgottlieb (changes requested):

src/components/AlertBanner/AlertBanner.tsx
SergeyRoyt: I've added `(event: React.MouseEvent<HTMLButtonElement>) => void` so it will match the Button onClick signature

PR #947, opened 2022-10-18
SplitButton: Typescript migration

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3390016536


src/components/SplitButton/SplitButton.tsx
```diff
@@ -1,69 +1,94 @@
...
+    className,
+    leftIcon,
+    rightIcon,
+    onClick,
+    children,
+    marginLeft,
+    marginRight
+  } = overrideButtonProps;
+
+  console.log("Button.defaultProps", Button.defaultProps);
```
hadasfa, line 88: let's remove console.logs ?

src/components/SplitButton/SplitButton.tsx
```diff
@@ -1,69 +1,94 @@
...
+  zIndex = null,
   secondaryDialogClassName,
-  secondaryDialogPosition,
-  dialogPaddingSize,
+  secondaryDialogPosition = SplitButtonSecondaryContentPosition.BOTTOM_START,
+  // @ts-ignore TODO TS-migration, when DialogContentContainer is converted to TS
+  dialogPaddingSize = DialogContentContainer.sizes.MEDIUM,
   ...buttonProps
 }) => {
+  const overrideButtonProps = { ...Button.defaultProps, ...buttonProps };
```
hadasfa, line 72: Why we need Button.defaultProps? when we pass it to button, the button default props will override undefined values?

hadasfa (changes requested):
Few small notes

src/components/SplitButton/SplitButton.tsx
SergeyRoyt: Yes, but these button props are used to calculate props and classes for SplitButton as well - not only for Button. That's why we have to know their values before passing them to Button

orrgottlieb (approved):

PR #950, opened 2022-10-20
fix button group after migration

hadasfa:
Fixes after external contribution which related to the typescript migration efforts


src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -6,39 +6,45 @@ import useMergeRefs from "../../hooks/useMergeRefs";
...
 import { IconType } from "../Icon/IconConstants";
+import "./ButtonGroup.scss";
 
 type ButtonGroupOption = {
   icon?: IconType;
   leftIcon?: IconType;
   ariaLabel?: string;
   subText?: string;
-  value: number;
+  value: string | number;
```
laviomri, line 20: nitpick - consider extracting to a type, or use an existing one (since I see you're using this in a few places)

src/components/ButtonGroup/ButtonWrapper.tsx
```diff
@@ -22,7 +25,7 @@ export const ButtonWrapper = ({
   tooltipContainerSelector,
   tooltipMoveBy,
   ...otherProps
-} : ButtonWrapperProps & ButtonProps) => {
+}: ButtonWrapperProps & ButtonProps) => {
```
laviomri, line 28: can you make `ButtonWrapperProps` be the exact type you need, instead of using `&` here?

laviomri (approved):

PR #959, opened 2022-10-24
TS-migration: Steps

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3417392432


src/components/Steps/Steps.tsx
```diff
@@ -1,30 +1,50 @@
...
+  steps?: ReactElement[];
+  type?: StepsType;
+  isOnPrimary?: boolean;
+  isContentOnTop?: boolean;
+  areButtonsIconsHidden?: boolean;
+  backButtonProps?: ButtonProps;
+  nextButtonProps?: ButtonProps;
+}
+
+const Steps: FC<StepsProps> & { types?: typeof StepsType } = forwardRef(
```
hadasfa, line 33: As far as I remember this is not the right type for a component written using using forwardRef, and it should not compile. Is it compile? if not I think using VibeComponent type will fix it. We usually use it for this case.

hadasfa (approved):
It's great. Let's change the main component to be vibe component type for consistency if it's not break something

src/components/Steps/Steps.tsx
SergeyRoyt: Fixed this one

PR #960, opened 2022-10-25
Modal: TS-migration

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3422179360


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,163 @@
...
+  description = "",
+  onClose = NOOP,
+  alertDialog = false,
+  children,
+  triggerElement,
+  width = ModalWidth.DEFAULT,
+  hideCloseButton = false,
+  closeButtonAriaLabel = "close"
+}) => {
+  const childrenArray: ReactElement[] = useMemo(() => React.Children.toArray(children) as ReactElement[], [children]);
```
orrgottlieb, line 84: what happens if `children === null` ?

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,163 @@
...
+      >
+        {header}
+        {content}
+        {footer}
+      </div>
+    </div>,
+    document.body
+  );
+
+  return <>{dialog}</>;
```
orrgottlieb, line 156: why do we need `<></>` can we change it to ```  return ReactDOM.createPortal(...```

src/components/Modal/useShowHideModal.ts
```diff
@@ -1,8 +1,22 @@
...
 
-export default function useShowHideModal({ instance, show, triggerElement, onClose, alertDialog }) {
+export default function useShowHideModal({
+  instance,
+  show,
+  triggerElement,
+  onClose,
+  alertDialog
+}: {
+  // the type is A11yDialog, but it's not exported, so can't be used
```
orrgottlieb, line 12: lets create a `A11yDialogType` - have it set to any and once we have something better we could replace it in one place

src/components/ModalFooter/ModalFooter.tsx
```diff
@@ -0,0 +1,18 @@
+import React, { FC, ReactElement } from "react";
+import cx from "classnames";
+import VibeComponentProps from "../../types/VibeComponentProps";
+import classes from "./ModalFooter.module.scss";
+
+interface ModalFooterProps extends VibeComponentProps {
+  children: ReactElement | ReactElement[] | string;
```
orrgottlieb, line 7: @SergeyRoyt Please add task to create a unified children type so would could use it everywhere?

src/components/ModalHeader/ModalHeader.tsx
```diff
@@ -1,23 +1,75 @@
...
   title,
   titleClassName,
-  description,
+  description = "",
   descriptionClassName,
   icon,
-  closeModal,
-  iconSize,
+  closeModal = NOOP,
+  iconSize = 24,
```
orrgottlieb, line 69: Magic number 🪄

orrgottlieb (approved):

src/components/ModalFooter/ModalFooter.tsx
SergeyRoyt: Yes, was thinking about that too. Will do

src/components/Modal/Modal.tsx
SergeyRoyt: Fixed

src/components/Modal/useShowHideModal.ts
SergeyRoyt: Done

src/components/ModalHeader/ModalHeader.tsx
SergeyRoyt: 🪄 🧙

PR #974, opened 2022-11-08
ComboboxItems: add categories to calculations.

m-binygal:
When a combobox has categories they are not calculated in the max options without scroll calculations. In some rare cases (one category with one option for example) the component is not usable at all.
This commit adds the categories number to the calculation.

#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


orrgottlieb:
@m-binygal thank you!

PR #977, opened 2022-11-09
POC for exporting storybook configuration (will move to a different repo in the near future)

hadasfa:
This is a POC for exporting storybook configurations for other storybooks. In this stage we tried to export only the interactions test. 
The reason we compile our code here by using babel instead of our regular mechanism with webpack is that:
1. We had issues using webpack for compiling the preset (the new storybook addon).
2. We plan to move this code really soon to a separate package which contains everything needed for config vibe-alike storybook.
3. The new package purpose will be to be used internally in different storybook and because of it we less care about it's bundle size if our new code will not increase the bundle size of vibe (I think it should be ok because it should not effect our main entries, do you think there is a reason to worry about?)

Is there a way to tag this as experimental? Maybe I will want to change apis in the future and I do not want to cause breaking changes)


.storybook/.babelrc
```diff
@@ -0,0 +1,7 @@
+{
+  "presets": [
```
hadasfa, line 2: for support ts in .storybook folder (for storybook configurations)

.storybook/addons.js
```diff
@@ -1,7 +1,3 @@
-import "@storybook/addon-docs/register";
```
hadasfa, line 1: addons already imported by main.ts, no need in double import

src/components/Accordion/Accordion/__tests__/accordion.interactions.ts
```diff
@@ -1,8 +1,8 @@
 import { expect } from "@storybook/jest";
-import { userEvent, within } from "@storybook/testing-library";
+import { within } from "@storybook/testing-library";
 import { Screen } from "@testing-library/react";
-import { delay, interactionSuite, resetFocus } from "../../../../__tests__/interactions-helper";
-
+import { resetFocus } from "../../../../__tests__/interactions-helper";
+import { userEvent, delay, interactionSuite } from "../../../../../dist/storybook-utils/services/interactions.js";
```
hadasfa, line 5: for using compiled version and show it's working. If you think it will be better I can try import it form source (even that the code there should not be compiled from webpack like other code)

src/storybook-utils/.babelrc
```diff
@@ -0,0 +1,8 @@
+{
```
hadasfa, line 1: only the storybook utils with the preset are compiled with babel (Because of trying to compile it with webpack create weird issues and we decide not to put effort on this right now because in the and we will move the code here outside)

src/storybook-utils/services/interactions.ts
```diff
@@ -0,0 +1,223 @@
+import { within, userEvent, fireEvent } from "@storybook/testing-library";
```
hadasfa, line 1: I had issues with using the right types here(which react testing library except me to use), I did not understand what should be the type for QueriedHTMLElement. I was not sure if just keep it as js or if it is ok to export it right this because we do want all the code in storybook utils will be ts.
I thought maybe it's fine for POC stage and I will solve it in a different task? wdyt?

PR #981, opened 2022-11-10
TS-migration: VirtualizedList

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3505931266


src/components/VirtualizedList/VirtualizedList.tsx
```diff
@@ -1,43 +1,156 @@
...
 } from "../../services/virtualized-service";
-import "./VirtualizedList.scss";
 import { ELEMENT_TYPES, getTestId } from "../../utils/test-utils";
+import VibeComponentProps from "src/types/VibeComponentProps";
+import VibeComponent from "../../types/VibeComponent";
+import "./VirtualizedList.scss";
 
-const VirtualizedList = forwardRef(
+type ItemType = {
+  value: string;
```
hadasfa, line 32: we are sure everything is not optional here?

hadasfa (approved):
Please check regarding comment if something can be optional in the item type before merge.

src/components/VirtualizedList/VirtualizedList.tsx
SergeyRoyt: Everything is optional now

PR #984, opened 2022-11-13
TS-migration: EditableInput

SergeyRoyt:
https://monday.monday.com/boards/2861766393/pulses/3516854188


src/components/EditableInput/EditableInput.tsx
```diff
@@ -101,6 +128,7 @@ const EditableInput = forwardRef(
         }
 
         const enrichedEvent = event;
+        // @ts-ignore
```
hadasfa, line 131: let's document the reason? if it's something that can be solved easily let's solve it.

src/components/EditableInput/EditableInput.tsx
```diff
@@ -164,7 +192,7 @@ const EditableInput = forwardRef(
           onArrowKeyDown &&
           (isArrowUpEvent(event) || isArrowDownEvent(event) || isArrowLeftEvent(event) || isArrowRightEvent(event))
         ) {
-          onArrowKeyDown(valueState, event);
+          onArrowKeyDown(event, valueState);
```
hadasfa, line 195: changing the order of function prop parameters is a breaking change... I think it will be better not to do it

hadasfa (approved):
Please see comments

PR #986, opened 2022-11-15
TS-Migration: Reduced amount of ts-ignore

khitrind:
Reduced count of ts-ignore in project. Added several TS hekpers `Split` and `LastArrayElement`


hadasfa (approved):
Super cool! Thank you so much for your contribution

PR #993, opened 2022-11-20
support tab index for text field

etaylib:
want to skip tab on textField
so passing a param to -1 its tab index


SergeyRoyt (approved):
Just provide default value for `tabIndex = 0` and that's it

PR #996, opened 2022-11-21
RadioButton: add disabled-text-color

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3553912892

### DIFF
--- a/src/components/RadioButton/RadioButton.scss
+++ b/src/components/RadioButton/RadioButton.scss
@@ -60,6 +60,7 @@
   &__radio-label {
     line-height: 21px;
     font-weight: 400;
+    font-size: var(--font-size-general-label);
     @include theme-prop(color, primary-text-color);
   }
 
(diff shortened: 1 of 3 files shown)


src/components/RadioButton/RadioButton.scss
```diff
@@ -60,11 +60,15 @@
...
     line-height: 21px;
     font-weight: 400;
+    font-size: var(--font-size-general-label);
     @include theme-prop(color, primary-text-color);
   }
 
   &.disabled {
     cursor: not-allowed;
+    .monday-style-radio-button-component__radio-label {
+      @include theme-prop(color, disabled-text-color);
```
hadasfa, line 70: we don't use theme-prop for new css anymore, you can just use the css variable like in other places.

hadasfa:
did you ask talya about screen reader aria attribute in the end?

PR #999, opened 2022-11-22
Support all edge cases for multi select drop with counter (hidden chips in dropdown with defined width edge case)

hadasfa:
We had an issue with some edge cases in multi drop down with single line states (small dropdown looks empty). The reason was that the drop down input placed in a separate line from the chips which was not hidden in the drop down (and therefore not include as part of the counter chips).

The fix is to add a flex wrapper to the last chip which is not hidden which will contain also the input and this way we are making sure that there is no case when the last visible chip and the input are not in the same line.


src/components/Dropdown/__stories__/dropdown.stories.mdx
```diff
@@ -241,7 +241,6 @@ The Dropdown component supports multiple options selection in two different stat
               defaultValue={[optionsAvatar[0]]}
               options={optionsAvatar}
               multi
-              multiline
```
hadasfa, line 244: will revert after design review

src/components/Dropdown/__stories__/dropdown.stories.scss
```diff
@@ -18,7 +18,7 @@
   }
 
   &_with-chips {
-    min-width: 300px;
```
hadasfa, line 21: will revert after design review

src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
```diff
@@ -1,6 +1,10 @@
 @import "../../../../styles/themes";
 @import "../../../../styles/typography";
 
+.chip-with-input-wrapper {
+  display: flex;
+  min-height: 30px;
```
SergeyRoyt, line 6: why we need `min-height` here?

SergeyRoyt (approved):

src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
hadasfa: otherwise when the chip is too small the avatar become an ellipses instead of a circle

PR #1004, opened 2022-11-24
Interactions tests exports

hadasfa:


src/__tests__/interactions-helper.js
```diff
@@ -1,6 +1,7 @@
 import { within, userEvent, fireEvent } from "@storybook/testing-library";
 import { waitFor } from "@testing-library/react";
-import { getTestId, ELEMENT_TYPES as types, NAVIGATIONS_COMMANDS } from "../utils/test-utils";
+import { getTestId, ELEMENT_TYPES as types } from "../utils/test-utils";
+import { NavigationCommand as NAVIGATIONS_COMMANDS } from "../interactions-tests/constants";
```
hadasfa, line 4: for backward compatibility (trying not to create a huge pr (will removed seperatly)

src/components/Accordion/Accordion/__tests__/accordion.interactions.ts
```diff
@@ -1,7 +1,8 @@
 import { expect } from "@storybook/jest";
 import { userEvent, within } from "@storybook/testing-library";
 import { Screen } from "@testing-library/react";
-import { delay, interactionSuite, resetFocus } from "../../../../__tests__/interactions-helper";
+import { resetFocus } from "../../../../__tests__/interactions-helper";
```
hadasfa, line 4: reset focus is something which is super specific to vibe (it using a storybook decorator).
At first I thought not to put it as part of anything we exporting but maybe I will try to find another solution for reset focus without a storybook decorator ? because it is something which will probably be needed.

hadasfa: wdyt?

src/components/Chips/__tests__/chips-tests.jest.tsx
```diff
@@ -15,7 +15,7 @@ describe("Chips tests", () => {
   });
 
   it("should call the click callback when clicked", () => {
-    const testId = getTestId(`${ELEMENT_TYPES.CHIP}-close`);
+    const testId = getTestId(ELEMENT_TYPES.CHIP, `close`);
```
hadasfa, line 18: it returns the same value (but without ts errors)

src/interactions-tests/constants.ts
```diff
@@ -0,0 +1,66 @@
+export enum ComponentDefaultTestId {
+  // Don't remove next line
+  // plop_marker:default-data-testid-declarations
+  COMBOBOX = "combobox",
```
orrgottlieb, line 4: this looks like something that could break, don't you think that this should be exported on the component level?

src/interactions-tests/utils.ts
```diff
@@ -0,0 +1,213 @@
...
+      throw result;
+    }
+    return result;
+  } else if (canvasOrValidTestElement.getByRole) return canvasOrValidTestElement;
+}
+
+// External functions
+export const interactionSuite =
+  ({
+    beforeEach = null,
```
orrgottlieb, line 71: as this something that is being exported, i would add `beforeAll` and `afterAll`

and as a general approach i would try to keep it as close to jest standard syntax

PR #1013, opened 2022-11-27
Feature/sergeyro/ts migration menu

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3571941807


src/components/Menu/Menu/hooks/useOnCloseMenu.tsx
```diff
@@ -0,0 +1,21 @@
+import { useCallback } from "react";
+import { CloseMenuOption } from "../MenuConstants";
+
+export default function useOnCloseMenu({
```
sahariko, line 4: This is changing this hook's signature, is this hook exported for consumption?

src/components/Menu/Menu/hooks/useLastNavigationDirection.tsx
```diff
@@ -14,21 +14,21 @@ const NAVIGATION_KEYS = [...ARROW_UP_KEYS, ...ARROW_RIGHT_KEYS, ...ARROW_DOWN_KE
...
 
-  const setLastNavigationDirection = useCallback(dir => {
+  const setLastNavigationDirection = useCallback((dir: NavDirections) => {
     lastNavigationDirectionRef.current = dir;
   }, []);
 
   const onKeyEvent = useCallback(
-    ({ key }) => {
-      if (ARROW_UP_KEYS.includes(key)) {
+    (event: KeyboardEvent) => {
```
sahariko, line 24: Nitpicky, but you can do:
```suggestion
    ({ key }: KeyboardEvent) => {
```

src/components/Menu/Menu/hooks/useCloseMenuOnKeyEvent.tsx
```diff
@@ -1,29 +1,37 @@
...
+    (event: React.KeyboardEvent) => {
       const isArrowLeftClick = event.key === "ArrowLeft";
 
       if (isArrowLeftClick && !isSubMenu) {
         return;
       }
 
       if (hasOpenSubMenu) return false;
-      onCloseMenu(event);
+      onCloseMenu(event as CloseMenuOption);
```
sahariko, line 32: This is super confusing IMO, on the one hand we're calling the argument's type "Options", on the other hand it's a DOM event.

Why not use the built-in type for the event instead?

src/components/Menu/Menu/hooks/useAdjacentSelectableMenuIndex.tsx
```diff
@@ -1,16 +1,17 @@
-import { useCallback } from "react";
+import { ReactElement, useCallback } from "react";
 
-export const useAdjacentSelectableMenuIndex = ({ children }) => {
+export const useAdjacentSelectableMenuIndex = ({ children }: { children: ReactElement[] }) => {
   const isChildSelectable = useCallback(
-    newIndex => {
+    (newIndex: number) => {
       const child = children[newIndex];
+      // @ts-ignore
```
sahariko, line 7: Why?

src/components/Menu/Menu/Menu.tsx
```diff
@@ -62,6 +94,7 @@ const Menu = forwardRef(
     const children = useMemo(() => {
       const allChildren = React.Children.toArray(originalChildren);
       return allChildren.filter(child => {
+        // @ts-ignore
```
sahariko, line 97: Why?

src/components/Menu/Menu/hooks/useSubMenuIndex.tsx
```diff
@@ -1,11 +1,11 @@
 import { useCallback, useState } from "react";
 
 export default function useSubMenuIndex() {
-  const [openSubMenuIndex, setOpenSubMenuIndex] = useState(null);
-  const hasOpenSubMenu = openSubMenuIndex || openSubMenuIndex === 0;
+  const [openSubMenuIndex, setOpenSubMenuIndex] = useState<number>(null);
+  const hasOpenSubMenu = !!openSubMenuIndex || openSubMenuIndex === 0;
```
sahariko, line 5: Unrelated to your PR, but this logic is a little bit messy and implicit, making it hard to understand what you're actually checking for here (which I'm assuming, whether that index is a number). How about something like:

```suggestion
  const hasOpenSubMenu = typeof openSubMenuIndex === 'number';
```

src/components/Menu/MenuDivider/MenuDivider.tsx
```diff
@@ -1,9 +1,15 @@
...
+import { FC } from "react";
 import "./MenuDivider.scss";
 
-const MenuDivider = ({
+interface MenuDividerProps extends VibeComponentProps {
+  /** Backward compatibility for props naming **/
+  classname?: string;
+}
+
+const MenuDivider: FC<MenuDividerProps> & { isMenuChild?: boolean } = ({
```
sahariko, line 12: Why not include this in `MenuDividerProps`?

src/components/Menu/MenuGridItem/useMenuGridItemNavContext.tsx
```diff
@@ -10,6 +11,14 @@ export const useMenuGridItemNavContext = ({
...
   isUnderSubMenu,
   closeMenu
+}: {
+  wrapperRef?: (node: HTMLElement) => void;
+  setActiveItemIndex?: (index: number) => void;
+  getNextSelectableIndex?: (activeItemIndex: number) => number;
+  getPreviousSelectableIndex?: (activeItemIndex: number) => number;
+  activeItemIndex?: number;
+  isUnderSubMenu?: boolean;
+  closeMenu?: (option: CloseMenuOption) => void;
```
sahariko, line 21:
```suggestion
  closeMenu?: (options: CloseMenuOption) => void;
```

Also, aren't these optional? If so:

```suggestion
  closeMenu?: (options?: CloseMenuOption) => void;
```

src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -14,46 +22,89 @@ import usePopover from "../../../hooks/usePopover";
...
+  closeMenu?: (option: CloseMenuOption) => void;
+  menuRef?: React.RefObject<HTMLElement>;
+  children?: ReactElement | ReactElement[];
+}
+
+const MenuItem: VibeComponent<MenuItemProps> & {
+  iconType?: typeof Icon.type;
+  tooltipPositions?: typeof DialogPosition;
+  isSelectable?: boolean;
+  isMenuChild?: boolean;
```
sahariko, line 70: Why not include these in `MenuItemProps`?

src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -14,46 +22,89 @@ import usePopover from "../../../hooks/usePopover";
...
+  setActiveItemIndex?: (index: number) => void;
+  index?: number;
+  key?: string;
+  isParentMenuVisible?: boolean;
+  resetOpenSubMenuIndex?: () => void;
+  hasOpenSubMenu?: boolean;
+  setSubMenuIsOpenByIndex?: (index: number, isOpen: boolean) => void;
+  useDocumentEventListeners?: boolean;
+  tooltipContent?: string;
+  tooltipPosition?: typeof MenuItem.tooltipPositions[keyof typeof MenuItem.tooltipPositions];
```
sahariko, line 52: Maybe export this as an enum for better readability/usability?

src/components/Menu/MenuTitle/MenuTitleConstants.ts
```diff
@@ -0,0 +1,5 @@
+export enum MenuTitleCaptionPositions {
```
sahariko, line 1: It's convention to name enums as sungular:

```suggestion
export enum MenuTitleCaptionPosition {
```

src/components/Menu/Menu/Menu.tsx
```diff
@@ -10,41 +18,65 @@ import useOnCloseMenu from "./hooks/useOnCloseMenu";
...
+  onClose?: (option: CloseMenuOption) => void;
+  focusItemIndex?: number;
+  isSubMenu?: boolean;
+  useDocumentEventListeners?: boolean;
+  focusItemIndexOnMount?: number;
+  shouldScrollMenu?: boolean;
+  children?: ReactElement | ReactElement[];
+}
+
+const Menu: VibeComponent<MenuProps> & {
```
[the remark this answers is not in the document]
SergeyRoyt: Cause all of them are static properties. In other places we've done the same, so it's just for consistency

src/components/Menu/Menu/hooks/useOnCloseMenu.tsx
SergeyRoyt: No, it's not exported outside. Used only in Menu

src/components/Menu/Menu/hooks/useMouseLeave.tsx
```diff
@@ -3,7 +3,17 @@ import { useLayoutEffect } from "react";
 import useIsMouseEnter from "../../../../hooks/useIsMouseEnter";
 import usePrevious from "../../../../hooks/usePrevious";
 
-export default function useMouseLeave(resetOpenSubMenuIndex, hasOpenSubMenu, ref, setActiveItemIndex) {
+export default function useMouseLeave({
```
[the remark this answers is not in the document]
SergeyRoyt: No, it's not exported outside. Used only in Menu

src/components/Menu/MenuDivider/MenuDivider.tsx
SergeyRoyt: Cause all of them are static properties. In other places we've done the same, so it's just for consistency

src/components/Menu/MenuItem/MenuItem.tsx
SergeyRoyt: static props

src/components/Menu/Menu/Menu.tsx
SergeyRoyt: This static `type` properties are throwing following error
`TS2339: Property 'type' does not exist on type 'string | number | ReactElement<any, string | JSXElementConstructor<any>> | ReactFragment | ReactPortal'.   Property 'type' does not exist on type 'string'.`
What would be a solution here @sahariko? We have this problem in many places

src/components/Menu/Menu/hooks/useAdjacentSelectableMenuIndex.tsx
SergeyRoyt: Same static `type` problem

src/components/Menu/Menu/hooks/useCloseMenuOnKeyEvent.tsx
SergeyRoyt: Instead of passing event here, I'm passing now `{propagate: false}` as event doesn't have a `propagate` property

PR #1018, opened 2022-11-29
Feature/sergeyro/chips hovered state + clickable

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3591193843
Chips hover (+ update colors-description) + clickability

Need to check with care - potentially dangerous changes


src/utils/colors-vars-map.ts
```diff
@@ -39,12 +40,14 @@ export const colorsMap = [
   },
   { color: "--positive-color-hover", description: "Use only as hover color on positive color" },
   { color: "--positive-color-selected", description: "Use only as selected indication for a positive colors" },
+  { color: "--positive-color-selected-hover", description: "description" }, // TODO
```
SergeyRoyt, line 43: TODO write description

src/components/Chips/Chips.tsx
```diff
@@ -75,12 +76,21 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
-    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef, hoverRef] });
 
     const backgroundColorStyle = useMemo(() => {
-      return { backgroundColor: disabled ? getCSSVar("disabled-background-color") : getElementColor(color, true) };
-    }, [disabled, color]);
+      let cssVar;
+      if (disabled) {
+        cssVar = getCSSVar("disabled-background-color");
+      } else if (isHovered) {
```
hadasfa, line 87: according to the design, hover state should be added only when the chip is clickable. This should happen only if there is a onMouseDown prop or onClick (which is not exist right now and we need to add)

src/components/Chips/Chips.tsx
```diff
@@ -75,12 +76,21 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
+        cssVar = getElementColor(color, true, true);
```
hadasfa, line 88: not all colors having selected hover color. so I guess it will have a bad color for not supported colors. we need to add a check somehow if the color is not supported. I have an idea, let's talk about it in zoom?

src/components/Chips/Chips.tsx
```diff
@@ -104,6 +120,7 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
         id={id}
         style={backgroundColorStyle}
         onMouseDown={onMouseDown}
+        onClick={onClick}
```
hadasfa, line 123: we know how to this accessible so let's not do it in a different pr (it's part of the definition of done)
You can add a clickable conditionally (just be aware that there is a different name to the data test id attribute0

src/components/Chips/Chips.tsx
```diff
@@ -47,6 +46,18 @@ interface ChipsProps extends VibeComponentProps {
...
+   */
+  onClick?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
+  /**
+   * Applies when element has onClick or onMouseDown props
+   */
+  ariaLabel?: string;
+  /**
+   * Should element be focusable & clickable - for backward compatability
+   */
+  isClickable?: boolean;
```
SergeyRoyt, line 60: Added in order not to break anything old, which is using chip - e.g. Dropdown with multi-state

PR #1024, opened 2022-11-30
Import optimizations

orrgottlieb:
optimize imports


src/hooks/usePopover.ts
```diff
@@ -3,7 +3,7 @@ import { usePopper } from "react-popper";
 import { Placement } from "./popoverConstants";
 
 import useForceUpdate from "./useForceUpdate";
-import * as PopperJS from "@popperjs/core";
+import { Options, State } from "@popperjs/core";
```
sahariko, line 6:
```suggestion
import type { Options, State } from "@popperjs/core";
```

Should make sure it doesn't find its way into the bundle 🙂

sahariko (approved):

src/utils/user-agent-utils.ts
```diff
@@ -1,15 +1,5 @@
...
-
-export function getBrowserName() {
-  if (browser && browser.name) {
-    return browser.name;
-  }
-}
-
 export function isFirefox() {
-  return getBrowserName() === "firefox";
+  return !!/Firefox\/([0-9\.]+)(?:\s|$)/.exec(window.navigator.userAgent);
```
hadasfa, line 2: let's just fix lint error?

hadasfa (approved):
let's just fix linter error/ignore it before merging?

PR #1025, opened 2022-11-30
Export interactions 2

hadasfa:
צילום מסך 2022-11-30 ב-13 49 23

 (Main is bigger in 1 kb than in master)


plop/component/component-ts.txt
```diff
@@ -23,12 +24,16 @@ const {{properCase componentName}}: VibeComponent<{{properCase componentName}}Pr
...
+        data-testid={dataTestId || getTestId(ComponentDefaultTestId.{{constantCase componentName}}, id)}
       >
         My awesome new component
       </div>
     );
   }
 );
 
+Object.assign({{properCase componentName}}, {
+   defaultTestId: ComponentDefaultTestId.{{constantCase componentName}}
```
SergeyRoyt, line 36: Where `defaultTestId` is used?

src/tests/constants.ts
```diff
@@ -0,0 +1,69 @@
...
+  TOAST_LINK = "toast-link",
+  TOGGLE = "toggle",
+  DIVIDER = "divider",
+  DATEPICKER = "date-picker",
+  DATEPICKER_HEADER = "date-picker-header",
+  DATEPICKER_YEAR_SELECTION = "date-picker-year-selection",
+  LOADER = "loader"
+}
+
+export const BUTTON_DEFAULT_TEST_ID = ComponentDefaultTestId.BUTTON;
```
SergeyRoyt, line 58: what for it's needed?

src/__tests__/interactions-helper.js
```diff
@@ -1,188 +1,6 @@
...
-
-export function delay(timeout) {
-  return new Promise(resolve => {
-    if (!timeout) return resolve();
-    setTimeout(resolve, timeout);
-  });
-}
+import { userEvent } from "@storybook/testing-library";
 
 export async function resetFocus() {
```
SergeyRoyt, line 3: why it's the only thing which left here, while everything else was moved?

src/tests/base-tests-utils.ts
```diff
@@ -0,0 +1,5 @@
+import { ComponentDefaultTestId as TestIds } from "./constants";
+export const ComponentDefaultTestId = TestIds;
+export const getTestId = (elementType: TestIds, id?: string | number) => {
```
SergeyRoyt, line 3: Why we decided to move to that instead of `ELEMENT_TYPES`? 
I'm afraid that'll make Vibe 2.0 branch support even harder :(

webpack.config.js
```diff
@@ -61,6 +61,8 @@ module.exports = options => {
 
   const entry = {
     main: path.join(__dirname, "/src/index.js"),
+    interactionTests: path.join(__dirname, "/src/tests/interactions-utils"),
+    testIds: path.join(__dirname, "/src/tests/base-tests-utils"),
```
SergeyRoyt, line 65: testIds is not very accurate name e.g. if we want to add to base-tests-utils something else in the future...
let's maybe call it baseTestUtils?
And are we sure we need two separate entries at all here?

SergeyRoyt (approved):

PR #1026, opened 2022-11-30
Add interactions basic readme file

hadasfa:


src/tests/readme.md
```diff
@@ -0,0 +1,158 @@
+#Implementing interactions tests using Storybook & Vibe
+For implementing more complex e2e tests in the storybook environment, we recommend using storybook interaction tests combined with the tools exported by our repository.
+For example, interaction tests can be a more straightforward tool for testing keyboard navigation support flow on your composites or pages, happy path flows, or reproducing error or empty complex scenarios.
+
+##Implement interactions tests for stories by using interactions suites
+Interactions tests are functions that can interact with your story elements.
+You can see an example of connecting a test function to a story below:
+````
+// LoginForm.stories.ts|tsx
```
SergeyRoyt, line 9: Maybe it would be better to give example in mdx format, as we mainly want people to write stories in mdx?

src/tests/readme.md
```diff
@@ -0,0 +1,158 @@
...
+
+const Template: ComponentStory<typeof LoginForm> = (args) => <LoginForm {...args} />;
+export const EmptyForm = Template.bind({});
+
+/*
+* See https://storybook.js.org/docs/react/writing-stories/play-function#working-with-the-canvas
+* to learn more about using the canvasElement to query the DOM
+*/
+export const FilledForm = Template.bind({});
+FilledForm.play = async ({ canvasElement }) => {
```
SergeyRoyt, line 30: Is it possible to write these tests in a separate file? Maybe we should consider it - just not to create a mess

src/tests/readme.md
hadasfa: I know that for mf it will not be the case: because mf components will be most of the time pages/ templates/ composites and the way of use will be different from vibe.
regarding monday ui components I don't know what will be more common ^^"

src/tests/readme.md
hadasfa: It's just an example of how to use the play function and write a test, as a standard I agree.

PR #1032, opened 2022-12-08
Icon: follow create component best practices.

m-binygal:
It is considered unsafe to render a component directly from the class or function if you're getting it from props. The best practice for that is to create the element with React and pass it around.
This fix will should remove many warnings about rendering custom components that are not following the PascalCase naming.

### DIFF
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -109,19 +109,16 @@ const Icon: VibeComponent<IconProps, HTMLElement> & { type?: typeof IconType } =
     const isFunctionType = typeof icon === "function";
     // Replace in major version change with more accurate check
     if (iconType === IconType.SVG || isFunctionType || typeof icon === "object") {
-      const IconComponent = icon;
-      return (
-        <IconComponent
-          id={id}
-          {...screenReaderAccessProps}
-          ref={isFunctionType ? undefined : mergedRef}
-          size={iconSize.toString()}
-          onClick={onClick}
-          className={computedClassName}
-          style={style}
-          data-testid={dataTestId}
-        />
-      );
+      return React.createElement(icon, {
+        id,
+        ...screenReaderAccessProps,
+        ref: isFunctionType ? undefined : mergedRef,
+        size: iconSize.toString(),
+        onClick,
+        className: computedClassName,
+        style,
+        "data-testid": dataTestId
+      });
     }
     if (iconType === IconType.SRC) {
       return (
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -21,7 +21,7 @@ export interface IconSubComponentProps {
   "data-testid"?: string;
 }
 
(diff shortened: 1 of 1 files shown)


m-binygal:
Here is the part of [React docs](https://reactjs.org/warnings/legacy-factories.html) with the approach in this PR.

orrgottlieb:
hi @m-binygal so basicly this is what happens behind the scences when you do something like this
you can build it locally in your enviorment 

the out come is 
```js
e(
          C,
          Object.assign({ id: d }, P, {
            ref: G ? void 0 : z,
            size: "" + g,
            onClick: f,
            className: h,
            style: S,
            "data-testid": O
          })
        )
```

where -
`import { jsx as e } from "react/jsx-runtime";`
and C is the icon compoennt

so basicly it is being transpiled to createClass  
this is the atricle on the new https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

m-binygal:
I was referring to the last piece in the doc. 
Here is the warning I get on our tests (in the picture)
This is the code, and if I'm not mistaken it raises this warning although we use it the exact way that the library wants as to use it. Am I missing something? 

```jsx
import { CloseSmall } from 'monday-ui-react-core/icons';
import PropTypes from 'prop-types';
import NOOP from 'lodash/noop';

const ClearIndicator = ({ size, className, onClick }) => (
  <div className={className} onClick={onClick}>
    <Icon iconType={Icon.type.SVG} icon={CloseSmall} iconSize={size} tabindex="-1" />
  </div>
);
```

image

sahariko:
I'm assuming that what's bothering React is the fact that `icon` is lower-cased, what if we change the prop's name to `Icon`? Or extract that rendering logic to a dedicated function that accepts an argument in pascal-case?

m-binygal:
> I'm assuming that what's bothering React is the fact that `icon` is lower-cased, what if we change the prop's name to `Icon`? Or extract that rendering logic to a dedicated function that accepts an argument in pascal-case?

I guess that will work. The solution that I suggested was just another way (and was in the past the best practice, I don't know if still)
Which approach do you want me to take in my fixes?

sahariko:
Splitting to a dedicated function sounds like the best path here, since it will also hopefully increase the readability of the parent component.

m-binygal:
@sahariko I'm not sure the new implementation is more readable. I've added the new code as a commit. Can you take a look and see which one do you like the best?

src/components/Icon/Icon.tsx
```diff
@@ -21,6 +21,10 @@ export interface IconSubComponentProps {
   "data-testid"?: string;
 }
 
+function renderIcon(Icon: SubIcon, props: IconSubComponentProps): JSX.Element {
```
sahariko, line 24: TS is usually pretty good at inferring types:

```suggestion
function renderIcon(Icon: SubIcon, props: IconSubComponentProps) {
```

m-binygal:
@sahariko round

sahariko (approved):

sahariko:
Looks good to me 🙂

PR #1046, opened 2022-12-27
MenuItem: tooltip display on whole item, not just a title

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3713740094
Previously no Tooltip was displaying at all in such cases
image

### DIFF
--- a/src/components/Menu/MenuItem/MenuItem.tsx
+++ b/src/components/Menu/MenuItem/MenuItem.tsx
@@ -271,64 +271,62 @@ const MenuItem: VibeComponent<MenuItemProps> & {
     }, [disableReason, disabled, title, tooltipContent]);
 
     return (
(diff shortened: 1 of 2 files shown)


src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -290,16 +290,17 @@ const MenuItem: VibeComponent<MenuItemProps> & {
...
-          </div>
+          {renderMenuItemIconIfNeeded()}
+          {title && (
+            <div ref={titleRef} className="monday-style-menu-item__title">
+              {title}
+            </div>
+          )}
         </Tooltip>
         {label && (
           <div ref={titleRef} className="monday-style-menu-item__label">
```
SergeyRoyt, line 306: Is it ok that we might have 2 titleRefs? Is there a difference between `title` and `label` or it's just need for backward compatibility?

PR #1050, opened 2023-01-05
Fix storybook menu colors

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3710647491

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -14,9 +14,11 @@
   #storybook-explorer-menu svg {
     color: #5034ff;
   }
-  #storybook-explorer-menu button[aria-expanded="true"],
+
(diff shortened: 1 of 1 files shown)


.storybook/manager-head.html
```diff
@@ -44,24 +47,30 @@
   //workaround to increase the left panel width just enough to show even the longer story names
```
hadasfa, line 47: just want to make sure - this is only eslint changes?

hadasfa (approved):
if it's only es lint changes

.storybook/manager-head.html
SergeyRoyt: Yes, I changes only content inside <style> tag

PR #1051, opened 2023-01-05
Tipseen TS-migration

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3710545050


src/components/Steps/__stories__/steps.stories.mdx
```diff
@@ -1,7 +1,7 @@
 import Steps from "../Steps";
 import { useMemo } from "react";
 import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
-import Tipseen from "../../Tipseen/Tipseen.jsx";
+import Tipseen from "../../Tipseen/Tipseen.tsx";
```
hadasfa, line 4: let's just remove the tsx suffix

src/components/Tipseen/Tipseen.tsx
```diff
@@ -1,46 +1,83 @@
...
 
-const Tipseen = forwardRef(
+interface TipseenProps extends VibeComponentProps {
+  position?: DialogPosition;
+  animationType?: AnimationType;
+  hideDelay?: number;
+  showDelay?: number;
+  title?: string;
+  isCloseButtonHidden?: boolean;
+  children?: ReactElement;
```
hadasfa, line 29: children is required no?

hadasfa: also let's put the required props before the optional props

src/components/Tipseen/Tipseen.tsx
```diff
@@ -1,46 +1,83 @@
...
 const bemHelper = BEMClass(TIPSEEN_BASE_CSS_CLASS);
 
-const Tipseen = forwardRef(
+interface TipseenProps extends VibeComponentProps {
+  position?: DialogPosition;
+  animationType?: AnimationType;
+  hideDelay?: number;
+  showDelay?: number;
+  title?: string;
+  isCloseButtonHidden?: boolean;
```
hadasfa, line 28: if we already on it let's add prop which called hideCloseButton which will do the same as this and set this as backward compatibility? (we are trying to reduce usage of is prefix on props)

src/components/Tipseen/Tipseen.tsx
```diff
@@ -1,46 +1,83 @@
...
+   * when false, the arrow of the tooltip is hidden
+   */
+  tip?: boolean;
+  /**
+   * PopperJS Modifiers type
+   * https://popper.js.org/docs/v2/modifiers/
+   */
+  modifiers?: Array<Modifier<any>>;
+  closeAriaLabel?: string;
+  isCloseButtonOnImage?: boolean;
```
hadasfa, line 47: same here, if we already on it lets add prop which called closeButtonOnImage / closeOnImage which will do the same as this and set this as backward compatibility? (we are trying to reduce usage of is prefix on props)

src/components/Tipseen/Tipseen.tsx
```diff
@@ -1,46 +1,83 @@
...
+}
+
+const Tipseen: VibeComponent<TipseenProps> & {
+  positions?: typeof DialogPosition;
+  animationTypes?: AnimationType;
+  justifyTypes?: JustifyType;
+} = forwardRef(
   (
     {
       className,
```
hadasfa, line 59: regarding className default value. we had issues in the past when changed empty string to undefined in cases that some conditions in the code checked if the prop is equal to undefined and then changed it cause change of behavior.
Please make sure change here the prop will not casue the same effect or change the default value to be empty string like before :)

src/components/Tipseen/TipseenBasicContent.tsx
```diff
@@ -1,11 +1,18 @@
...
+import { FC } from "react";
+import { ElementContent } from "../../types/ElementContent";
 import styles from "./TipseenBasicContent.module.scss";
 
 const BASE_CSS_CLASS = "monday-style-tipseen-basic-content";
 
-const TipseenBasicContent = ({ title, children, className }) => {
+interface TipseenBasicContentProps extends VibeComponentProps {
+  title?: string;
+  children?: ElementContent | ElementContent[];
```
hadasfa, line 12: ElementContent is not enough?

src/components/Tipseen/TipseenContent.tsx
```diff
@@ -1,64 +1,77 @@
...
 const bemHelper = BEMClass(BASE_CSS_CLASS);
-const EMPTY_OBJECT = {};
+const EMPTY_OBJECT: ButtonPropsBackwardCompatible = {};
 
-const TipseenContent = ({
+interface TipseenContentProps extends VibeComponentProps {
+  title?: string;
+  isDismissHidden?: boolean;
+  children?: ElementContent;
+  isSubmitHidden?: boolean;
```
hadasfa, line 21: if we already change this component let's add prop: "hideSubmit" and make this one as backward compatibility prop (same reason)

src/components/Tipseen/TipseenContent.tsx
```diff
@@ -1,64 +1,77 @@
...
 
 const BASE_CSS_CLASS = "monday-style-tipseen-content";
 const bemHelper = BEMClass(BASE_CSS_CLASS);
-const EMPTY_OBJECT = {};
+const EMPTY_OBJECT: ButtonPropsBackwardCompatible = {};
 
-const TipseenContent = ({
+interface TipseenContentProps extends VibeComponentProps {
+  title?: string;
+  isDismissHidden?: boolean;
```
hadasfa, line 19: if we already change this component let's add prop: "hideDismiss" and make this one as backward compatibility prop (same reason)

src/components/Tipseen/TipseenContent.tsx
```diff
@@ -1,64 +1,77 @@
...
+  children?: ElementContent;
```
hadasfa, line 20: I think it is a required prop, no?

src/components/Tipseen/TipseenImage.tsx
```diff
@@ -0,0 +1,17 @@
+import cx from "classnames";
+import { VibeComponentProps } from "../../types";
+import { FC } from "react";
+import styles from "./TipseenImage.module.scss";
+
+const BASE_CSS_CLASS = "monday-style-tipseen-image";
+
+interface TipseenImageProps extends VibeComponentProps {
+  src?: string | undefined;
```
hadasfa, line 9: Both are required, an image without src or alt is not an image or not accessible

src/components/Tipseen/TipseenTitle.tsx
```diff
@@ -0,0 +1,20 @@
+import React, { FC } from "react";
+import cx from "classnames";
+import { VibeComponentProps } from "../../types";
+import styles from "./TipseenTitle.module.scss";
+
+const BASE_CSS_CLASS = "monday-style-tipseen-title";
+
+interface TipseenTitleProps extends VibeComponentProps {
+  text?: string;
```
hadasfa, line 9: Text is required for a title

src/components/Tipseen/__stories__/tipseen.stories.mdx
```diff
@@ -58,7 +58,7 @@ Tipseen is an virtual unboxing experience that helps users get started with the
 </Canvas>
 
 ## Props
-<ArgsTable of={ Tipseen } />
+<ArgsTable story="Overview" />
```
hadasfa, line 61: yay!

src/constants/positions.ts
```diff
@@ -1,4 +1,4 @@
-export enum BASE_POSITIONS {
+export enum BasePosition {
```
hadasfa, line 1: Did we change every usage of this enum? If we do please set this as resolve :)

src/constants/common.ts
```diff
@@ -0,0 +1,7 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const EMPTY_ARR: any[] = [];
+
+// TODO replace EMPTY_OBJECT declarations
+// export const EMPTY_OBJECT: any = {};
+
+// TODO move NOOP here?
```
hadasfa, line 7: where is noop right now?

src/constants/common.ts
```diff
@@ -0,0 +1,7 @@
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const EMPTY_ARR: any[] = [];
+
+// TODO replace EMPTY_OBJECT declarations
+// export const EMPTY_OBJECT: any = {};
```
hadasfa, line 5: Can you write here some context regarding this note?

src/components/Tipseen/TipseenImage.tsx
hadasfa: I know it was not the case before but it will be more accurate :)

hadasfa (changes requested):
Wrote some notes

src/components/Tipseen/TipseenContent.tsx
SergeyRoyt: Not sure, cause 
1) there is a default children
2) And there is a logic related to that `{children ? <span className={cx(bemHelper({ element: "content" }))}>{children}</span> : null}` ... 
So I don't know

src/components/Tipseen/TipseenBasicContent.tsx
SergeyRoyt: No cause in this case e.g. it's recieving an array of `ReactElement`s
image

src/constants/common.ts
SergeyRoyt: Let's leave this comments and return to it later? Don't want to do a lot of unrelated changes here

src/constants/common.ts
SergeyRoyt: In `utils/function-utils`
Let's leave this comments and return to it later? Don't want to do a lot of unrelated changes here

src/components/Tipseen/Tipseen.tsx
```diff
@@ -1,67 +1,124 @@
...
+  modifiers?: Array<Modifier<any>>;
+  closeAriaLabel?: string;
+  /**
+   * Backward compatability for isCloseButtonOnImage prop
+   */
+  isCloseButtonOnImage?: boolean;
+  closeButtonOnImage?: boolean;
+  onClose?: () => void;
+  // Better be required, but it might be a breaking change
+  content?: ElementContent;
```
SergeyRoyt, line 60: I've written comments like this, so we can do it later

hadasfa (approved):

PR #1055, opened 2023-01-11
Feature/sergeyro/fix clickable chips breaking changes

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3710545243


src/components/Clickable/Clickable.tsx
```diff
@@ -1,33 +1,25 @@
...
   children?: React.ReactNode;
   role?: AriaRole;
-  onClick?: (event: React.MouseEvent) => void;
+  onClick?: (event: React.MouseEvent | React.KeyboardEvent) => void;
   enableTextSelection?: boolean;
   onMouseDown?: (event: React.MouseEvent) => void;
   ariaLabel?: string;
   ariaHidden?: boolean;
-  ariaHasPopup?: boolean | string;
+  ariaHasPopup?: boolean;
```
SergeyRoyt, line 20: breaking change, but...

src/components/Clickable/Clickable.tsx
```diff
@@ -1,33 +1,25 @@
...
   ariaExpanded?: boolean;
-  tabIndex?: string;
+  tabIndex?: string | number;
```
SergeyRoyt, line 22: decided not to do breaking change here, thought tabIndex should be a number - converting it to number afterwards

src/components/Clickable/Clickable.tsx
SergeyRoyt: we can avoid breaking it by doing converting `ariaHasPopup` afterwards, but this is confusing
wdyt?

src/components/Clickable/Clickable.tsx
```diff
@@ -1,33 +1,25 @@
...
+import useClickable from "../../hooks/useClickable";
 
 export interface ClickableProps extends VibeComponentProps {
   elementType?: keyof JSX.IntrinsicElements | string;
   className?: string;
   id?: string;
   children?: React.ReactNode;
   role?: AriaRole;
-  onClick?: (event: React.MouseEvent) => void;
+  onClick?: (event: React.MouseEvent | React.KeyboardEvent) => void;
```
SergeyRoyt, line 15: can be called by keyboard as well - avoiding TS errors

src/components/Chips/Chips.tsx
```diff
@@ -129,86 +129,96 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
+        dataTestId: overrideDataTestId,
+        ariaLabel: ariaLabel || label,
+        ariaHidden: false,
+        ariaHasPopup: false,
+        ariaExpanded: false,
+        style: backgroundColorStyle
+      },
+      mergedRef
+    );
+    const wrapperProps = hasClickableWrapper
```
SergeyRoyt, line 156: doesn't look too good maybe, but at least it's more straightforward which props will wrapper have

src/components/Chips/Chips.tsx
```diff
@@ -129,86 +129,96 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
+        className: overrideClassName,
+        enableTextSelection: allowTextSelection,
+        disabled,
+        id,
+        dataTestId: overrideDataTestId,
+        ariaLabel: ariaLabel || label,
+        ariaHidden: false,
+        ariaHasPopup: false,
+        ariaExpanded: false,
+        style: backgroundColorStyle
```
hadasfa, line 152: not need to be part of hook because hook is logical only and indifferent to style settings

src/components/Chips/Chips.tsx
```diff
@@ -129,86 +129,96 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
+      [styles.noAnimation]: noAnimation,
+      [styles.withUserSelect]: allowTextSelection
+    });
+
+    const clickableProps = useClickableProps(
+      {
+        onClick: onClickCallback,
+        onMouseDown,
+        className: overrideClassName,
+        enableTextSelection: allowTextSelection,
```
hadasfa, line 144: not need to be part of hook because hook is logical only and indifferent to style settings. the purpose of this enableTextSelection prop related to component css

src/components/Chips/Chips.tsx
```diff
@@ -129,86 +129,96 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
+      [styles.withClose]: hasCloseButton,
+      [styles.noAnimation]: noAnimation,
+      [styles.withUserSelect]: allowTextSelection
+    });
+
+    const clickableProps = useClickableProps(
+      {
+        onClick: onClickCallback,
+        onMouseDown,
+        className: overrideClassName,
```
hadasfa, line 143: not need to be part of hook because hook is logical only and indifferent to style settings

src/components/Chips/Chips.tsx
hadasfa: maybe this will be more clear?
```suggestion
    const clickableWrapperProps = hasClickableWrapper
```

SergeyRoyt: Not sure, cause it's more generic than clickableWrapperProps. Can be clickable props, but can be not related to clickable
I think `wrapperProps` are better here

src/components/Chips/Chips.tsx
SergeyRoyt: Ok, let's imagine I'm removing all styles related stuff from the hook. How then it would look here
We still need to get clickable-styles from somewhere, e.g. with classNames it would look this:
```
className: cx(CLICKABLE_CSS_BASE_CLASS, overrideClassName, {
      disabled,
      [bemHelper({ state: "disable-text-selection" })]: !enableTextSelection
    })
```
So on every use of `useClickableProps` I'll need also to import `Clickable.scss` and duplicate this code? For now it would be 2 duplications - in `Clickable` and in `Chips`... Shouldn't work like this

So I can understand that hooks should contain logic and not including stylesheets, but not in this case IMHO - hook is just returning props, which you can spread on element and it would become clickable and styles related stuff are also an important part of this transformation

hadasfa: let's talk over it on zoom ?

src/components/Chips/Chips.tsx
hadasfa: let's talk about it on zoom ? didnt understand really

src/components/Clickable/Clickable.tsx
hadasfa: I am ok with this but according to what Orr say in the last planning about this we need to wait to 2.0, let's add to do or something

src/components/Clickable/Clickable.tsx
hadasfa: let's add a note for remove string on vibe 2.0 ?

src/hooks/useKeyboardButtonPressedFunc.ts
```diff
@@ -1,12 +1,10 @@
...
+import { keyCodes } from "../constants";
 
-const KEYDOWN_SPACE_EVENT = 32;
-const KEYDOWN_ENTER_EVENT = 13;
-
-export function useKeyboardButtonPressedFunc(onClick: (event: KeyboardEvent) => void) {
+export function useKeyboardButtonPressedFunc(onClick: (event: React.KeyboardEvent) => void) {
   return useCallback(
-    (e: KeyboardEvent) => {
-      if (e.keyCode === KEYDOWN_SPACE_EVENT || e.keyCode === KEYDOWN_ENTER_EVENT) {
```
hadasfa, line 9: why changing keyCode to key?

hadasfa: is it necessary?

hadasfa (changes requested):
Let's talk about the things we did not decided yet about on zoom

src/hooks/useKeyboardButtonPressedFunc.ts
SergeyRoyt: just standardising kinda (why not?), not necessary at all

src/components/Chips/Chips.tsx
hadasfa: According to what I know, hooks should be only logical and not include any styles or related. As far as I understand, useClickable is a logical hook that should receive only the relevant props for making it accessible (keyboard navigation aria...)
In other open-source implementations for similar use cases as React Aria, they also ignore styles or class names for the same reason.
I understand that this way, the developer which using the use clickable will need to implement the style by himself (disabled cursor and blocking selection of text), but this is the reason we most of the time will recommend use this clickable component over the hook and recommend using the hook only when there is no other choice.

src/components/Chips/Chips.tsx
```diff
@@ -129,86 +129,96 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
-                hideTooltip
-                icon={CloseSmall}
-                onClick={onDeleteCallback}
-                dataTestId={`${overrideDataTestId}-close`}
-                ref={iconButtonRef}
-              />
-            )}
-          </div>
-        </ClickableWrapper>
+      <div {...wrapperProps}>
```
SergeyRoyt, line 167: only this line were changes, the rest below remains still

src/components/Clickable/Clickable.tsx
SergeyRoyt: note added

src/components/Clickable/Clickable.tsx
SergeyRoyt: note added

src/components/Chips/Chips.tsx
SergeyRoyt: removed from the hook

src/components/Chips/Chips.tsx
SergeyRoyt: removed from the hook

src/components/Chips/Chips.tsx
SergeyRoyt: removed from the hook

src/hooks/index.js
```diff
@@ -10,6 +10,7 @@ export { default as useSetFocus } from "./useSetFocus";
 export { default as useMergeRefs } from "./useMergeRefs";
 export { default as useIsMouseOver } from "./useIsMouseOver";
 export { default as useHover } from "./useHover";
+export { default as useFocus } from "./useFocus";
```
SergeyRoyt, line 13: should be exposed?

src/components/Chips/Chips.tsx
```diff
@@ -98,21 +99,23 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
...
     const componentRef = useRef(null);
-    const mergedRef = useMergeRefs({ refs: [ref, componentRef, hoverRef] });
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef, hoverRef, focusRef] });
 
     const backgroundColorStyle = useMemo(() => {
       let cssVar;
       if (disabled) {
         cssVar = getCSSVar("disabled-background-color");
-      } else if (isHovered && hasClickableWrapper) {
+      } else if (hasClickableWrapper && (isHovered || isFocused)) {
```
SergeyRoyt, line 112: Fixing hovered style on keyboard focus

src/hooks/useFocus.ts
```diff
@@ -0,0 +1,25 @@
+import { MutableRefObject, useEffect, useRef, useState } from "react";
+
+// Same us https://usehooks.com/useHover/ but for blur and focus
+export default function useFocus(): [MutableRefObject<HTMLElement>, boolean] {
```
SergeyRoyt, line 4: How come we didn't had this hook? Are we using some other hook for it?

src/components/Chips/Chips.tsx
hadasfa: Cool!

src/components/Chips/Chips.tsx
```diff
@@ -93,21 +99,23 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
     const hasCloseButton = !readOnly && !disabled;
 
     const [hoverRef, isHovered] = useHover();
+    const [focusRef, isFocused] = useFocus();
```
hadasfa, line 102: Sorry for being evil but because our standard right now is no "is" prefix for new props for alignment, let's convert isClickable to be "backward compatibility " and add clickable prop :(

src/components/Chips/__stories__/chips.stories.mdx
```diff
@@ -113,7 +113,7 @@ Chips are compact elements that represent an input, attribute, or action. They m
   </Story>
 </Canvas>
 
-### Clickable
+### Clickable chips
```
hadasfa, line 116: Great!

src/hooks/index.js
hadasfa: yes :)

src/hooks/useFocus.ts
hadasfa: We had this: https://style.monday.com/?path=/docs/hooks-usesetfocus--overview
Is there an issue with it? https://style.monday.com/?path=/docs/hooks-usesetfocus--overview

hadasfa (approved):

src/components/Chips/Chips.tsx
SergeyRoyt: 😈

PR #1064, opened 2023-01-16
fix dropdown with modal/dialog state (support scroll inside dropdown)

hadasfa:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests

### DIFF
--- a/src/components/Dropdown/Dropdown.jsx
+++ b/src/components/Dropdown/Dropdown.jsx
@@ -141,10 +141,7 @@ const Dropdown = ({
     return mergedStyles;
   }, [size, rtl, insideOverflowContainer, insideOverflowWithTransformContainer, extraStyles, multi, multiline]);
 
-  const Menu = useCallback(
-    props => <MenuComponent innerRef={menuRef} {...props} Renderer={menuRenderer} />,
-    [menuRenderer]
-  );
+  const Menu = useCallback(props => <MenuComponent {...props} Renderer={menuRenderer} />, [menuRenderer]);
 
   const DropdownIndicator = useCallback(props => <DropdownIndicatorComponent {...props} size={size} />, [size]);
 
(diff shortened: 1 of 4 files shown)


src/components/Dropdown/Dropdown.jsx
```diff
@@ -240,14 +243,16 @@ const Dropdown = ({
     })
   };
 
+  const menuRef = useRef();
   const closeMenuOnScroll = useCallback(
     event => {
       const scrolledElement = event.target;
-      const dropdownContainer = document.getElementById(id);
-      if (dropdownContainer?.contains(scrolledElement)) return false;
+      if (scrolledElement?.parentElement?.classList.contains(MENU_WRAPPER_CLASS_NAME)) {
```
hadasfa, line 250: for some reason dropdownContainer = document.getElementById(id); return the same component but without the menu (the state of the component before opening the menu) which caused if (dropdownContainer?.contains(scrolledElement) to be false.

I can't pass there ref or id and prefer not changing the nodes structure to avoid potential css overrides issues so I choose to identify the menu by it's classes even it is not the best practice for this kind of issues.

src/components/Dropdown/Dropdown.jsx
```diff
@@ -141,7 +141,10 @@ const Dropdown = ({
     return mergedStyles;
   }, [size, rtl, insideOverflowContainer, insideOverflowWithTransformContainer, extraStyles, multi, multiline]);
 
-  const Menu = useCallback(props => <MenuComponent {...props} Renderer={menuRenderer} />, [menuRenderer]);
+  const Menu = useCallback(
+    props => <MenuComponent innerRef={menuRef} {...props} Renderer={menuRenderer} />,
+    [menuRenderer]
```
orrgottlieb, line 146: menu ref as a dep?

hadasfa: it's a mistake will revert

PR #1067, opened 2023-01-17
Checkbox: checkbox and label className props

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3822719046

### DIFF
--- a/src/components/Checkbox/Checkbox.jsx
+++ b/src/components/Checkbox/Checkbox.jsx
@@ -114,11 +114,11 @@ Checkbox.propTypes = {
   id: PropTypes.string,
   className: PropTypes.string,
   /**
-   * ClassName for icon container
+   * Class name for the checkbox marker container
    */
   checkboxClassName: PropTypes.string,
   /**
-   * ClassName for label
+   * Class name for the checkbox label
    */
   labelClassName: PropTypes.string,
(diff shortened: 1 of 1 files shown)


src/components/Checkbox/Checkbox.jsx
```diff
@@ -107,6 +113,14 @@ export const Checkbox = forwardRef(
 Checkbox.propTypes = {
   id: PropTypes.string,
   className: PropTypes.string,
+  /**
+   * ClassName for icon container
```
hadasfa, line 117:
```suggestion
   * ClassName for checkbox sketch container
```
It's not must to be this description suggestion, but in my opinion the issue here is that developers can be confused because they are not understand how we implement our fake checkbox and if it's an icon or not, so I think the description need to be focused about that we affecting the fake checkbox and not about the fake checkbox structure.

hadasfa (approved):

src/components/Checkbox/Checkbox.jsx
```diff
@@ -107,6 +113,14 @@ export const Checkbox = forwardRef(
...
+   */
+  checkboxClassName: PropTypes.string,
+  /**
+   * ClassName for label
```
hadasfa, line 121:
```suggestion
   * Class name for the checkbox's label
```

PR #1072, opened 2023-01-25
Fix tipssen hide close button

hadasfa:


orrgottlieb (approved):

src/components/Tipseen/Tipseen.tsx
```diff
@@ -74,9 +74,9 @@ const Tipseen: VibeComponent<TipseenProps> & {
       hideDelay = 0,
       showDelay = 0,
       title,
-      hideCloseButton = false,
+      hideCloseButton,
```
hadasfa, line 77: If there is a default value to hideCloseButton we will never check the value of isCloseButtonHidden

orrgottlieb: @hadasfa lets add a test please

SergeyRoyt (approved):

PR #1075, opened 2023-01-25
Tipseen: snapshot tests & fix bug on onDismiss and onSubmit overrides

hadasfa:


SergeyRoyt (approved):
Cool!
Only need to update snapshots

PR #1078, opened 2023-01-26
AccordionItem and ExpandAble preparation for css modules transition

orrgottlieb:
* Adding classnames to customize specific place in Expanable
* Improve CSS structure for Expandable

### DIFF
--- a/src/components/ExpandCollapse/ExpandCollapse.tsx
+++ b/src/components/ExpandCollapse/ExpandCollapse.tsx
@@ -7,12 +7,15 @@ import { VibeComponentProps } from "../../types";
 import cx from "classnames";
 import { ElementContent } from "../../types/ElementContent";
 import "./ExpandCollapse.scss";
+import extraStyles from "./styles.module.scss";
 
 interface ExpandCollapseProps extends VibeComponentProps {
   /**
    * Component as parameter to be rendered as header
    */
   headerComponentRenderer?: () => ReactElement;
+  headerClassName?: string;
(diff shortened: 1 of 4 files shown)


SergeyRoyt (approved):

src/components/ExpandCollapse/styles.module.scss
```diff
@@ -0,0 +1,7 @@
+.hideBorder {
```
SergeyRoyt, line 1: Maybe to rename the file to `ExpandCollapse.module.scss` or smth? Just to keep the convention

src/components/ExpandCollapse/ExpandCollapse.tsx
```diff
@@ -7,12 +7,15 @@ import { VibeComponentProps } from "../../types";
 import cx from "classnames";
 import { ElementContent } from "../../types/ElementContent";
 import "./ExpandCollapse.scss";
+import extraStyles from "./styles.module.scss";
```
SergeyRoyt, line 10: Maybe to rename the file to`ExpandCollapse.module.scss` or smth? Just to keep the convention

SergeyRoyt: Or maybe it's better to leave it this way - would be less confusing when I'll merge with Vibe 2.0

orrgottlieb: thought about 2.0 so afterwards we would just merge the both

PR #1080, opened 2023-01-29
Skeleton - wrapperClassName and fullWidth props added

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3889970261

### DIFF
--- a/src/components/Skeleton/Skeleton.scss
+++ b/src/components/Skeleton/Skeleton.scss
@@ -46,6 +46,12 @@
 }
 
 .monday-style-skeleton {
+  display: flex;
+
+  .full-width {
+    width: 100%;
+  }
+
   &_rectangle {
     $rectangle-default-size: 160px;
     height: $rectangle-default-size;
--- a/src/components/Skeleton/Skeleton.scss
+++ b/src/components/Skeleton/Skeleton.scss
@@ -48,10 +48,6 @@
 .monday-style-skeleton {
   display: flex;
 
-  .full-width {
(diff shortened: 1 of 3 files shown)


src/components/Skeleton/Skeleton.scss
```diff
@@ -46,6 +46,12 @@
 }
 
 .monday-style-skeleton {
+  display: flex;
+
+  .full-width {
```
hadasfa, line 51: "full-width" is very generic class name, the danger of using a very generic class name is that we can receive by mistake styles which defined by other developers in the monolith which will affect our skeleton.
@orrgottlieb suggest that we will create a new scss modules file for the skeleton and add only this class to there for avoiding conflicits. Wdty?

src/components/Skeleton/Skeleton.tsx
```diff
@@ -19,17 +19,30 @@ interface SkeletonProps extends VibeComponentProps {
   size?: keyof typeof SKELETON_ALLOWED_SIZES;
   width?: number;
   height?: number;
+  wrapperClassName?: string;
+  /**
+   * Is wrapper width = 100%
```
hadasfa, line 24:
```suggestion
   * Is the skeleton wrapper size equal to its container
```

hadasfa (approved):

src/components/Skeleton/Skeleton.scss
SergeyRoyt: Moved it to css modules 👍

PR #1085, opened 2023-02-01
AlertBannerLink - className prop added

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3908730986

### DIFF
--- a/src/components/AlertBanner/AlertBannerLink/AlertBannerLink.tsx
+++ b/src/components/AlertBanner/AlertBannerLink/AlertBannerLink.tsx
@@ -9,15 +9,23 @@ export interface AlertBannerLinkProps extends LinkProps {
   isDarkBackground?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


src/components/AlertBanner/AlertBannerLink/AlertBannerLink.tsx
```diff
@@ -9,15 +9,23 @@ export interface AlertBannerLinkProps extends LinkProps {
...
+      "monday-style-alert-banner-link-margin-left": marginLeft,
+      "monday-style-alert-banner-link-dark-background": isDarkBackground
+    },
+    className
+  );
 
   return (
     <div className={classNames}>
-      <Link {...linkProps} componentClassName="monday-style-alert-banner-link" />
+      <Link {...linkProps} className="monday-style-alert-banner-link" />
```
hadasfa, line 28: Because the reason for this is to apply style which should be the style of the link already let's wait with merge this until understand why it's needed :)

PR #1087, opened 2023-02-02
Counter: counterClassName added

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3918610983


hadasfa (approved):
let's add a snapshot test with the new prop before merge?

PR #1094, opened 2023-02-06
Modal: Add zIndex prop

SergeyRoyt:
https://monday.monday.com/boards/3015775582/pulses/3934480198


src/components/Modal/Modal.tsx
```diff
@@ -95,6 +99,12 @@ const Modal: FC<ModalProps> & { width?: typeof ModalWidth } = ({
   // lock body scroll when modal is open
   useBodyScrollLock({ instance });
 
+  useEffect(() => {
+    if (attr?.container?.ref?.current) {
+      attr.container.ref.current.style.setProperty("--monday-modal-z-index", zIndex.toString());
+    }
+  }, [attr.container.ref, zIndex]);
```
orrgottlieb, line 106: just make sure that we always have `container`

orrgottlieb (approved):

PR #1101, opened 2023-02-08
Modal: fix multiple onClose callbacks issue

SergeyRoyt:
https://monday.monday.com/boards/3015775582/pulses/3949733333

### DIFF
--- a/src/components/Modal/useShowHideModal.ts
+++ b/src/components/Modal/useShowHideModal.ts
@@ -60,7 +60,9 @@ export default function useShowHideModal({
     return () => {
       instance?.off("hide");
     };
-  }, [instance, onClose]);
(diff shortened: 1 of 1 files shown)


src/components/Modal/useShowHideModal.ts
```diff
@@ -60,7 +60,9 @@ export default function useShowHideModal({
     return () => {
       instance?.off("hide");
     };
-  }, [instance, onClose]);
+    // needs to be run only once - to avoid multiple callbacks
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
SergeyRoyt, line 64: not sure it's the best (and the right) solution here. Wdyt?

PR #1105, opened 2023-02-12
Set chips clickable behavior as default instead of extra prop for backward support and clearer API 

hadasfa:
Set chips clickable behavior as default instead of extra prop for backward support and clearer API + tests

### DIFF
--- a/src/components/Chips/Chips.tsx
+++ b/src/components/Chips/Chips.tsx
@@ -177,7 +177,7 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
         }
       : {
           ref: mergedRef,
-          onClick: onClickCallback,
+          onClick: onClick ? onClickCallback : undefined,
           onMouseDown,
           id: id,
           "data-testid": overrideDataTestId,
--- a/src/components/Chips/Chips.tsx
+++ b/src/components/Chips/Chips.tsx
@@ -14,7 +14,6 @@ import { SubIcon, VibeComponent, VibeComponentProps } from "../../types";
 import useHover from "../../hooks/useHover";
 import useSetFocus from "../../hooks/useSetFocus";
 import { ComponentDefaultTestId } from "../../tests/constants";
-import { backwardCompatibilityForProperties } from "../../helpers/backwardCompatibilityForProperties";
 import useClickableProps from "../../hooks/useClickableProps";
 import { BEMClass } from "../../helpers/bem-helper";
 import "../Clickable/Clickable.scss";
@@ -66,13 +65,9 @@ interface ChipsProps extends VibeComponentProps {
    */
   ariaLabel?: string;
   /**
-   * Should element be focusable & clickable - for backward compatability
+   * If you are using onMouseDown or onClick but do not want your chip to be used as a button (in terms of appearance or aria attributes)
    */
-  clickable?: boolean;
-  /**
-   * Backward compatibility for props naming - please use clickable instead
-   */
-  isClickable?: boolean;
+  disableClickableBehavior?: boolean;
 }
 
 const Chips: VibeComponent<ChipsProps, HTMLElement> & {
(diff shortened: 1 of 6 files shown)


src/components/Chips/Chips.tsx
```diff
@@ -66,13 +65,9 @@ interface ChipsProps extends VibeComponentProps {
    */
   ariaLabel?: string;
   /**
-   * Should element be focusable & clickable - for backward compatability
+   * If you are using onMouseDown or onClick but do not want your chip to be used as a button (in terms of appearance or aria attributes)
```
hadasfa, line 68: Until now if you were using onClick or onMouseDown it would not work unless you used also the clickable/isClickable extra prop which was breaking change (this prop is new) and also create less convenient API  (because we want that the chips will contain the correct behavior as default)

src/components/Chips/Chips.tsx
```diff
@@ -184,6 +177,8 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
         }
       : {
           ref: mergedRef,
+          onClick: onClickCallback,
```
hadasfa, line 180: If someone using onclick/onmousedown as hack and not want to treat chip as a button (like we do in drop down multi state) we still want that the onClick/onMouseDown will work as usual (but without hover appearance changes or aria attributes)

src/components/Chips/__tests__/chips-tests.jest.tsx
```diff
@@ -1,21 +1,55 @@
 import React from "react";
 import { render, fireEvent, screen } from "@testing-library/react";
 import Chips from "../Chips";
+import { getTestId } from "../../../tests/test-ids-utils";
```
hadasfa, line 4: The reason we were not aware for the braking changes which happen because of this bug was lack of tests for this ability, so I add some...

src/components/Chips/Chips.tsx
```diff
@@ -184,6 +177,8 @@ const Chips: VibeComponent<ChipsProps, HTMLElement> & {
         }
       : {
           ref: mergedRef,
+          onClick: onClick ? onClickCallback : undefined,
```
hadasfa, line 180: If someone using onclick/onmousedown as hack and not want to treat chip as a button (like we do in drop down multi state) we still want that the onClick/onMouseDown will work as usual (but without hover appearance changes or aria attributes)

src/components/Chips/Chips.tsx
orrgottlieb, line 68: what do you mean by "appearance"

orrgottlieb: removing these will break if people are using it with typescript, you can do but in Version 2.0

src/components/Chips/Chips.tsx
orrgottlieb: i would love to use a prop and not a function to check

src/components/Chips/__tests__/chips-tests.jest.tsx
```diff
@@ -1,21 +1,55 @@
...
 
-  it("should call the click callback when clicked", () => {
+  it("Should call the onDelete callback when on close button clicked", () => {
     const testId = `${Chips.defaultTestId}-close`;
+    const onDeletedMock = jest.fn();
+    render(<Chips className={className} onDelete={onDeletedMock} label={label} />);
     fireEvent.click(screen.getByTestId(testId));
     expect(onDeletedMock.mock.calls.length).toBe(1);
   });
+  it("Should call onClick callback when chips clicked", () => {
```
orrgottlieb, line 19: very cool!!!

orrgottlieb (changes requested):

src/components/Chips/Chips.tsx
hadasfa: chips color is change while hovering for now

src/components/Chips/Chips.tsx
```diff
@@ -66,13 +65,16 @@ interface ChipsProps extends VibeComponentProps {
    */
   ariaLabel?: string;
   /**
-   * Should element be focusable & clickable - for backward compatability
+   * Deprecated, there is no need to use this prop for implementing clickable chips. Please use onClick for this purpose.
+   * @deprecated
```
orrgottlieb, line 69: noice!! maybe worth adding this for other props as well - WDYT?

orrgottlieb (approved):

src/components/Chips/Chips.tsx
hadasfa: can be done :)

PR #1108, opened 2023-02-13
Docs/hadas/modal improvements

hadasfa:
fix code examples and docs, add some do's and don'ts with examples


src/components/Modal/ModalButtons/ModalButtons.tsx
```diff
@@ -0,0 +1,39 @@
+import React, { FC } from "react";
+import VibeComponentProps from "../../../types/VibeComponentProps";
+import Button from "../../Button/Button";
+import Flex from "../../Flex/Flex";
+import { ComponentDefaultTestId } from "../../../tests/constants";
+import { getTestId } from "../../../tests/test-ids-utils";
+
+interface ModalButtonsProps extends VibeComponentProps {
+  className: string;
```
hadasfa, line 9: It's used to be internal component only for docs, But I guess we want alignment between all instance appearance of the modal footer buttons

orrgottlieb: this should not be here, should be in Atmosphere please remove

src/components/index.js
```diff
@@ -79,7 +79,8 @@ export { default as Flex } from "./Flex/Flex";
...
 export { default as AvatarGroup } from "./AvatarGroup/AvatarGroup";
 export { default as Modal } from "./Modal/Modal";
-export { default as ModalHeader } from "./ModalHeader/ModalHeader";
-export { default as ModalContent } from "./ModalContent/ModalContent";
-export { default as ModalFooter } from "./ModalFooter/ModalFooter";
+export { default as ModalHeader } from "./Modal/ModalHeader/ModalHeader";
+export { default as ModalContent } from "./Modal/ModalContent/ModalContent";
+export { default as ModalFooter } from "./Modal/ModalFooter/ModalFooter";
 export { default as Box } from "./Box/Box";
+export { default as ModalButtons } from "./Modal/ModalButtons/ModalButtons";
```
orrgottlieb, line 86: please remove

orrgottlieb (changes requested):

PR #1110, opened 2023-02-14
useHover - hook story

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3949034925


src/components/Chips/Chips.tsx
```diff
@@ -11,7 +11,7 @@ import { getTestId } from "../../tests/test-ids-utils";
 import { ChipsSize } from "./ChipsConstants";
```
hadasfa, line 11: is useHover part of index and publishedcomponents files? if not, let's add it to there (otherwise it will not be expose)

src/hooks/useHover/__stories__/useHover.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+# useHover
+
+- [Overview](#overview)
+- [Arguments](#arguments)
+- [Returns](#returns)
+- [Feedback](#feedback)
+
+## Overview
+
+Attaches a listener for <code>mouseover</code> and <code>mouseout</code> DOM events on a specific element, returning <code>isHovered</code> state.
```
hadasfa, line 17: I think that this explanation is too technical and about the implementation details 
What about something like: 
```suggestion
Detect whether the mouse is hovering an element by returning is <code>isHovered</code> state.
```
feel free to change it according to what you think is best.

src/hooks/useHover/__stories__/useHover.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+    }}
+  </Story>
+</Canvas>
+
+## Usage
+
+<UsageGuidelines
+  guidelines={[
+    "Use this hook to get element hovered state",
+    "No arguments are needed just assign returned ref to the element"
```
hadasfa, line 37:
```suggestion
    "Please assign the returned hoverRef as the reference of the element"
```

src/hooks/useHover/__stories__/useHover.stories.mdx
```diff
@@ -0,0 +1,58 @@
...
+
+## Returns
+
+<FunctionArguments>
+  <FunctionArgument
+    name="ref"
+    type="React.MutableRefObject"
+    description={
+      <>
+        A React{" "}
```
hadasfa, line 49: let's change to: React ref object to assign to the element which we want to track its hover state

and set the link over react ref.

by the way the link is cool!

src/components/Chips/Chips.tsx
SergeyRoyt: Thanks for noticing!

src/hooks/useActiveDescendantListFocus/__stories__/useActiveDescendantListFocus.stories.mdx
```diff
@@ -1,14 +1,14 @@
...
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
 import useActiveDescendantListFocus from "../index";
 import UseActiveDescendantListFocus from "./hooksDummyComponents/UseActiveDescendantListFocus";
 import { FunctionArgument, FunctionArguments, UnstyledList, UnstyledListItem } from "../../../storybook";
-import styles from "./useActiveDescendantListFocus.module.scss";
 import { overviewInteractionSuite } from "../__tests__/useActiveDescendantListFocus.interactions";
+import styles from "./useActiveDescendantListFocus.module.scss";
 
-<Meta title="Hooks/use Active Descendant List Focus" component={UseActiveDescendantListFocus} />
+<Meta title="Hooks/useActiveDescendantListFocus" component={UseActiveDescendantListFocus} />
```
hadasfa, line 11: I am not sure what is our standard regarding spacing. I think in components we want spacing between words but in hooks we don't it... So I guess this change is fine (but please no do changes like this in components docs)

webpack/published-ts-components.js
```diff
@@ -92,9 +94,8 @@ const publishedTSComponents = {
   useIsOverflowing: "hooks/useIsOverflowing.ts",
   useMediaQuery: "hooks/useMediaQuery",
   useVibeMediaQuery: "hooks/useVibeMediaQuery",
-  useActiveDescendantListFocus: "hooks/useActiveDescendantListFocus"
-  // Don't remove next line
-  // plop_marker:published-hooks
+  useActiveDescendantListFocus: "hooks/useActiveDescendantListFocus",
+  useHover: "hooks/useHover/useHover.ts"
```
hadasfa, line 98: Our standard is not to add suffix (I know we there are suffix in some files but it's also mistakes)
```suggestion
  useHover: "hooks/useHover/useHover"
```

hadasfa (approved):
Approved with some notes, please check before merging :)

PR #1112, opened 2023-02-14
useClickableProps - hook story

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3949035686


src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+# useClickableProps
+
+- [Overview](#overview)
+- [Arguments](#arguments)
+- [Returns](#returns)
+- [Feedback](#feedback)
+
+## Overview
+
+Return props for adding clickable functionality to the element except for the styles and classNames.
```
hadasfa, line 18:
```suggestion
Return props for making an element or component clickable by mouse and keyboard with screen reader support.
```

The important thing about this hook is that it make element to clickable in accessible way. It's less important to mention that we are not supporting style props because it's not related to the hook's purpose

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+<Canvas>
+  <Story name="Overview">
+    {() => {
+      const ref = useRef(null);
+      const onClickCallback = useCallback(() => alert("click!"), []);
```
hadasfa, line 24:
```suggestion
      const onClick= useCallback(() => alert("click!"), []);
```
callback is redundant

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+      return (
+        <div {...clickableProps} className="monday-storybook-use-clickable-props">
+          I act like a button
+        </div>
+      );
+    }}
+  </Story>
+</Canvas>
+
+<Tip title="Alternative">
```
hadasfa, line 44: mmm not sure it's the right title in terms of UX. this is my suggestion, worth to also check this with Meirav:
```suggestion
<Tip title="Check out our component solution for this use case">
```

And then the content can be something like: 
"For most simple use cases of clickable elements, you also can use our Clickable component wrapper (Higher order component). "

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
+import { useCallback, useRef } from "react";
+import { Canvas, Meta, Story } from "@storybook/addon-docs";
+import { Link, Tip } from "../../../storybook";
+import useClickableProps from "../useClickableProps";
+import "./useClickableProps.stories.scss";
+
+<Meta title="Hooks/useClickableProps" />
+
+# useClickableProps
```
hadasfa, line 9: Also worth to add the following tips in the Clickable component:
Tip
If you'd like to set clickable functionality on a specific element inside your React component instead of using a wrapper, you can also look in our useClickableProps hook.

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+    description={<>Click callback.</>}
+  />
+  <FunctionArgument
+    name="onMouseDown"
+    type="(event: React.MouseEvent) => void"
+    description={<>Mouse down callback.</>}
+  />
+  <FunctionArgument name="disabled" type="boolean" description={<>Whether element is disabled or not</>} />
+  <FunctionArgument name="id" type="string" description={<>Id of the element</>} />
+  <FunctionArgument name="dataTestId" type="string" description={<>Id of the element for test purposes</>} />
```
hadasfa, line 67: wdyt? not a muse
```suggestion
  <FunctionArgument name="dataTestId" type="string" description={"Set an element's id for testing purposes"} />
```

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+  <FunctionArgument
+    name="role"
+    type="string"
+    description={
+      <>
+        Provide semantic meaning to content.
+        <Link size={Link.sizes.SMALL} href={"https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles"}>
+          More details
```
hadasfa, line 75: wdyt?
```suggestion
         Read more about this here.
```

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+        </Link>
+      </>
+    }
+  />
+  <FunctionArgument
+    name="tabIndex"
+    type="number"
+    description={
+      <>
+        Specifies the tab order of an element.
```
hadasfa, line 85: There is a way to mention what is the default value? I think it is important. Also worth to add a link here to mdn explanation about what is tab index.
```suggestion
        Specifies the tab order of the element.
```

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+        <Link
+          size={Link.sizes.SMALL}
+          href={"https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex"}
+        >
+          More details
```
hadasfa, line 90: same comment

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
hadasfa: if there is no way to mention default value in the function argument props, let's add the default value to the description?

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+        </Link>
+      </>
+    }
+  />
+  <FunctionArgument
+    name="ariaLabel"
+    type="string"
+    description={
+      <>
+        Defines a string value that labels an interactive element.
```
hadasfa, line 100: Maybe too long? but I think it will be clearer this way... wdyt? 

Aria label is an attribute used in HTML to provide a human-readable label for interactive elements without visible text. This attribute is important for improving web accessibility, specifically for users who rely on screen readers, as it allows them to understand the purpose and function of the element.

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+        <Link
+          size={Link.sizes.SMALL}
+          href={"https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label"}
+        >
+          More details
```
hadasfa, line 105: same comment

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+        </Link>
+      </>
+    }
+  />
+  <FunctionArgument
+    name="ariaHidden"
+    type="boolean"
+    description={
+      <>
+        Used to hide non-interactive content from the accessibility API.
```
hadasfa, line 115: The description is not accurate enough, maybe: 
```suggestion
        Aria hidden is an HTML attribute for hiding content from screen readers and other assistive technologies.
```

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+        <Link
+          size={Link.sizes.SMALL}
+          href={"https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden"}
+        >
+          More details
```
hadasfa, line 120: same comment

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
```diff
@@ -0,0 +1,275 @@
...
+
+## Returns
+
+<FunctionArguments>
+  <FunctionArgument
+    name="ref"
+    type="(node: HTMLElement) => void"
+    description={
+      <>
+        A React
```
hadasfa, line 166: React ref object for the clickable element?

webpack/published-ts-components.js
```diff
@@ -92,9 +92,10 @@ const publishedTSComponents = {
   useIsOverflowing: "hooks/useIsOverflowing.ts",
   useMediaQuery: "hooks/useMediaQuery",
   useVibeMediaQuery: "hooks/useVibeMediaQuery",
-  useActiveDescendantListFocus: "hooks/useActiveDescendantListFocus"
+  useActiveDescendantListFocus: "hooks/useActiveDescendantListFocus",
   // Don't remove next line
   // plop_marker:published-hooks
+  useClickableProps: "hooks/useClickableProps/useClickableProps.ts"
```
hadasfa, line 98: please remove .ts (I know we include this in other places but this is also a mistake)

hadasfa (approved):
Approved with comments :) please fix before merging

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
SergeyRoyt: Too verbose I think. Takes too much space

src/hooks/useClickableProps/__stories__/useClickableProps.stories.mdx
SergeyRoyt: I think it's too long - we are providing links to read about it and I assume that most of our users are aware about `aria-label` purpose

PR #1113, opened 2023-02-14
Fix responsive dos and dont's by change min width and fix tipseen story

hadasfa:
Change min width of do/don't rule in stories
override default for tipseen (the examples are bigger there

### DIFF
--- a/src/components/Tipseen/__stories__/tipseen.stories.mdx
+++ b/src/components/Tipseen/__stories__/tipseen.stories.mdx
@@ -19,19 +19,20 @@ import "./tipseen.stories.scss";
 
 export const metaSettings = createStoryMetaSettings({
   component: Tipseen,
-  enumPropNamesArray: ["position", "animationType", "justify"],
+  enumPropNamesArray: ["position", "animationType", "justify"]
 });
 
 <Meta
   title="Popover/Tipseen"
-  component={ Tipseen }
-  argTypes={ metaSettings.argTypes }
-  decorators={ metaSettings.decorators }
+  component={Tipseen}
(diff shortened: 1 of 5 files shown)


src/components/Tipseen/__stories__/tipseen.stories.mdx
```diff
@@ -145,74 +157,80 @@ Use when the tip is too long or when you want to teach something in steps.
...
-      )
+      );
     }}
   </Story>
 </Canvas>
 
 ## Do’s and Don’ts
+
 <ComponentRules
+  className="monday-storybook-tipseen_rules"
```
hadasfa, line 175: the other things in this file are lint related, this is the only real change

src/components/Tipseen/__stories__/tipseen.stories.scss
```diff
@@ -1,7 +1,19 @@
 @import "../../../styles/themes";
+@import "../../../styles/global-css-settings";
 @import "../../../styles/typography";
 
+$component-rules-grid-gap: $spacing-xxl;
```
hadasfa, line 5: it's not beautiful I know :(

hadasfa:
It's not a beautiful solution but it is working and I did not want to refactor the mixin and damage it for give an override option which right now needed only for one use case (I thought it will be overkill)

src/components/Tipseen/__stories__/tipseen.stories.scss
SergeyRoyt: why u need individual variables for that?

SergeyRoyt (approved):

PR #1122, opened 2023-02-19
Basic dialog docs (only basic variants and examples)

hadasfa:
there is a lot of things which not here yet (modifiers, ignore class, hide reference when hidden) but I think it's a good start. I stopped here because I wanted to be able to finish more tasks in the hackathon 
Also contain:
1. skeleton ts fixes
2. skeleton full width fixes
3. exporting use switch to external usage (and some modifications in it for make it relevant not only to checkbox) 
4. small aria fix for button active


src/components/AvatarGroup/AvatarGroupCounterTooltipContentVirtualizedList.tsx
```diff
@@ -12,7 +12,7 @@ const LIST_OPTIONS = Object.freeze({
 });
 
 export type AvatarItem = {
-  value: AvatarProps & { tooltipContent: string | ReactElement };
+  value: AvatarProps & { tooltipContent: string | ReactElement | ReactElement[] };
```
hadasfa, line 15: Because it was throwing error because the tooltip receive a list of people info to display in the tooltip content instead of one react element

src/components/Button/Button.tsx
```diff
@@ -264,11 +264,11 @@ const Button: VibeComponent<ButtonProps, unknown> & {
         "aria-label": ariaLabel,
         "aria-haspopup": ariaHasPopup,
         "aria-expanded": ariaExpanded,
-        "aria-controls": ariaControls
+        "aria-controls": ariaControls,
+        "aria-pressed": active
```
hadasfa, line 268: for active state to be more accessible

src/components/Dialog/__stories__/Dialog.stories.scss
```diff
@@ -0,0 +1,26 @@
+@import "../../../styles/themes.scss";
```
hadasfa, line 1: did not implement it with css modules because we have an issue with typescript+css modules in stories, tried to understand what is not working but did not want to waste time on this because out of this task scope

src/components/Skeleton/Skeleton.module.scss
```diff
@@ -1,3 +1,17 @@
+@import "./SkeletonVariables.module";
```
hadasfa, line 1: full width did not work on text skeleton. fixed it for using it in the dialog docs

hadasfa: I am using global because the rest of skeleton css is not converted to css modules

src/components/Switch/hooks/useSwitch.ts
SergeyRoyt, line 1: don't we have same hook here `src/hooks/useSwitch/index.ts`?

src/components/Dialog/__stories__/Dialog.stories.mdx
```diff
@@ -0,0 +1,506 @@
...
+          >
+            <Button
+              onFocus={onFocusButton}
+              onBlur={onFocusButton}
+              kind={Button.kinds.SECONDARY}
+              active={focusButtonActive}
+            >
+              On focus
+            </Button>
+          </Dialog>
```
SergeyRoyt, line 329: Wasn't able to see this dialog after keyboard focus

hadasfa: fixed

SergeyRoyt (approved):

PR #1124, opened 2023-02-21
Dialog, Tooltip addKeyboardHideShowTriggersByDefault prop

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/3949004433

### DIFF
--- a/src/components/Dialog/Dialog.tsx
+++ b/src/components/Dialog/Dialog.tsx
@@ -132,6 +132,10 @@ export interface DialogProps extends VibeComponentProps {
   getDynamicShowDelay?: () => { showDelay: number; preventAnimation: boolean };
   content?: (() => JSX.Element) | JSX.Element;
   children?: ReactElement | ReactElement[] | string;
+  /**
+   * Treats keyboard focus/blur events as mouse-enter/mouse-leave events
+   */
+  addKeyboardHideShowTriggersByDefault?: boolean;
 }
 
 interface DialogState {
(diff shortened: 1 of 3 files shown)


src/components/Dialog/Dialog.tsx
```diff
@@ -132,6 +132,10 @@ export interface DialogProps extends VibeComponentProps {
   getDynamicShowDelay?: () => { showDelay: number; preventAnimation: boolean };
   content?: (() => JSX.Element) | JSX.Element;
   children?: ReactElement | ReactElement[] | string;
+  /**
```
hadasfa, line 135: how complicated is to add test for this ability? (if flag is on and hover trigger pass check if focus trigger the dialog)

hadasfa: after removing the flag we will remove the flag from the test also

src/components/Dialog/Dialog.tsx
```diff
@@ -351,13 +356,29 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
   }
 
   isShowTrigger(eventName: HideShowEvent) {
-    const { showTrigger } = this.props;
-    return convertToArray(showTrigger).indexOf(eventName) > -1;
+    const { showTrigger, addKeyboardHideShowTriggersByDefault } = this.props;
```
hadasfa, line 359: this one feel a bit hard coded but I guess there is no prettier solution to do  now.
Maybe when we will add enter event support we could refactor it a bit.

src/components/Dialog/Dialog.tsx
hadasfa: maybe let's also add test that this ability not working when the flag is off

SergeyRoyt: will try to write the test 👍

hadasfa (approved):

PR #1125, opened 2023-02-22
Optimize style injection

orrgottlieb:
optimizing the css injection in runtime


rollup.config.js
```diff
@@ -55,8 +56,37 @@ export default {
...
+  
+    if (style.styleSheet) {
+      style.styleSheet.cssText = css
+    } else {
+      style.appendChild(document.createTextNode(css))
+    }
+  }
+  
+  styleInject(${cssVariableName}, { insertAt: 'top' });
+
```
sahariko, line 88: Kinda hard to read it like that, how about we extract it to a `.js` file so we get syntax highlighting etc. and do `fs.readFileSync` instead?

rollup.config.js
```diff
@@ -55,8 +56,37 @@ export default {
        * we're externalizing all "node_modules", it doesn't exist.
        * This little hack makes sure we're using "node_modules" instead of what the plugin expects.
        */
-      inject(cssVariableName) {
-        return `import styleInject from 'style-inject';\nstyleInject(${cssVariableName}, { insertAt: 'top' });`;
+      inject: function (cssVariableName) {
+        const hashValue = `s_id-${version}-${performance.now()}`.replaceAll(".", "_");
+        return `function styleInject(css, { insertAt } = {}) {
+    const head = document.head || document.getElementsByTagName('head')[0]
```
sahariko, line 62: Do all our browsers support `const`?

rollup.config.js
```diff
@@ -55,8 +56,37 @@ export default {
...
+  
+    if (insertAt === 'top') {
+      if (head.firstChild) {
+        head.insertBefore(style, head.firstChild)
+      } else {
+        head.appendChild(style)
+      }
+    } else {
+      head.appendChild(style)
+    }
```
sahariko, line 78: Can be simplified to:
```suggestion
    if (head.firstChild) {
      head.insertBefore(style, head.firstChild);
    } else {
      head.appendChild(style);
    }
```

And since we now control the implementation of `styleInject`, we can simplify it by not passing `injectAt` at all.

sahariko (commented):
Since we're controlling the implementation of `styleInject`, let's remove this dependency.

sahariko (approved):

PR #1128, opened 2023-02-23
Skeleton types + aria pressed + switch

hadasfa:
changes related to the dialog docs which not depend on them


src/hooks/useSwitch/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  defaultChecked,
+  onChange,
+  isDisabled
+}: {
+  isChecked?: boolean;
+  defaultChecked?: boolean;
+  onChange?: (value: boolean) => void;
+  isDisabled?: boolean;
+}) {
+  const overrideCheckedInitial = isChecked === undefined ? (defaultChecked ? defaultChecked : false) : isChecked;
```
SergeyRoyt, line 19: `(defaultChecked ? defaultChecked : false)` - strange construction
Maybe just `!!defaultChecked `

src/hooks/useSwitch/index.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+    const newChecked = !overrideChecked;
+    if (isChecked === undefined) {
+      setOverrideChecked(newChecked);
+    }
+    onChange && onChange(newChecked);
+  }, [isChecked, isDisabled, onChange, overrideChecked]);
+
+  useEffect(() => {
+    isChecked !== undefined && setOverrideChecked(isChecked);
```
SergeyRoyt, line 34: probably if statement would be clearer here

src/components/Skeleton/SkeletonConstants.ts
```diff
@@ -1,32 +1,35 @@
...
+  H3 = "h3",
+  H4 = "h4",
+  H5 = "h5",
+  H6 = "h6",
+  P = "p",
+  SMALL = "small"
+}
 
-export const SKELETON_ALLOWED_TYPES = Object.values(SKELETON_TYPES);
+export enum ShapeSkeletonSize {
```
SergeyRoyt, line 21: why enum is better here than just const? Also the name is rather strange

src/components/Skeleton/SkeletonConstants.ts
```diff
@@ -1,32 +1,35 @@
 // Fixed names
-export const SKELETON_CIRCLE_TYPE = "circle";
-export const SKELETON_TEXT_TYPE = "text";
-export const SKELETON_RECTANGLE_TYPE = "rectangle";
 export const SKELETON_CUSTOM_SIZE = "custom";
```
SergeyRoyt, line 2: and this const is still here. Maybe just leave it as it is?

src/components/Skeleton/Skeleton.tsx
```diff
@@ -28,27 +28,31 @@ interface SkeletonProps extends VibeComponentProps {
...
   return (
     <div className={cx(SKELETON_CSS_BASE_CLASS, wrapperClassName, { [styles.fullWidth]: fullWidth })}>
       <div
         className={cx(
           bemHelper({ element: skeletonType }),
           bemHelper({ element: skeletonType, state: skeletonSize }),
-          className
+          className,
+          {
+            [styles.fullWidth]: bemHelper({ element: skeletonType, state: "full-width" })
```
SergeyRoyt, line 54: got confused here... why bemHelper result is a condition for applying `styles.fullWidth`?

src/components/Skeleton/Skeleton.module.scss
```diff
@@ -1,3 +1,17 @@
+@import "SkeletonVariables";
+
 .fullWidth {
   width: 100%;
 }
+
+:global(.monday-style-skeleton):not(.fullWidth) {
```
SergeyRoyt, line 7: Probably better to find some other way here, as we want to remove old classnames soon

SergeyRoyt (approved):
Wrote a few notes. Nothing critical, only maybe using `:global` pseudo-class

src/components/Skeleton/Skeleton.module.scss
hadasfa: at the moment we will remove old class name we will replace :global(.monday-style-skeleton) with the new skeleton css class name

src/components/Skeleton/Skeleton.tsx
hadasfa: you righttt i will fix

src/components/Skeleton/SkeletonConstants.ts
hadasfa: The reason I had to change it was I had issues using the skeleton with typescript in the previous version (when I used the skeleton for the dialog). So I must do change there for typescript users.
There is 2 types here: shape type (circle, rectangle and text) and size (there is a enum for text skeleton and only one valid size or the other skeleton types). because of this weird structure which I can't break, I create 2 enums: skeleton type and skeleton text size and also create a const for the size of circle or rectangle.
I admit the name is a bit weird, If you have a better idea we can change

src/components/Skeleton/SkeletonConstants.ts
hadasfa: removed it also :)

PR #1183, opened 2023-03-23
change the build of css modules to calculate the hash according to co…

orrgottlieb:


rollup.config.js
```diff
@@ -59,7 +59,14 @@ export default {
         return `import styleInject from 'style-inject';\nstyleInject(${cssVariableName}, { insertAt: 'top' });`;
       },
       plugins: [autoprefixer(), postCssImport()],
-      autoModules: true
+      modules: {
+        generateScopedName: function (name, filename, css) {
+          const start = css.indexOf(`${name} {`);
+          const end = css.indexOf("}", start);
+          const content = css.slice(start + name.length + 1, end).replace(/[\r\n]/, "");
+          return `${name}_${sha256(content).slice(0, 10)}`;
```
sahariko, line 67: This could result in very long classnames, I think today we're using 5-6 letters.

Also, not sure if intentional, but this will keep the original classname as part of this in production, which I think we don't want.

orrgottlieb: currently we have it everywhere, looked at class names which are shorter than what we have now

SergeyRoyt (approved):

PR #1188, opened 2023-03-26
Fix/sergeyro/heading overflow

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4191044291

### DIFF
--- a/src/styles/typography.scss
+++ b/src/styles/typography.scss
@@ -87,7 +87,7 @@ $small-empty-state-height: 12px;
 }
 
 @mixin heading-padding {
-  padding: 0 4px;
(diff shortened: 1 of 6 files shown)


src/styles/typography.scss
```diff
@@ -87,7 +87,7 @@ $small-empty-state-height: 12px;
 }
 
 @mixin heading-padding {
-  padding: 0 4px;
+  padding: 0.5px 4px;
```
orrgottlieb, line 90: @SergeyRoyt this looks like a "hack" what is the root issue ? maybe this changed because of Poppins ?

orrgottlieb (approved):

PR #1205, opened 2023-04-09
fix expand collapse border bug

hadasfa:
Fix bug: when prop hideBorder is true bottom border is still displayed

### DIFF
--- a/src/components/ExpandCollapse/ExpandCollapse.scss
+++ b/src/components/ExpandCollapse/ExpandCollapse.scss
@@ -32,12 +32,6 @@
     color: var(--icon-color);
   }
 }
-
-.expand-collapse__header--open {
-  border-bottom: 1px solid;
-  border-color: var(--ui-border-color);
-}
-
 .expand-collapse__content {
   padding: var(--spacing-medium);
 }
--- a/src/components/MenuButton/__stories__/MenuButton.stories.mdx
+++ b/src/components/MenuButton/__stories__/MenuButton.stories.mdx
@@ -13,7 +13,8 @@ import MoveArrowDown from "../../Icon/Icons/components/MoveArrowDown";
 
 export const metaSettings = createStoryMetaSettings({
   component: MenuButton,
-  iconPropNamesArray: ["component"]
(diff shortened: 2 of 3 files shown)


src/components/MenuButton/__stories__/MenuButton.stories.mdx
```diff
@@ -13,7 +13,8 @@ import MoveArrowDown from "../../Icon/Icons/components/MoveArrowDown";
 
 export const metaSettings = createStoryMetaSettings({
   component: MenuButton,
-  iconPropNamesArray: ["component"]
+  iconPropNamesArray: ["component"],
```
hadasfa, line 16: Docs fix I was to lazy to split to a different pr

src/components/ExpandCollapse/ExpandCollapse.scss
```diff
@@ -32,12 +32,6 @@
     color: var(--icon-color);
   }
 }
-
-.expand-collapse__header--open {
```
shlomitc, line 36: just making sure it's ok to remove from the non-module file

shlomitc (approved):

src/components/ExpandCollapse/ExpandCollapse.scss
hadasfa: It will only be a problem if there are some mf users who haven't updated their vibe version to 1.86.0 or higher and are using the Expand collapse / Accordion components. However, this issue is already happening for those users who didn't update, because of a previous fix that we implemented in the same area, and it's already part of the latest version of vibe package.

If a mf hasn't updated their version, the bottom border of their accordion/expand collapse components may be hidden (which isn't a frequently used component anyway) until they update their version.

hadasfa: It's worth mentioning that we regularly deal with CSS bugs due to conflicts between the CSS on the MFS and the monolith. We can't avoid these conflicts until we complete the CSS module migration because in not completely migrated components, different versions are using the same class names.

PR #1208, opened 2023-04-13
refactor: ColorPicker TS Migration

shlomitc:
This PR includes the migration of `ColorPicker` component to Typescript 🫳🎤
Few notes and requests:

1. I've updated the migration guide with some code examples and extras, please go over it as well.
2. The `<ColorPickerContent/>` component is exported as a top-level component of the library, so it might be a good idea to refactor it as a standalone component.
3. I left several `TODO` notes in the code, let's
  3.1 in `ColorPickerColorsGrid.tsx`, there is usage for `useGridKeyboardNavigation`. When I migrated to TS, it started yelling for incorrect properties, so I had to change the implementation there. please take a look and give your opinion on my changes.
  3.2 the `value`, `onValueChange` and `onSave` props - It wasn't clear what should be signature so I decided a value will be the following: `string | string[] | CONTENT_COLORS_VALUES | CONTENT_COLORS_VALUES[];`. Please make sure it's correct


src/utils/colors-utils.ts
```diff
@@ -3,7 +3,11 @@ import { COLOR_STYLES, contentColors } from "./colors-vars-map";
 const ColorUtils = {
   modes: COLOR_STYLES,
   contentColors,
-  getMondayColorAsStyle: (color: string, mode = COLOR_STYLES.REGULAR, withVar = true) => {
+  getMondayColorAsStyle: (
+    color: string,
+    mode: "regular" | "hover" | "selected" = COLOR_STYLES.REGULAR,
```
hadasfa, line 8: if we convert color style to be enum we could just use the enum as the type no? (instead of "regular" | "hover" | "selected")

src/components/ColorPicker/ColorPickerConstants.ts
```diff
@@ -14,4 +14,6 @@ export const DIALOG_WIDTH_PADDING = 24;
 export const COLOR_SHAPES = {
```
hadasfa, line 14: Why not convert color shapes to be regular enum?

src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -1,35 +1,71 @@
...
+  shouldRenderIndicatorWithoutBackground?: boolean;
+  isBlackListMode?: boolean;
+  colorsList?: CONTENT_COLORS_VALUES[] | string[];
+  isMultiselect?: boolean;
+  colorSize?: BASE_SIZES_VALUES;
+  numberOfColorsInLine?: number;
+  focusOnMount?: boolean;
+  colorShape?: COLOR_SHAPES_VALUES;
+  /**
+   * Used to force the component render the colorList prop as is. Usually, this flag should not be used. It's intended only for edge cases.
```
hadasfa, line 31: Thank you!

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
```diff
@@ -1,40 +1,69 @@
...
 import { ColorPickerClearButton } from "./ColorPickerClearButton";
 import { ColorPickerColorsGrid } from "./ColorPickerColorsGrid";
+import { VibeComponentProps, VibeComponent } from "../../../../types";
 
-const ColorPickerContentComponent = forwardRef(
+export interface ColorPickerContentProps extends VibeComponentProps {
+  value: string | string[] | CONTENT_COLORS_VALUES | CONTENT_COLORS_VALUES[]; //TODO - make sure this is correct
+  onValueChange: (value: CONTENT_COLORS_VALUES[] | string[]) => any; //TODO - make sure this is correct
+  colorsList: CONTENT_COLORS_VALUES[] | string[];
+  ColorIndicatorIcon?: ({ size, className }: { size?: string; className?: string }) => JSX.Element;
```
hadasfa, line 22: I think we have a type of icon: SubIcon
There was an issue with this type?

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
```diff
@@ -1,40 +1,69 @@
...
+  SelectedIndicatorIcon?: ({ size, className }: { size?: string; className?: string }) => JSX.Element;
```
hadasfa, line 23: Same icon type comment

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
```diff
@@ -1,40 +1,69 @@
...
+  NoColorIcon?: ({ size, className }: { size?: string; className?: string }) => JSX.Element;
```
hadasfa, line 24: Same icon type comment

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
```diff
@@ -1,40 +1,69 @@
...
+  colorStyle?: "regular" | "selected";
```
hadasfa, line 25: I think it could look better if we used enums here for colorStyle, colorSize, colorShape. WDYT?

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.tsx
```diff
@@ -0,0 +1,112 @@
...
+  colorStyle: COLOR_STYLES_VALUES;
+  shouldRenderIndicatorWithoutBackground: boolean;
+  ColorIndicatorIcon: SubIcon;
+  SelectedIndicatorIcon: SubIcon;
+  isSelected: boolean;
+  colorSize: BASE_SIZES_VALUES;
+  tooltipContent: ElementContent;
+  isActive: boolean;
+  colorShape: COLOR_SHAPES_VALUES;
+  ["data-testid"]?: string;
```
hadasfa, line 32: I think there is no need to add "data-testid" because we declare it in VibeComponentProps

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.tsx
```diff
@@ -0,0 +1,112 @@
...
+import Tooltip from "../../../Tooltip/Tooltip";
+import Clickable from "../../../Clickable/Clickable";
+import { COLOR_SHAPES, COLOR_SHAPES_VALUES } from "../../ColorPickerConstants";
+import { getTestId } from "../../../../tests/test-ids-utils";
+import { ComponentDefaultTestId } from "../../../../tests/constants";
+import { SubIcon, VibeComponent, VibeComponentProps, ElementContent } from "../../../../types";
+import { BASE_SIZES_VALUES } from "../../../../constants";
+
+export interface ColorPickerItemComponentProps extends VibeComponentProps {
+  color: CONTENT_COLORS_VALUES | string;
```
hadasfa, line 21: In case of internal components which are building blocks of our design system exported components, we are not really require to use the VibeComponentProps and VibeComponent types. 
The purpose of these types is to make sure that all our exported components contain the mandatory props (class name, data test id, id). 
By the way, right now these props are not mandatory for some reason (which is bad, not sure why we choose to do this), maybe worth to convert it.
---
In internal props we are not really need to keep the same props and structure as in the exported props so there is no need to use these types (for example this component not support class name prop which should be supported in all external components)

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.tsx
```diff
@@ -0,0 +1,112 @@
...
+      isSelected,
+      colorSize,
+      tooltipContent,
+      isActive,
+      colorShape,
+      "data-testid": dataTestId
+    },
+    _ref
+  ) => {
+    const isMondayColor = useMemo(() => (contentColors as readonly string[]).includes(color), [color]); //TODO - casting hack
```
hadasfa, line 53: If we keep the hack maybe it's worth to replace comment with explanation of why hack was needed?

src/components/ColorPicker/components/ColorPickerContent/ColorPickerClearButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+interface ColorPickerClearButtonProps extends VibeComponentProps {
+  onClick: (e: React.MouseEvent<HTMLButtonElement>) => any;
+  text?: string;
+  Icon: ({ size, className }: { size?: string; className?: string }) => JSX.Element;
+}
+
+export const ColorPickerClearButton: VibeComponent<ColorPickerClearButtonProps> = React.forwardRef(
+  ({ onClick, text, Icon }, ref) => {
+    const onItemClicked = useCallback(() => onClick(null), [onClick]);
```
hadasfa, line 15: what is the purpose of this change?

src/components/ColorPicker/components/ColorPickerContent/ColorPickerColorsGrid.tsx
```diff
@@ -0,0 +1,95 @@
...
+      showColorNameTooltip: showColorNameTooltip
+    },
+    ref
+  ) => {
+    const getItemByIndex = useCallback((index: number) => colorsToRender[index], [colorsToRender]);
+
+    const { activeIndex, onSelectionAction } = useGridKeyboardNavigation({
+      focusOnMount,
+      ref: ref as React.MutableRefObject<HTMLUListElement>,
+      onItemClicked: onColorClicked as any, //TODO - not sure how to utilize this hook correctly
```
hadasfa, line 66: I saw we change a bit the signature of onItemClicked, after the change we still need this casting to any?

TYPESCRIPT_MIGRATION.md
```diff
@@ -7,21 +7,59 @@ While starting migrating components on your own, please pay attention to the fol
 2. Please convert the component tests. There is no need to convert the component stories in the stage (MDX file and js files which are related).
```
hadasfa, line 7: Thank you so much for this update!

TYPESCRIPT_MIGRATION.md
```diff
@@ -7,21 +7,59 @@ While starting migrating components on your own, please pay attention to the fol
...
 
 ### Types and interfaces
-1. If you want to add any general types or interfaces relevant to more than one component during the conversion, please add them to src/types.
-2. Please ensure your component props interface extends VibeComponentProps or another type that extends `VibeComponentProps`. If your component does not support one of the props written in the VibeComponentProps interface, please add it (`id`, `className`, `data-testid`…)
+1. If you want to add any general types or interfaces relevant to more than one component during the conversion, please add them to `src/types`.
```
hadasfa, line 10: Thank you so much for this update!  Much better now

src/components/ColorPicker/components/ColorPickerContent/ColorPickerClearButton.tsx
shlomitc: Will change it to
```tsx
{ 
  onItemClicked: onClick
}
```

src/components/ColorPicker/components/ColorPickerContent/ColorPickerClearButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+      onItemClicked, // hack - we don't really have a grid, it's just for keyboard navigation outside the clear button
+      getItemByIndex: NOOP //hack - we don't really have a grid, it's just for keyboard navigation outside the clear button
+    });
+
+    return (
+      <Button
+        ref={ref}
+        size={Button.sizes.SMALL}
+        kind={Button.kinds.TERTIARY}
+        onClick={() => onSelectionAction(-1)} //hack - we don't really have a grid, it's just for keyboard navigation outside the clear button
```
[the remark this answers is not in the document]
shlomitc: The Button's signature of `onClick` doesn't match to `onSelectionAction`.
Given that, `onSelectionAction` expects the item's index in the Grid but since there's only a single`ClearButton`, it can be anything

src/utils/colors-vars-map.ts
```diff
@@ -154,13 +154,17 @@ export const contentColors = [
   "steel",
   "lilac",
   "pecan"
-];
+] as const;
+
+export type CONTENT_COLORS_VALUES = typeof contentColors[number];
 
 export const COLOR_STYLES = {
   REGULAR: "regular",
```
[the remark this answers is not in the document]
shlomitc: Enum will allow to access only the keys, I want to get the values of the object.

In this case, I want “regular” as a value and not “REGULAR” as a key

src/utils/colors-utils.ts
shlomitc: will change to the proper type

src/components/ColorPicker/ColorPickerConstants.ts
shlomitc: same as for the previous question regarding enum

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
```diff
@@ -1,40 +1,69 @@
...
   GridKeyboardNavigationContext,
   useGridKeyboardNavigationContext
 } from "../../../GridKeyboardNavigationContext/GridKeyboardNavigationContext";
 import { ColorPickerClearButton } from "./ColorPickerClearButton";
 import { ColorPickerColorsGrid } from "./ColorPickerColorsGrid";
+import { VibeComponentProps, VibeComponent } from "../../../../types";
 
-const ColorPickerContentComponent = forwardRef(
+export interface ColorPickerContentProps extends VibeComponentProps {
+  value: string | string[] | CONTENT_COLORS_VALUES | CONTENT_COLORS_VALUES[]; //TODO - make sure this is correct
```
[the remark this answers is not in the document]
shlomitc: extracted to a common type

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
shlomitc: change to SubIcon

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
shlomitc: change to SubIcon

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
shlomitc: change to SubIcon

src/components/ColorPicker/components/ColorPickerContent/ColorPickerContent.tsx
shlomitc: The `COLOR_STYLES` looks like this:
```tsx
export const COLOR_STYLES = {
  REGULAR: "regular",
  HOVER: "hover",
  SELECTED: "selected"
} as const;
```
It contains also the "hover" state which isn't a valid style for the component.
There are a few options:
1. create a common type only for the `ColorPicker` component
2. create a common type in the general types file, but this makes sense only if there are more such cases.
3. keep it as-is since the inner function that uses it `getMondayColorAsStyle` expect to get one for the `COLOR_STYLES` keys, so it is type-safe like this.

I think that option 1 is good enough for now, but let me know if you think it should be something else

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.tsx
shlomitc: I actually think it's a good practice to have the same props convention all across components, even if they are internal.

For example -`className` and `data-testid` are both really important on all levels.
So maybe the `VibeComponent`'s `ref` forwarding isn't necessary, but I'd say that the common props should be always defined

src/components/ColorPicker/components/ColorPickerContent/ColorPickerColorsGrid.tsx
shlomitc: looks good now, remove the `any`

src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.tsx
shlomitc: added a comment

src/components/ColorPicker/components/ColorPickerContent/ColorPickerClearButton.tsx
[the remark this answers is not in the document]
hadasfa: Ok, sound good to me.
If that's the case, we can avoid re-creating anonymous function every time this component is rendered. Let's use useCallback hook for recreating the on click function only when necessary (only once when on this component first render)

src/utils/colors-vars-map.ts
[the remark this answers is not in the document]
hadasfa: I'm not quite sure I understand. If an enum is defined using string values, we can still access those values by their keys. For example, take the following enum:
`enum Color {
  Red = "FF0000",
  Green = "00FF00",
  Blue = "0000FF",
}`

In this case, we can access the string value "FF0000" associated with the key "Red" in the enum by writing Color.Red.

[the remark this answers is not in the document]
shlomitc: you're correct, but when creating a type definition for a prop like this one `color: PossibleColorsStyle`
the `PossibleColorsStyle` must be the values themselves `FF0000` | `00FF00` | `0000FF`.
the enum is in the format of the key-value object which isn't what `color` prop means

src/components/ColorPicker/components/ColorPickerContent/ColorPickerClearButton.tsx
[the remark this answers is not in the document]
shlomitc: not using an anonymous function anymore

hadasfa (approved):

PR #1213, opened 2023-04-16
TooltipProps extend DialogProps

SergeyRoyt:


src/components/Tooltip/Tooltip.tsx
```diff
@@ -4,17 +4,15 @@ import React, { CSSProperties, isValidElement, PureComponent, ReactElement } fro
...
 import { DialogPosition } from "../../constants/positions";
-import VibeComponentProps from "../../types/VibeComponentProps";
 import { TooltipArrowPosition, TooltipTheme } from "./TooltipConstants";
 import { ElementContent } from "../../types/ElementContent";
 import { MoveBy } from "../../types/MoveBy";
 import "./Tooltip.scss";
 
-// TODO TS-migration extend DialogProps, once Dialog is migrated to TS
-export interface TooltipProps extends VibeComponentProps {
+export interface TooltipProps extends Omit<DialogProps, "content"> {
```
hadasfa, line 15: But why we ignore content prop?we support it both for dialog and tooltip?

SergeyRoyt: It's different types

PR #1214, opened 2023-04-17
Css modules comment

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4191061674


src/components/Dialog/DialogContent/DialogContent.tsx
```diff
@@ -71,6 +71,7 @@ export const DialogContent: VibeComponent<DialogContentProps> = React.forwardRef
     }
     return (
       <span
+        // don't remove old classname - override from Monolith
```
shlomitc, line 74: maybe we can add a prefix to such class name that your script won't replace

PR #1220, opened 2023-04-18
MenuItem id

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4328988728

By request: https://monday.slack.com/archives/C02S6F817JQ/p1681833672598009?thread_ts=1681828071.936349&cid=C02S6F817JQ

Feels like an overkill most definitely... not sure if we even need this...

Other solution would be just to omit Id from Menuitem props - and let people use its index combined with menuId

### DIFF
--- a/src/components/Menu/Menu/Menu.tsx
+++ b/src/components/Menu/Menu/Menu.tsx
@@ -76,7 +76,7 @@ const Menu: VibeComponent<MenuProps> & {
     const mergedRef = useMergeRefs({ refs: [ref, forwardedRef] });
 
     const overrideClassName = backwardCompatibilityForProperties([className, classname]);
-    const [activeItemIndex, setActiveItemIndexState] = useState(focusItemIndex);
+    const [activeItemIndex, setActiveItemIndex] = useState(focusItemIndex);
     const [isInitialSelectedState, setIsInitialSelectedState] = useState(false);
 
     const children = useMemo(() => {
@@ -92,9 +92,9 @@ const Menu: VibeComponent<MenuProps> & {
       });
     }, [originalChildren]);
 
-    const setActiveItemIndex = useCallback(
+    const updateActiveItemIndex = useCallback(
       (index: number) => {
-        setActiveItemIndexState(index);
+        setActiveItemIndex(index);
 
         const activeChild = children[index];
         const ariaActiveDescendant = React.isValidElement(activeChild)
@@ -111,19 +111,19 @@ const Menu: VibeComponent<MenuProps> & {
 
     const onSetActiveItemIndexCallback = useCallback(
       (index: number) => {
-        setActiveItemIndex(index);
+        updateActiveItemIndex(index);
         setIsInitialSelectedState(false);
       },
-      [setActiveItemIndex]
+      [updateActiveItemIndex]
     );
 
     useEffect(() => {
       if (focusItemIndexOnMount === -1) {
         return;
       }
-      setActiveItemIndex(focusItemIndexOnMount);
+      updateActiveItemIndex(focusItemIndexOnMount);
       setIsInitialSelectedState(true);
-    }, [focusItemIndexOnMount, setActiveItemIndex, setIsInitialSelectedState]);
+    }, [focusItemIndexOnMount, updateActiveItemIndex, setIsInitialSelectedState]);
 
     const { setSubMenuIsOpenByIndex, hasOpenSubMenu, openSubMenuIndex, setOpenSubMenuIndex, resetOpenSubMenuIndex } =
       useSubMenuIndex();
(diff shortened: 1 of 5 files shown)


src/components/Menu/Menu/Menu.tsx
```diff
@@ -50,7 +50,7 @@ const Menu: VibeComponent<MenuProps> & {
 } = forwardRef(
   (
     {
-      id,
+      id = "menu",
```
hadasfa, line 53: It's not standard to set default id to a react component. It can cause a case of multiple elements with the same id in the page, which can cause a11y issues and JS issues because id should be unique per element

src/components/Menu/Menu/Menu.tsx
```diff
@@ -71,16 +71,48 @@ const Menu: VibeComponent<MenuProps> & {
     forwardedRef
   ) => {
     const ref = useRef<HTMLElement>(null);
+    const mergedRef = useMergeRefs({ refs: [ref, forwardedRef] });
+
     const overrideClassName = backwardCompatibilityForProperties([className, classname]);
-    const [activeItemIndex, setActiveItemIndex] = useState(focusItemIndex);
+    const [activeItemIndex, setActiveItemIndexState] = useState(focusItemIndex);
```
hadasfa, line 77: why? it's not the convention regarding naming state

src/components/Menu/Menu/Menu.tsx
hadasfa: I understand that we need the menu key for accessibility, maybe we can use here third packages for generate id dynamically as uuid? I'm not sure if it's an overkill, let's talk about it in the daily

src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -92,6 +92,7 @@ const MenuItem: VibeComponent<MenuItemProps> & {
       setActiveItemIndex,
       index,
       key,
+      id,
```
hadasfa, line 95: awsome

src/components/Menu/Menu/Menu.tsx
SergeyRoyt: Thought that midleware method, which also changes active-descendant should be called like this
But renamed it now to `updateActiveItemIndex`

src/components/Menu/Menu/Menu.tsx
SergeyRoyt: dynamic uuid added
haven't implemented warnings - I think it's too noisy and not really required here

src/components/Menu/Menu/Menu.tsx
```diff
@@ -70,27 +70,15 @@ const Menu: VibeComponent<MenuProps> & {
     },
     forwardedRef
   ) => {
+    const overrideId = useMemo(() => id || `menu-${crypto.randomUUID()}`, [id]);
```
SergeyRoyt, line 73: Not sure about this one `crypto.randomUUID()`, mabe to use uuid package? Or better to use some custom function not to increase the bundle size?

SergeyRoyt:
Removed id generator function, cause of snapshots issues with it - each time it's new id, which fails test. 
Can use 2 solution here:
1) Mock Menu component id, but not only in our package, but in Monolith and in Tridents as well - which is complicated enough
2) Just remove default menu id. If somebody will need they can just pass it there

src/components/Menu/Menu/Menu.tsx
```diff
@@ -104,6 +90,39 @@ const Menu: VibeComponent<MenuProps> & {
...
 
+    const updateActiveItemIndex = useCallback(
+      (index: number) => {
+        setActiveItemIndex(index);
+
+        const activeChild = children[index];
+        const ariaActiveDescendant = React.isValidElement(activeChild)
+          ? activeChild?.props?.id || `${id}-${index}`
+          : undefined;
+        if (ariaActiveDescendant) {
```
hadasfa, line 101: As we talked before: let's add console warning if developer did not define menu id

hadasfa: Also let's add a test that make sure aria-activedescendant added when user using focus and arrows? (snapshot tests checking component state while it's not focused)

src/components/Menu/Menu/Menu.tsx
```diff
@@ -104,6 +90,45 @@ const Menu: VibeComponent<MenuProps> & {
...
+      }
+    }, [id]);
+
+    const updateActiveItemIndex = useCallback(
+      (index: number) => {
+        setActiveItemIndex(index);
+
+        const activeChild = children[index];
+        const ariaActiveDescendant = React.isValidElement(activeChild)
+          ? activeChild?.props?.id || `${id}-${index}`
```
hadasfa, line 105: The only problematic thing here is that we define the default id of menu item twice: here and in the menu item component (`${id}-${index}`), so I think it will be the best to set the menu item default id as the default value (and then we will be better to remove `${id}-${index}` from here).

Because this is the only comment I have, I am approving the pr for not be a blocker.

hadasfa (approved):
Approve with one comment

PR #1230, opened 2023-04-20
fix for Vite [prerelease]

orrgottlieb:


src/components/Dropdown/Dropdown.jsx
```diff
@@ -297,8 +295,7 @@ const Dropdown = ({
         ...(multi && {
           MultiValue: NOOP, // We need it for react-select to behave nice with "multi"
           ValueContainer: MultiValueContainer
-        }),
-        ...(isVirtualized && { MenuList: WindowedMenuList })
```
SergeyRoyt, line 301: Why we got rid off Virtulization support?

src/components/Dropdown/__tests__/__snapshots__/dropdown-snapshot-tests.jest.js.snap
SergeyRoyt, line 1: Is that ok that so many lines were changed there?
Is it cause of virtualization removal...?

PR #1240, opened 2023-04-30
Css modules/sergeyro/attention box warning color + icons color + heading style fix

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4332740500

### DIFF
--- a/src/components/AttentionBox/AttentionBox.scss
+++ b/src/components/AttentionBox/AttentionBox.scss
@@ -13,7 +13,6 @@
 
   &__title {
     margin: 0;
-    font: var(--font-h5);
     font-weight: 700;
   }
 
--- a/src/components/AttentionBox/AttentionBox.scss
+++ b/src/components/AttentionBox/AttentionBox.scss
@@ -13,6 +13,8 @@
 
   &__title {
     margin: 0;
+    font: var(--font-h5);
+    font-weight: 700;
   }
 
   &_text--compact {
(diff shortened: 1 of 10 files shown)


src/components/AttentionBox/AttentionBox.scss
```diff
@@ -13,6 +13,8 @@
 
   &__title {
     margin: 0;
+    font: var(--font-h5);
```
hadasfa, line 16: why? Attention box using heading for heading component so this override is weird. changing the heading type was not enough?

src/components/AttentionBox/AttentionBox.scss
```diff
@@ -28,43 +30,24 @@
...
       @include theme-prop(color, primary-text-color);
+    }
 
-      &__title-container {
-        &__icon {
-          @include theme-prop(color, primary-text-color);
-        }
-      }
+    &-warning {
+      @include theme-prop(background-color, warning-color-selected);
```
hadasfa, line 49: not a blocker but next time there is no need to use  theme-prop

hadasfa: we can just use var and it will do the same thing

src/components/AttentionBox/AttentionBox.scss
hadasfa: After check with yael: we need to fix the weight of h5 heading to 700. Regarding the font it should work without override, what is the problem?

hadasfa (approved):
Approved with comments

src/components/AttentionBox/AttentionBox.scss
SergeyRoyt: ok, don't know why we are using theme-prop in a first place)

src/components/AttentionBox/AttentionBox.scss
SergeyRoyt: no need in that - agree

PR #1252, opened 2023-05-04
feat: <Dropdown> - ref forwarding

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4408947808

### DIFF
--- a/src/components/Dropdown/__tests__/driver.js
+++ b/src/components/Dropdown/__tests__/driver.js
@@ -38,10 +38,6 @@ export default class DropdownDriver {
     return this.renderResult.asFragment();
   }
 
-  get dropdownWrapper() {
-    return this.renderResult.container.querySelector("#dropdown-menu-id");
-  }
-
   get dropdownMenu() {
     return this.renderResult.container.querySelector(".menu");
   }


shlomitc (commented):
I suggest adding a test, just to make sure the functionality works and its another way to explain how to use it

src/components/Dropdown/__stories__/dropdown.stories.scss
```diff
@@ -37,4 +37,8 @@
     align-items: center;
     justify-content: center;
   }
+
+  &_button {
```
shlomitc, line 41: My two cents about our code examples - they should be copy-pastable into the user's code.
In that case, className is not something that can be easily copied.
I'd suggest using `Flex`'s `gap` and maybe `Box` if it's something that's really necessary.
We can also wrap with our internal example presentation components that know how to arrange elements as a proper example. Then, when the user will copy the example, he can simply delete it.

src/components/Dropdown/__tests__/driver.js
```diff
@@ -38,6 +38,10 @@ export default class DropdownDriver {
     return this.renderResult.asFragment();
   }
 
+  get dropdownWrapper() {
```
shlomitc, line 41: I'm not sure this is in use in your current PR changes, so I wonder if you need it

src/components/Dropdown/__stories__/dropdown.stories.scss
SergeyRoyt: Yeah, but this is not really an example we expect user to copy

src/components/Dropdown/__tests__/driver.js
SergeyRoyt: Removing it - eventually not needed it

shlomitc (approved):

PR #1256, opened 2023-05-07
chore: fix typography docs to map to the exact typography css vars

shlomitc:
Following the definition Figma: https://www.figma.com/file/UcjPooakzpU4lZyKCD2Pim/%F0%9F%A6%8B-New-typography-styles-%F0%9F%A6%8B?type=design&node-id=1-33&t=Ub5dYimpeTjCUNWs-4

1. map directly to the CSS vars
2. change titles a bit
3. remove the story CSS overrides to align the typography example to the description. The example might look bad, but this is the actual code we have and it should be resolved

</div>
Before:
image

After:
image

### DIFF
--- a/src/storybook/stand-alone-documentaion/typography/text-styles/text-styles.jsx
+++ b/src/storybook/stand-alone-documentaion/typography/text-styles/text-styles.jsx
@@ -61,9 +61,7 @@ export const TextStyles = () => {
         description="Use for general text or labels"
         className={CSS_BASE_CLASS}
       >
-        <span style={{ font: "var(--font-general-label)" }} className={cx(bemHelper({ element: "visual-element" }))}>
-          text
-        </span>
+        <span style={{ font: "var(--font-general-label)" }}>text</span>
       </VisualDescription>
       <VisualDescription
         ariaLabel="paragraph"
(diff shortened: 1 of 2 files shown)


src/storybook/stand-alone-documentaion/typography/text-styles/text-styles.jsx
```diff
@@ -13,137 +12,83 @@ export const TextStyles = () => {
       <VisualDescription
         className={CSS_BASE_CLASS}
         ariaLabel="H1"
-        title="Main heading (Poppins 30px bold)"
```
SergeyRoyt, line 16: Are we sure we don't need this "(Poppins 30px bold)"? I think this information might be relevant for designers - but if Meirav or Yael think otherwise, I'm good

src/storybook/stand-alone-documentaion/typography/text-styles/text-styles.jsx
```diff
@@ -13,137 +12,83 @@ export const TextStyles = () => {
...
+        code="font: var(--font-h1)"
```
SergeyRoyt, line 15: nice!

SergeyRoyt (approved):

PR #1283, opened 2023-05-17
Feature/sergeyro/button brand color

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4455194421

V1 PR - https://github.com/mondaycom/monday-ui-react-core/pull/1285


src/components/Button/Button.module.scss
```diff
@@ -210,6 +220,12 @@ $disabled-on-primary-color-opacity: 0.5;
...
 
+.kindSecondary.colorBrandActive,
+.kindSecondary.colorBrandActive:hover {
+  background-color: var(--brand-selected-color);
+  border-color: var(--brand-color);
+}
+
 .kindSecondary.colorPrimary:hover:not(.colorPrimaryActive),
 .kindSecondary.colorPrimary:focus:not(.colorPrimaryActive) {
   @include theme-prop(background, primary-background-hover-color);
```
SergeyRoyt, line 231: Not sure if it's needed - check in V1

shlomitc (commented):
please also add a story example for this.
in its description, please state that this is still an experimental feature and can be changed, so it should be used only if been instructed to

PR #1285, opened 2023-05-17
V1/feature/sergeyro/button brand colors

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4455194421

V2 PR - https://github.com/mondaycom/monday-ui-react-core/pull/1283


shlomitc (commented):
please also add a story example for this.
in its description, please state that this is still an experimental feature and can be changed, so it should be used only if been instructed to

PR #1294, opened 2023-05-21
change Dialog click outside event to be mousedown

doronbrikman:
Related to that issue: https://monday.monday.com/boards/3957717237/views/91932741/pulses/2521139501?userId=39129653
When a Dialog is opened and the user is right clicking to open the context menu the "click outside" doesn't happen.
In the depulse monorepo there's a 3rd party package `react-onclickoutside` that does it right for the other use cases, and the default event are like here:
https://github.com/Pomax/react-onclickoutside/blob/2cad822a8ca75319c662c83cd42bad548cc06e9d/src/index.js#LL42C21-L42C30

#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


src/hooks/useClickOutside/index.ts
```diff
@@ -24,7 +24,7 @@ export default function useOnClickOutside({
   const documentRef = useRef(document.body);
 
   useEventListener({
-    eventName: "click",
+    eventName: "mousedown",
```
orrgottlieb, line 27: @doronbrikman thanks for the PR, can you add this as a prop, like - `eventName` or `eventStage` or something that doesn't change the behaviour for all usecases

shlomitc: @orrgottlieb - any reason why not to do this in all cases?

doronbrikman: I thought maybe we would like it to be the default value, do you want it to be `click` as default?
Also to which level we would like to pass this decision? Dialog level? or the usages of Dialog, like StatusPicker?, I would say Dialog

shlomitc: My initial thought says that it should be mouseup for all, but if we are not sure about it, then I agree with Orr to provide this as an additional configuration and not change the original behavior for now

orrgottlieb: `click` unlike `mouseup`/`mousedown` should happen on the same element. (it also should include the pointer events as well)

IMO - this indicates an intent to click outside of the element.

We have an issue in our monolith where a 3rd party is using the `mousedown` event, which doesn't indicate the intention of the user.

Because of that i thought of adding a configuration - `eventName` with the subset of `mousedown` `mouseup` `click`  - and add the relevant pointer events

doronbrikman: WDYT about that? @orrgottlieb 

> Also to which level we would like to pass this decision? Dialog level? or the usages of Dialog, like StatusPicker?, I would say Dialog

shlomitc: @orrgottlieb - how should this be tested in the library to make sure there won't be regressions?

hadasfa: As far as I understand, the proposed change still presents the same issue. We are modifying our component's behavior based on choices made by third parties, which may not necessarily be the 'correct' approach.
Why not just adding another closing trigger to the dialog onMouseUpOutside / onMouseDownOutside? @shlomitc @orrgottlieb

hadasfa: Regarding tests: 
For the useOnClickOutside hook, I would suggest adding two tests that verify if the correct events are triggered for different event names.

As for the Dialog component, it depends on the chosen implementation approach. It might be easier to add interaction-based tests using Storybook to test the dialog's behavior (test that the dialog with "close by click outside ability" actually closing after click outside)

Alternatively, we can consider implementing regular Jest tests for the Dialog component, checking if both click and mouseup events cause the dialog disappear. However, if this proves to be challenging, utilizing Storybook for interaction testing might be simpler and good enough.

shlomitc: Regarding the test: so sounds like the storybook interaction testing is better here. is there an example to get inspired by?

Regarding your suggestion - if the solution in this PR works without any regression - I think it will be better than exposing another configuration

hadasfa: Tests: there are some example in our codebase in combobox component and useActiveDescendantListFocus
API: let's talk about it in a meeting?

src/components/Dialog/__tests__/Dialog.jest.js
```diff
@@ -0,0 +1,23 @@
...
+import Dialog from "../Dialog";
+
+describe("Dialog tests", () => {
+  describe("Click outside", () => {
+    it("should call onClickOutside and close dialog", async () => {
+      const onClickOutsideMock = jest.fn();
+      const element = document.createElement("div");
+      document.body.appendChild(element);
+
+      render(<Dialog onClickOutside={onClickOutsideMock} content={<div>Dialog</div>} open />, { baseElement: element });
```
hadasfa, line 13: this test only check onclickoutside function it's not test if the component hide the dialog when user click outside :/

src/hooks/__tests__/useClickOutside.jest.ts
```diff
@@ -48,4 +48,27 @@ describe("useClickOutside", () => {
       return expect(callbackStub.mock.calls.length).toEqual(0);
     });
   });
+
+  describe("contextMenu - using mousedown", () => {
+    it("should not call the callback when click outside the element", () => {
+      act(() => {
+        fireEvent.mouseDown(document.body);
```
hadasfa, line 55: this test the check for click use case so:
```suggestion
        fireEvent.click(document.body);
```

src/components/Dialog/__tests__/Dialog.jest.js
doronbrikman: I check both things:
expect(onClickOutsideMock).toBeCalled();
expect(screen.findByText("Dialog")).not.toBeInTheDocument;

src/components/Dialog/__tests__/Dialog.jest.js
```diff
@@ -0,0 +1,23 @@
...
+
+      await screen.findByText("Dialog");
+
+      userEvent.click(document.body);
+
+      expect(onClickOutsideMock).toBeCalled();
+      expect(screen.findByText("Dialog")).not.toBeInTheDocument;
```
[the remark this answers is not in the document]
doronbrikman: It's check it's not shown `.not.toBeInTheDocument`

src/hooks/__tests__/useClickOutside.jest.ts
doronbrikman: `click` will work here because it's the default value, but `mouseDown` won't so Im checking this one. next test will check the same but with the new `eventName` passing

shlomitc:
This PR has been stale for a while so I'm closing it. Thanks

PR #1303, opened 2023-05-22
Add support on close button theme for tipseen (V2)

hadasfa:
There is a special reason to add it also to v1? Or it's ok to add it only to V2?

### DIFF
--- a/src/components/Tipseen/__stories__/tipseen.stories.mdx
+++ b/src/components/Tipseen/__stories__/tipseen.stories.mdx
@@ -154,7 +154,7 @@ Use when the tip is too long or when you want to teach something in steps.
           // Therefore, there is no need to move this prop in your implementations.
           position={Tipseen.positions.RIGHT}
           modifiers={modifiers}
-          isCloseButtonOnImage
+          closeButtonTheme={Tipseen.closeButtonThemes.DARK}
           content={
             <>
               <TipseenImage className="monday-storybook-tipseen_image" src={picture} />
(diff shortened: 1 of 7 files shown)


src/components/Tipseen/__stories__/tipseen.stories.mdx
```diff
@@ -19,7 +19,7 @@ import "./tipseen.stories.scss";
 
 export const metaSettings = createStoryMetaSettings({
   component: Tipseen,
-  enumPropNamesArray: ["position", "animationType", "justify"]
+  enumPropNamesArray: ["position", "animationType", "justify", "closeButtonTheme"]
```
hadasfa, line 22: For display enum keys instead of values in the props section in storybook

src/storybook/functions/createStoryMetaSettings.js
```diff
@@ -14,6 +14,14 @@ const allowedIcons = iconsMetaData.reduce(
   { options: [], mapping: {} }
 );
 
+function cleanTSEnums(enumObj) {
```
hadasfa, line 17: Until now all enums had string values because backward compatibility, this is the first enum which does not have this issue. Do we still want to define string values for each ts enum key? or we are prefer to use number values?
If we prefer using number values we need to remove them for the options object because TS enums with number values contains both keys and values as keys.

hadasfa: If we prefer to continue using string values I will revert this change

shlomitc (approved):
just worth improving the example by using the dark skin close button
image

PR #1308, opened 2023-05-22
Add tokens.css in Vibe 2.0

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4494861693

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "monday-ui-react-core",
-  "version": "2.1.0",
+  "version": "2.2.0",
   "description": "Official monday.com UI resources for application development in React.js",
   "main": "./dist/main.js",
   "types": "./dist/types.d.ts",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
       "default": "./dist/main.js"
     },
     "./dist/": "./dist/",
+    "./tokens": "./dist/esm/src/tokens.scss.js",
```
shlomitc, line 14: This will create a js entry which is interesting  approach.
There might be an issue in terms of css loading order in case this will be loaded after other files were already loaded.

If we want to do the same as today - we actually need to create a css output “bundle”

SergeyRoyt: Won't it inject styles at the top as we are using `style-inject`?

shlomitc: that's how it works, but I'm not sure if the order of loading is guaranteed, just need to check it

PR #1314, opened 2023-05-24
V1 - feat: <Button> - add loaderWrapperClassName prop

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4526538976

V2 - https://github.com/mondaycom/monday-ui-react-core/pull/1315

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -23,6 +23,8 @@ export interface ButtonProps {
   children?: React.ReactNode;
   /** Custom class names to pass to the component */
   className?: string;
+  /** Custom class name to override Loader wrapper, when loading prop equals to true */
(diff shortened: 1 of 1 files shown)


src/components/Button/Button.tsx
```diff
@@ -314,7 +317,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
     if (loading) {
       return (
         <button {...buttonProps}>
-          <span className="monday-style-button__loader">
+          <span className={cx("monday-style-button__loader", loaderWrapperClassName)}>
```
hadasfa, line 320: why? it's not a good practice

SergeyRoyt:
Not needed eventually - override is gonna be removed 👍

PR #1323, opened 2023-05-25
Add tokens.css in Vibe 2.0

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4494861693

Monolith PR - https://github.com/DaPulse/dapulse/pull/56798


shlomitc:
@orrgottlieb - WDYT? providing tokens as a css-in-js mode instead of importing directly from scss files (and eventually in the CSS bundle of the consuming app)?

orrgottlieb:
@shlomitc , all of our components are looking at the CSS variables. so you would need to set them accordingly in the main of your application ? or i didn't understand what you are saying

shlomitc:
@orrgottlieb - that's correct.
The suggested PR here exposes a new entry point for tokens, however, they are different from the conventional way of importing a CSS file from SCSS file. In this case, you will need to import the tokens from your main app js file

orrgottlieb:
OK, in our main use-case we need to have the ability to import the tokens once - all of our micro frontend are on the same window object and importing more than once might have collisions or have multiple declarations of the same file.

More than that - we need the ability to have one source of truth that will change all of our css variables in an outer scope which should effect all MF's and the main application

shlomitc:
@orrgottlieb - I'll try to explain myself better.
The question here is about how the import will be and how will they affect

With this solution, this is how importing from `App.js` will look like:
```tsx
import 'monday-ui-react-core/tokes;
```

in oppose to what could have been if there had been a dedicated CSS bundle:
```tsx
import "monday-ui-react-core/dist/tokens.css"
```

The first is using Vibe's CSS injection mechanism, while the latter will probably end up in the CSS of the consuming app (and extracted to a static file)

orrgottlieb:
@shlomitc Ah, ok thanks.
I'd also go with the Injection solution. But would map `"monday-ui-react-core/dist/tokens.css"` to `import "monday-ui-react-core/tokes"` so the import would be prettier and consistent

shlomitc:
There's actually no `monday-ui-react-core/dist/tokens.css` at all right now as far as I understand
This might require changing webpack/rollup configuration to do so.

SergeyRoyt:
There is `monday-ui-react-core/dist/tokens.css` - it's getting copied from `monday-ui-style` css
image

shlomitc (approved):
looks good, just make changes to the readme file as well

README.md
```diff
@@ -1,70 +1,91 @@
...
 ```javascript
 import Button from "monday-ui-react-core/dist/Button";
 ```
 
-You also need to import Main CSS file to your project
+Then you also need to import Main CSS file to your project
+
 ```javascript 
 import "monday-ui-react-core/dist/main.css"
 ```
```
SergeyRoyt, line 47: Please check these changes @shlomitc

README.md
```diff
@@ -1,70 +1,91 @@
...
+
 ```
 $ npm install monday-ui-react-core
 ```
 
 ## Usage
+
+### ESM imports (recommended)
+
 You can import from the main bundle:
```
shlomitc, line 23:
```suggestion
Components are imported from the library's root entry:
```

README.md
```diff
@@ -1,70 +1,91 @@
...
+
 ```javascript
 import { Button } from "monday-ui-react-core";
 ```
-or you might want to import directly the component from the dist folder
+
+You also need to import CSS tokens file to your project
```
shlomitc, line 29:
```suggestion
In order for all fonts and colors to be loaded, you should import the CSS tokens at the root of your application file
```

README.md
```diff
@@ -1,70 +1,91 @@
...
+
+```javascript
+import "monday-ui-react-core/tokens";
+```
+
+### CommonJS imports (will be deprecated soon)
```
shlomitc, line 35: I think this should all be removed to a different markdown file and not be part of the main README since we don't want people to use it.

Under the "Usage" section write something like "If your project is importing files differently - read more here" and link to that page.

shlomitc (changes requested):

shlomitc (approved):

PR #1333, opened 2023-05-31
fix: <Tipseen> increase css specificity

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4547079857


shlomitc:
Can you explain why does it happen only in the monolith and how did you fix it?

PR #1334, opened 2023-05-31
feat: <Modal> custom header title

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4556458677

Chromatic: https://62037c72e0e4d6004a9a450f-yiocueayfe.chromatic.com/?path=/docs/feedback-modal--modal-with-editable-title
V1 - https://github.com/mondaycom/monday-ui-react-core/pull/1335

### DIFF
--- a/src/components/Modal/ModalHeader/ModalHeader.module.scss
+++ b/src/components/Modal/ModalHeader/ModalHeader.module.scss
@@ -12,7 +12,6 @@
     flex-grow: 1;
     margin-block-start: var(--spacing-large);
     margin-block-end: 0;
-
   }
 
   .icon {
@@ -29,6 +28,6 @@
   .closeButton {
     position: absolute;
     right: 20px;
-    top: 20px;
+    top: 26px;
   }
 }
--- a/src/components/Modal/ModalHeader/ModalHeader.module.scss
+++ b/src/components/Modal/ModalHeader/ModalHeader.module.scss
@@ -28,6 +28,6 @@
   .closeButton {
     position: absolute;
     right: 20px;
-    top: 26px;
+    top: 20px;
   }
 }
--- a/src/components/Modal/ModalHeader/ModalHeader.module.scss
+++ b/src/components/Modal/ModalHeader/ModalHeader.module.scss
@@ -1,3 +1,5 @@
+@import "../../../styles/typography";
+
 .container {
   display: flex;
   align-items: flex-start;
@@ -5,9 +7,7 @@
   margin-block-end: var(--spacing-large);
 
   .title {
-    font-family: var(--title-font-family);
-    font-size: var(--font-size-h2);
-    font-weight: var(--font-weight-bold);
+    @include font-title();
     color: var(--primary-text-color);
     flex-grow: 1;
     margin-block-start: var(--spacing-large);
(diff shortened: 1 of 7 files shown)


src/components/Modal/ModalHeader/ModalHeader.tsx
```diff
@@ -10,9 +10,9 @@ import styles from "./ModalHeader.module.scss";
 
 export interface ModalHeaderProps extends VibeComponentProps {
   /**
-   * Title of the modal
+   * Title of the modal - using string is a recommended standard
    */
-  title: string;
+  title: ElementContent;
```
shlomitc, line 15: let's document the `EditableHeading` example

src/components/Modal/ModalHeader/ModalHeader.tsx
```diff
@@ -10,9 +10,9 @@ import styles from "./ModalHeader.module.scss";
 
 export interface ModalHeaderProps extends VibeComponentProps {
   /**
-   * Title of the modal
+   * Title of the modal - using string is a recommended standard
```
shlomitc, line 13: worth explaining why it's recommended.
It's recommended because it will provide the well-defined styles of a title.
Using it by passing other content should be according to design guidlines

src/components/Modal/ModalHeader/ModalHeader.tsx
SergeyRoyt: Not sure we want to encourage people to use it more by promoting in the story. I think if there is a need, people will figure it out, if no, they won't even notice - we also used to have this approach in such cases before

shlomitc: but if it's a good use case - why not show it? then it also help us to see (and test) the feature

src/components/Modal/ModalHeader/ModalHeader.tsx
SergeyRoyt: Expanded the description here

src/components/Modal/ModalHeader/ModalHeader.module.scss
```diff
@@ -29,6 +28,6 @@
   .closeButton {
     position: absolute;
     right: 20px;
-    top: 20px;
+    top: 26px;
```
SergeyRoyt, line 31: So it would be aligned to the center of the title

SergeyRoyt: Reverted

shlomitc (changes requested):
1. I'd expect its total width to be longer and that the text input would reach the end
image

2. This is a bug
image

image

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -378,6 +378,65 @@ clicking outside. Use this variant in case of sensitive or important messages, a
...
+              title={
+                <EditableHeading
+                  size={EditableHeading.sizes.SMALL}
+                  placeholder={"Modal title"}
+                  className="monday-storybook-modal-editable-heading-example"
+                />
+              }
+            />
+            <ModalContent>
+              <p className="modal-content">Modal content goes here</p>
```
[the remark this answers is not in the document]
SergeyRoyt: removed

shlomitc (approved):
Perfect 👏

PR #1337, opened 2023-05-31
fix: <Tipseen> dependencies array

SergeyRoyt:


src/components/Tipseen/Tipseen.tsx
```diff
@@ -146,7 +146,7 @@ const Tipseen: VibeComponent<TipseenProps> & {
           <div className={cx(styles.tipseenContent)}>{content}</div>
         </div>
       ),
-      [content, onClose, overrideCloseAriaLabel, overrideHideCloseButton, title, titleClassName]
+      [closeButtonTheme, content, onClose, overrideCloseAriaLabel, overrideHideCloseButton, title, titleClassName]
```
shlomitc, line 149: The fix is ok, but is this `useMemo` really needed?
I don't think there's so much rendering happening in this component so maybe it would just be simpler to render it regularly as node instead of a function

SergeyRoyt: @hadasfa it was written by you, so what do you think?

hadasfa: I think it is ok to remove,  I don't remember why I added it.
Just worth checking it's not causing weird rerendering issues (but I think it will be ok)

src/components/Tipseen/__stories__/tipseen.stories.mdx
```diff
@@ -115,19 +116,23 @@ Use when the tip is too long or when you want to teach something in steps.
...
+        <div>Popover message №1 will appear here</div>,
+        <div>Popover message №2 will appear here</div>,
+        <div>Popover message №3 will appear here</div>,
+        <div>Popover message №4 will appear here</div>,
+        <div>Popover message №5 will appear here</div>
       ];
+      const [activeStepIndex, setActiveStepIndex] = useState(2);
+      const onChangeActiveStep = useCallback((_e, stepIndex) => {
+        setActiveStepIndex(stepIndex);
+      }, []);
```
SergeyRoyt, line 128: Added clickability to wizzard example

shlomitc (approved):

PR #1338, opened 2023-06-01
feat: <StepIndicator> font-weight redesign + small padding fixes

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4569744883

V1 - https://github.com/mondaycom/monday-ui-react-core/pull/1339


shlomitc:
is this how it should look like?
image

SergeyRoyt:
@shlomitc This is design approved by Yael and we also we have an override in the Monolith with exactly same font-weight. Let me double check though

shlomitc (approved):

PR #1342, opened 2023-06-04
Support dialog context menu hide trigger

hadasfa:
Add new hide trigger to the dialog for closing when right click (context menu native event) outside of the dialog component.
This pr also contains tests for the new feature, data-testid prop for some components which we used for implementing the dialog tests) and some storybook interactions types fixes


src/components/Dialog/__tests__/Dialog.interactions.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export const closeTriggersInteractionSuite = interactionSuite({
+  tests: [isDialogHiddenAfterClickOutside, isDialogHiddenAfterContextMenu],
+  beforeAll: async canvas => {
+    // From some reason we have an issue with rendering the dialogs according to the container selector in the initial mount, after clicking
+    // all dialog render in the right placements
+    const clickOutsideButton = await getByTestId(canvas, CLICK_OUTSIDE_DIALOG_BUTTON);
+    await userEvent.click(clickOutsideButton);
+    // wait for hide default delay
+    await new Promise(resolve => setTimeout(resolve, 100));
```
hadasfa, line 28: We can also use the hide delay and show delay props in the story but the stories are too big anyway.
I guess you may think it is a signal for that we need to separate between the stories and the tests in the future, there is a chance that is the case, but also the dialog api is really complex compare to other components in Vibe.

src/constants/dialog.ts
```diff
@@ -15,7 +15,8 @@ export enum HideShowEvent {
   MOUSE_DOWN = "mousedown",
   FOCUS = "focus",
   BLUR = "blur",
-  CONTENT_CLICK = "onContentClick"
+  CONTENT_CLICK = "onContentClick",
+  CONTEXT_MENU = "contextmenu"
```
shlomitc, line 19: isn't this const duplicated?

src/components/Flex/Flex.tsx
```diff
@@ -47,7 +48,8 @@ const Flex: VibeComponent<FlexProps> & {
       style,
       ariaLabelledby,
       ariaLabel,
-      tabIndex
+      tabIndex,
+      "data-testid": dataTestId = getTestId(ComponentDefaultTestId.FLEX, id)
```
shlomitc, line 52: I don't think this is a relevant change for this PR and in general, I don't think this component should have a test id since it's not functional, only UI

src/tests/interactions-utils.ts
```diff
@@ -116,19 +116,19 @@ export const interactionSuite =
     }
   };
 
-export const getByTestId = (rootElement: HTMLElement, dataTestId: string) => {
+export const getByTestId = (rootElement: HTMLElement | BoundFunctions<typeof queries>, dataTestId: string) => {
```
shlomitc, line 119: what's `BoundFunctions` and why do you need it?

src/components/DialogContentContainer/__tests__/dialogContentContainer.jest.tsx
```diff
@@ -6,3 +6,8 @@ it("renders correctly with empty props", () => {
   const tree = renderer.create(<DialogContentContainer />).toJSON();
   expect(tree).toMatchSnapshot();
 });
+
+it("renders correctly with data-testid", () => {
```
shlomitc, line 10: let's add a backlog task to automate this test, kind of a sanity test for all components

src/components/Dialog/__tests__/Dialog.interactions.ts
shlomitc, line 28: instead of statically waiting - I suggest a retry mechanism that waits until the expected result happens. in this case - the modal should open

Looks at something like this or use a library that does it
https://learnersbucket.com/examples/interview/retry-promises-n-number-of-times-in-javascript/
https://github.com/wix-incubator/wix-eventually

src/components/Dialog/__tests__/Dialog.interactions.ts
```diff
@@ -0,0 +1,59 @@
...
+import { queries, userEvent, fireEvent, waitFor } from "@storybook/testing-library";
+
+const isDialogHiddenAfterClickOutside = createTestIfDialogHiddenAfterTrigger(CLICK_OUTSIDE_DIALOG, () =>
+  userEvent.click(getDialogContainer())
+);
+const isDialogHiddenAfterContextMenu = createTestIfDialogHiddenAfterTrigger(CONTEXT_MENU_DIALOG, () =>
+  fireEvent.contextMenu(getDialogContainer())
+);
+
+export const closeTriggersInteractionSuite = interactionSuite({
```
shlomitc, line 20: in general, the test boilerplate looks really complex, so I think we should think about how to make it more simple

shlomitc (changes requested):

src/components/Dialog/DialogContent/DialogContent.tsx
```diff
@@ -38,6 +38,7 @@ export interface DialogContentProps extends VibeComponentProps {
   disableOnClickOutside?: boolean; // TODO prop is passsed, but not used. How it should behave?
   containerSelector?: string;
   disableContainerScroll?: boolean | string;
+  onContextMenu?: (e: React.MouseEvent) => void;
```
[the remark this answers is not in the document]
hadasfa: context menu is an event which is meaning is right click (it's an actual event like click or mousedown)/
I can add comment here and on the PR description

src/components/Dialog/DialogContent/DialogContent.tsx
```diff
@@ -85,7 +96,7 @@ export const DialogContent: VibeComponent<DialogContentProps> = React.forwardRef
           enableScroll();
         }
       }
-    }, [disableContainerScroll, disableScroll, enableScroll, isOpen]);
+    }, [isOpen]);
```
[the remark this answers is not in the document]
hadasfa: by mistake will fix

src/components/Dialog/__stories__/Dialog.stories.mdx
```diff
@@ -349,17 +362,19 @@ We can set the triggers which will be responsible for hide the dialog
           <Dialog
             modifiers={modifiers}
             shouldShowOnMount
+            containerSelector={`#${HIDE_TRIGGERS_CONTAINER}`} // for testing purpose
```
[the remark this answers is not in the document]
hadasfa: I think it problematic to add tests category to our main dialog doc (can be confusing). What about adding new Tests story which will contain tests stories for all relecant components (right now only dialog)

src/components/Dialog/__tests__/DialogDataTestIds.jsx
```diff
@@ -0,0 +1,4 @@
+export const CLICK_OUTSIDE_DIALOG = "click-outside-dialog";
```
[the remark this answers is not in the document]
hadasfa: 1. ok
2. it's for the story and for the interaction tests

src/components/Flex/Flex.tsx
hadasfa: This change added to the pr because I needed to click on the container which was a flex component. I think layout components can use data-testsid even if they are pure UI.
Also, our current guidelines are that every Vibe component should have data test id

src/tests/interactions-utils.ts
hadasfa: BoundFunctions is the type of the Canvas param in the storybook interaction tests, this is a fix for being able to use the function in these files in TS code

src/components/Dialog/__tests__/Dialog.interactions.ts
hadasfa: Agree, adding a backlog task / reaserch task

src/components/Dialog/DialogContent/DialogContent.tsx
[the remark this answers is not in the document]
hadasfa: By the way this component is internal and none will be expose to this comment

src/components/Dialog/__tests__/Dialog.interactions.ts
hadasfa: https://monday.monday.com/boards/3532714909/pulses/4587109938

src/components/Dialog/__tests__/Dialog.interactions.ts
hadasfa: Used storybook waitFor ability which provide similar mechanism

src/components/Dialog/__stories__/Dialog.stories.mdx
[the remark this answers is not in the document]
shlomitc: I didn't see yet how jest works with those test stories but I think the best practice should be separating components instead of creating one big file in terms of rerunning specific tests on failure and reporting.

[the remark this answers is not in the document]
shlomitc: Anyway, my second option would be to have a separate storybook just for testing, but I think that would take longer.
Regarding exposing a "Test" category - I don't see an issue about it

src/components/Flex/Flex.tsx
shlomitc: so I'm fine with allowing to pass a data-testid, but I don't think there should be a default one

src/components/Dialog/__stories__/Dialog.stories.mdx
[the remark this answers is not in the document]
hadasfa: let's talk about it in a team level then?

shlomitc (approved):

PR #1343, opened 2023-06-04
fix: <Dropdown> multi on option delete callback

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4571615283

### DIFF
--- a/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.jsx
+++ b/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.jsx
@@ -37,6 +37,14 @@ export default function Container({ children, selectProps, ...otherProps }) {
     selectedOptionsCount: selectedOptions.length
   });
   const isCounterShown = hiddenOptionsCount > 0;
+
+  const onDelete = useCallback(
+    option => {
+      onSelectedDelete(option, { action: "remove-value", removedValue: option });
+    },
+    [onSelectedDelete]
+  );
+
   const renderOptions = useCallback(
     (from = 0, to = selectedOptions.length) =>
       selectedOptions.map((option, index) => {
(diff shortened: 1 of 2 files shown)


src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.jsx
```diff
@@ -37,6 +37,14 @@ export default function Container({ children, selectProps, ...otherProps }) {
     selectedOptionsCount: selectedOptions.length
   });
   const isCounterShown = hiddenOptionsCount > 0;
+
+  const onDelete = useCallback(
+    option => {
+      onSelectedDelete(option, { action: "remove-value", removedValue: option });
```
SergeyRoyt, line 43: Basically creating event identical to what react-select has in such case

shlomitc (commented):
I suggest to add a test both for the new behavior (multiple) and for a single selection in case there isn't one

src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.jsx
```diff
@@ -52,7 +60,7 @@ export default function Container({ children, selectProps, ...otherProps }) {
             disabled={isDisabled}
             id={option.value}
             label={option.label}
-            onDelete={onSelectedDelete}
+            onDelete={() => onDelete(option)}
```
shlomitc, line 63: shouldn't it simply be
`onDelete={onDelete}`?

the `useCallback should have handled the callback

SergeyRoyt: Where would the `option` param come from then?

shlomitc: I see. so in that case you should pass the `onDelete` callback to the dependency array of the `useCallback` one

SergeyRoyt: Fixed

shlomitc (approved):

PR #1345, opened 2023-06-05
docs: Vibe 2 migration guide

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4587817551

### DIFF
--- a/VERSION-2-MIGRATION.md
+++ b/VERSION-2-MIGRATION.md
@@ -12,43 +12,78 @@ Version 1 won't receive new features, only critical bug fixes - so if you want t
 
 ## How to migrate?
 
-- Remove overrides of Vibe's global CSS classes (mostly prefixed with `monday-style-`). Some of them could be replaced with corresponding className props.  
-  Before:
-  ```diff
-  - .monday-style-button {
-  -   margin-left: var(--spacing-small);
-  - }
-  - <Button>Click me</button>
-  ```
-  After:
-  ```diff
-  + .click-me-button {
-    +   margin-left: var(--spacing-small);
-    + }
-  + <Button className="click-me-button">Click me</Button>
-  ```
-- Remove direct usages of Vibe's global CSS classes.  
-  Before:
-  ```diff
-  - <button className="monday-style-button ...">Click me</button>
-  ```
-  After:
-  ```diff
-  + <Button>Click me</Button>
-  ```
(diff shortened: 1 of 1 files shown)


VERSION-2-MIGRATION.md
```diff
@@ -0,0 +1,15 @@
+# Vibe 2 Migration
+
+## What's changed?
```
shlomitc, line 3: supported browsers
importing only from the root index file

VERSION-2-MIGRATION.md
```diff
@@ -0,0 +1,15 @@
...
+
+- CSS modules migration - we are now using css modules instead of global css classes, this allows us to have better encapsulation and fewer conflicts with other css classes.
+- `data-testid` attributes were added, so you can target components by them instead of global class names.
+
+## Why migrate?
+
+Version 1 won't receive new features, only critical bug fixes - so if you want to use new features you should migrate to version 2.
+
+## How to migrate?
```
shlomitc, line 12: The section should also include code examples. For example, for the data-testid:

Before:
```diff
- document.querySelector('monday-style-button').click();
```

After:
```diff
<Button data-testid="my-button"/>
+ document.querySelector(['data-testid="my-button"']).click();
```

For the style overrides - split it into two. the first one will be using the global monday-style classes and the second is overriding CSS directly which will no longer work.

Also, add here the topics from my comment above.

shlomitc (changes requested):
You can get some inspiration from this guide
https://mui.com/material-ui/migration/migration-v4/

VERSION-2-MIGRATION.md
SergeyRoyt: browsers list didn't get updated, though we wanted it originally, right?

SergeyRoyt: What about importing only from the root index file?

VERSION-2-MIGRATION.md
SergeyRoyt: done

VERSION-2-MIGRATION.md
shlomitc: browsers list - then let's do it. It should be only in the docs so no big change here.
index file - to prevent users from using `/dist/src` and simply entries

shlomitc: and also add that we no longer export the global class names, meaning only components should be used

PR #1352, opened 2023-06-06
added compact mode to MultiStepIndicator component

rami-monday:


src/components/MultiStepIndicator/MultiStepIndicator.jsx
```diff
@@ -27,17 +27,22 @@ const MultiStepIndicator = forwardRef(
...
+    const finalMode = useMemo(
+      () => (textPlacement === TEXT_PLACEMENTS.VERTICAL ? MODES.REGULAR : mode),
+      [mode, textPlacement]
+    );
 
     const renderHorizontalStepIndicator = useCallback(
       (step, index) => {
         return (
-          <>
+          <React.Fragment key={index}>
```
hadasfa, line 45: why we need to add key here?

rami-monday: Because it's mapped into array of step indicators below

src/components/MultiStepIndicator/MultiStepIndicator.jsx
```diff
@@ -8,7 +8,7 @@ import Check from "../../components/Icon/Icons/components/Check";
 import Divider from "../../components/Divider/Divider";
 import { NOOP } from "../../utils/function-utils";
 import StepIndicator from "./components/StepIndicator/StepIndicator";
-import { MULTI_STEP_TYPES, STEP_STATUSES, TEXT_PLACEMENTS } from "./MultiStepConstants";
+import { MODES, MULTI_STEP_TYPES, STEP_STATUSES, TEXT_PLACEMENTS } from "./MultiStepConstants";
```
hadasfa, line 11: We decided to replace modes with sizes for now :) can you replace it everywhere needed?

src/components/MultiStepIndicator/MultiStepIndicator.jsx
```diff
@@ -27,17 +27,22 @@ const MultiStepIndicator = forwardRef(
...
       textPlacement,
       id,
+      mode,
       "data-testid": dataTestId
     },
     ref
   ) => {
     const componentRef = useRef(null);
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    const finalMode = useMemo(
```
hadasfa, line 37: no need for use memo here :)

src/components/MultiStepIndicator/__stories__/multiStepIndicator.stories.mdx
```diff
@@ -194,6 +194,52 @@ There are 4 types: Primary, success, danger, dark.
   </Story>
 </Canvas>
 
+### Modes
```
hadasfa, line 197: please update title and description according to the figma (talk with Yael)

src/components/MultiStepIndicator/__stories__/multiStepIndicator.stories.mdx
```diff
@@ -194,6 +194,52 @@ There are 4 types: Primary, success, danger, dark.
...
+            key: "PENDING",
+            status: MultiStepIndicator.stepStatuses.PENDING,
+            titleText: "Pending",
+            subtitleText: "Pending subtitle"
+          }
+        ],
+        []
+      );
+      return (
+        <div className="monday-storybook-multiStepIndicator_column-wrapper">
```
hadasfa, line 228: As we discuss we have components for description and for layout which should be better for this case (will be consistent with other stories and code docs will be less confusing) - please use Flex and StoryDescription. You can see examples on Avatar or AttentionBox docs

hadasfa: I know that here we use these divs but these stories are relatively old

src/components/MultiStepIndicator/MultiStepIndicator.module.scss
```diff
@@ -9,4 +9,10 @@
   margin: 16px;
   height: 1px;
   width: 48px;
+
+  &.compact {
+    width: 16px;
+    margin-left: var(--spacing-small);
```
hadasfa, line 15: let's switch to margin-inline-start and margin-inline-end

hadasfa: Today, it is considered a best practice in the industry to utilize CSS that remains unaffected by the orientation of text (i.e., right-to-left or left-to-right). This approach ensures easier support for right-to-left languages in the future.

rami-monday: Nice, thanks

PR #1359, opened 2023-06-19
New title component and some docs fixes and tests

hadasfa:


src/components/Dialog/Dialog.tsx
```diff
@@ -14,6 +14,7 @@ import { AnimationType, HideShowEvent } from "../../constants/dialog";
 import { VibeComponentProps } from "../../types";
 import * as PopperJS from "@popperjs/core";
 import styles from "./Dialog.module.scss";
+import { ComponentDefaultTestId, getTestId } from "../../tests/test-ids-utils";
```
hadasfa, line 17: for tooltip interaction tests

src/components/Tooltip/__tests__/tooltip.interactions.ts
```diff
@@ -0,0 +1,36 @@
+import { expect } from "@storybook/jest";
+import { Screen } from "@testing-library/react";
+import { userEvent, waitFor } from "@storybook/testing-library";
+import { ComponentDefaultTestId, getTestId } from "../../../tests/test-ids-utils";
+import { ONE_LINE_ELLIPSIS_TEST_ID } from "../../Text/__tests__/text-tests-constants";
+
+export async function isTestTooltipShowOnHover(
```
shlomitc, line 7: This function is doing a multiple stuff
1. trigger show
2. assertion
3. trigger hide

so I think it's better to call it ("triggerTooltipAndAssert")

src/components/Typography/TypographyConstants.ts
```diff
@@ -0,0 +1,12 @@
+export enum TypographyColor {
+  PRIMARY = "primary",
+  SECONDARY = "secondary",
+  ON_PRIMARY = "onPrimary",
+  ON_INVERTED = "onInverted"
+}
+
+export enum TypographyAlign {
```
shlomitc, line 8: I thought we decided to remove it

shlomitc (changes requested):

src/components/Typography/TypographyConstants.ts
hadasfa: we not talked about it as far as I remember, why to remove it?

shlomitc: in most cases, layout is related to containers and not the element itself.
for example - a table cell will control the alignment of its content, and not the textual content itself
in other cases - you would probably use flex/position to align content that also contains text

src/components/Tooltip/__tests__/tooltip.interactions.ts
```diff
@@ -0,0 +1,36 @@
...
+  canvas: Screen,
+  getHoverableElement: () => Promise<HTMLElement>,
+  tooltipData: { id?: string; content?: string } = {}
+) {
+  // If we trying to get the element before it finishing it's first render we receive the instance without the tooltip wrapper and
+  // the userEvent.hover will not work (I did not find a way to implement this using react testing library or storybook interaction tests
+  await new Promise(resolve => {
```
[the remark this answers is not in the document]
hadasfa: Did no found any other solution that worked (wait for did not work here) - what do you suggest

src/components/Tooltip/__tests__/tooltip.interactions.ts
hadasfa: Assert is not clear enough... I can change to testHoverTooltipTrigger?

shlomitc (approved):
just try to improve the test and lets go

shlomitc (approved):

PR #1369, opened 2023-06-25
fix: icons, interactionsTests, testIds types exports

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4633056724


webpack/published-ts-components.js
```diff
@@ -103,8 +103,6 @@ const publishedTSComponents = {
   useActiveDescendantListFocus: "hooks/useActiveDescendantListFocus",
   useListenFocusTriggers: "hooks/useListenFocusTriggers",
   useSwitch: "hooks/useSwitch",
-  // Don't remove next line
```
shlomitc, line 106: are you sure? :)

package.json
```diff
@@ -12,14 +12,27 @@
...
+    "*": {
+      "icons": [
+        "./dist/esm/components/Icon/Icons/index.d.ts"
+      ],
+      "interactionsTests": [
+        "./dist/esm/tests/interactions-utils.d.ts"
+      ],
+      "testIds": [
+        "./dist/esm/tests/test-ids-utils.d.ts"
+      ]
```
shlomitc, line 33: Why are these two needed to be publically exposed APIs?
The `interactionsTests` are mostly wrappers from regular DOM functions and they are very platform specific

webpack/published-ts-components.js
SergeyRoyt: Pretty sure
We have this marker on the line 87

package.json
SergeyRoyt: `testIds` are containing `ComponentDefaultTestId` which contains all default data-testid for components and it's being used already in the monolith in cypress tests selectors
image

`interactionsTests` are exporting functions which helps to build storybook interaction tests and this endpoint is being used in the `monday-ui-components` rn
image

Also third one should be `storybookComponents` to export components we are using to build our storybook, but as they are not converted to ts, they are not here

shlomitc: ok. I'm not too fond of it as a public API. it creates a coupling between Vibe's internal testing tools with others who consume it, making it harder to refactor. if that's the current state, then fine, but let's try to reduce such dependencies

shlomitc (approved):

PR #1374, opened 2023-06-26
Not complete: Infra for run css plugins for removing legacy

hadasfa:
Current blockers:
- Parser can't handle inline comments 
- Parser try to parse monday-ui-react-core node modules (will fixed when moved to external package)


codemod/src/scripts/modify.ts
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env node
+
```
hadasfa, line 2: I will write docs for these in the documentation phase (small readme about how to use this runner and plugins)

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.module.scss
```diff
@@ -36,7 +36,7 @@
 }
 
 .textContainerTitleText {
-  font: var(--font-h5);
+  font: var(--font-text-bold);
   font-size: 14px;
```
shlomitc, line 40: 1. I'd expect the codemod to remove all font-related properties
2. we might want to change the codemod to replace the variables by extending the global classes

codemod/src/plugins/add-imports-if-needed/test.spec.tsx
```diff
@@ -0,0 +1,43 @@
+import * as postcss from "postcss";
```
hadasfa, line 1: add import if not already exists

codemod/src/plugins/plugins.ts
```diff
@@ -0,0 +1,6 @@
+import { Plugin } from "postcss";
+import { replaceTypographyTokensPlugin } from "./replace-typography-tokens";
+const PLUGIN_ARRAY = [replaceTypographyTokensPlugin];
```
hadasfa, line 3: All plugins which available to run by using the package runer

codemod/src/plugins/replace-tokens-with-mixins/index.ts
```diff
@@ -0,0 +1,31 @@
+import * as postcss from "postcss";
```
hadasfa, line 1: replace one token usage with include mixin

codemod/src/plugins/replace-tokens/index.ts
```diff
@@ -0,0 +1,14 @@
+import * as postcss from "postcss";
+
```
hadasfa, line 2: old plugin of replacing one token with other token. not needed after the changes we made but kept it for future use, I can remove it if you prefer

codemod/src/plugins/replace-typography-tokens/index.ts
```diff
@@ -0,0 +1,166 @@
+import * as postcss from "postcss";
```
hadasfa, line 1: This plugin will replace font tokens with the typography mixins with the matched params, and if there is usage with combined font-size, font-weight and line-height with the matched params it will replace it with the match mixin

codemod/src/scripts/modify.ts
hadasfa, line 2: you can run this like this: modify file/path replace-typography-tokens

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.module.scss
hadasfa: changed it to be as we discussed before

PR #1379, opened 2023-06-28
Migrate to new text components part 1

hadasfa:
#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -118,7 +122,9 @@ const AttentionBox: React.FC<AttentionBoxProps> & {
             iconLabel={iconLabel}
           />
         )}
-        {text || children}
+        <Text size="small" paragraph>
+          {text || children}
+        </Text>
```
shlomitc, line 127: I wonder if it should have been 
```tsx
text ? <Text size="small" paragraph>{text}</Text> : children
```

src/components/Heading/Heading.module.scss
```diff
@@ -7,6 +7,7 @@
   border-radius: $border-radius-small;
   min-width: 0;
   @include heading-padding();
+  margin: revert;
```
shlomitc, line 10: why exactly is this here?

src/components/Heading/Heading.module.scss
```diff
@@ -27,63 +28,6 @@
...
-
-.elementTypeP {
-  @include font-paragraph();
-}
-
-.elementTypeSmall {
-  @include font-caption();
-}
-
 .elementTypeLink {
```
shlomitc, line 31: what about this one?

src/components/Heading/HeadingHooks.ts
```diff
@@ -0,0 +1,22 @@
+import { HeadingSizes, HeadingTypes } from "./HeadingConstants";
+
+export function useLegacyHeadingClassNameByType(
```
shlomitc, line 3: that's not exactly a hook but rather a simple function, no?

src/components/Heading/Heading.tsx
```diff
@@ -54,17 +52,12 @@ const Heading: React.FC<HeadingProps> & {
...
-      [styles.singleLineEllipsis]: ellipsis && ellipsisMaxLines <= 1,
-      [styles.suggestEditOnHover]: suggestEditOnHover
-    }
-  );
+  const typographyClassName = useLegacyHeadingClassNameByType(type, size);
+  const classNames = cx(styles.headingComponent, typographyClassName, className, {
+    [styles.multiLineEllipsis]: ellipsis && ellipsisMaxLines > 1,
+    [styles.singleLineEllipsis]: ellipsis && ellipsisMaxLines <= 1,
+    [styles.suggestEditOnHover]: suggestEditOnHover
+  });
```
shlomitc, line 60: aren't these coming from the `Title` component?

src/components/Title/__tests__/__snapshots__/text-snapshot-tests.jest.js.snap
```diff
@@ -1,21 +1,21 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Text renders correctly with align 1`] = `
-<span
+<h1
```
shlomitc, line 4: just making sure this is correct

src/components/Toast/ToastLink/ToastLink.module.scss
```diff
@@ -3,7 +3,7 @@
 
 .actionLink {
   @include theme-prop(color, text-color-on-primary);
-  text-decoration: underline;
+  text-decoration: underline !important;
```
shlomitc, line 6: that should probably be a prop for the `TextLink` component

src/components/Typography/Typography.module.scss
```diff
@@ -23,6 +23,9 @@
 
   &.onPrimary {
     color: var(--text-color-on-primary);
+    a {
+      color: var(--text-color-on-primary);
+    }
```
shlomitc, line 28: shouldn't this be for every type?

src/components/Typography/Typography.tsx
```diff
@@ -11,7 +11,7 @@ import { useEllipsisClass, useTooltipProps } from "./TypographyHooks";
 import Tooltip, { TooltipProps } from "../Tooltip/Tooltip";
 import styles from "./Typography.module.scss";
 
-export interface TypographyProps extends VibeComponentProps {
+export interface TypographyProps extends VibeComponentProps, HTMLAttributes<HTMLElement> {
```
shlomitc, line 14: how would this affect the docs?

shlomitc (changes requested):

src/components/AttentionBox/AttentionBox.tsx
hadasfa: if it is not a text, it will be a composition of text and other stuff and in that case we still prefer that the component receive our typography styles

src/components/Heading/Heading.module.scss
hadasfa: heading component contains the default margin which received from the browser, removing it will be a breaking change

src/components/Heading/HeadingHooks.ts
hadasfa: i'm a huge fan of custom hook for readability but you right... it  will be better to set this as function, will fixed

src/components/Heading/Heading.tsx
hadasfa: This is the legacy heading one, for avoiding weird bug and changes outside of our scope, I only changed the css and kept the funcunality  as it was before (our title component also support eliipsis and max lines... but I used here only classes)

src/components/Title/__tests__/__snapshots__/text-snapshot-tests.jest.js.snap
hadasfa: Yes it is a bug fix, bug I will update also the snapshot file name (text => title)

src/components/Toast/ToastLink/ToastLink.module.scss
hadasfa: what do you mean? Right now we don't have link component only link skin. 
The reason we want to add here the underline is because  link in on primary is in the same color as text... maybe we can just set underline as default on link skin with onPrimary color?

src/components/Typography/Typography.tsx
hadasfa: No effect on docs :)

src/components/AttentionBox/AttentionBox.tsx
shlomitc: I agree, but it might affect it badly in case of ellipsis wrapping

src/components/Heading/Heading.module.scss
shlomitc: so worth adding a comment and make sure the chromatic visual tests are ok

src/components/Heading/Heading.tsx
shlomitc: ok

src/components/Heading/Heading.module.scss
hadasfa: I removed all font detentions and kept only link appearance related styles (which are not part of the class currently...)

Not sure if we want to define color related styles in the class

src/components/Toast/ToastLink/ToastLink.module.scss
shlomitc: As I see it - the toast link button should be either the `<Link/>` which can take the appropriate skin (onPrimary?) and have the ability to control the underline (e.g. never, onHover, always)

src/components/Typography/Typography.module.scss
hadasfa: In other colors the link color should be link-color, but onPrimary and onInverted links should be in the same color as the other text (probably with underline)

src/components/AttentionBox/AttentionBox.tsx
hadasfa: paragraph is without ellipsis by default... Attention box is not support ellipsis before and still not supporting it

src/components/Heading/Heading.module.scss
hadasfa: Adding a comment and will check chromatic tests before merge

src/components/Toast/ToastLink/ToastLink.module.scss
hadasfa: After I discuss with Meirav I added the link appearance behaviour to the component behviour (no underline unless it's primary, that's the appearance and behivour Meirav want).

src/types/ElementContent.ts
```diff
@@ -1,3 +1,3 @@
-import { ReactElement } from "react";
+import { ReactNode } from "react";
 
-export type ElementContent = string | ReactElement | ReactElement[] | JSX.Element | JSX.Element[];
+export type ElementContent = string | string[] | ReactNode | ReactNode[];
```
shlomitc, line 3: isn't `string` already covered in `ReactNode`?

src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -96,11 +96,19 @@ const AttentionBox: React.FC<AttentionBoxProps> & {
...
+            {title}
+          </Text>
+        )}
       </Flex>
-      <div
+      <Text
+        element={compact ? "div" : "p"}
+        size="small"
+        paragraph={!compact}
         className={cx(styles.text, getStyle(styles, camelCase(classNameWithType + "__text")), {
```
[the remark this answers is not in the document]
hadasfa: removed in the new pr

src/types/ElementContent.ts
hadasfa: fix in new pr

PR #1390, opened 2023-07-06
chore: enable source maps in tsconfig

YossiSaadi:
note: in current config, this should allow having full source map in a prod build and eval in dev mode (storybook)


SergeyRoyt (approved):

SergeyRoyt:
We are also usually including link to the monday task here in the PR description
And inserting the link to the PR in `Github link` field on the board + setting the status on review
I've moved your task to current iteration, could you please do the rest?

PR #1392, opened 2023-07-06
chore: fail chromatic action check in case has visual changes

YossiSaadi:


shlomitc:
That's perfect
Please take a look at this change
image

It seems like the DatePicker tests are alwais failing because the date constantly changes.
So let's fix this and we can close this ticket as well (https://monday.monday.com/boards/3532714909/pulses/4474765293)

hadasfa (approved):

YossiSaadi:
Note: Will be merged once we have a defined baseline for Chromatic on master

PR #1403, opened 2023-07-10
feat(MultiStepIndicator): add ts support

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/3938868803


src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
```diff
@@ -86,6 +116,7 @@ const StepIndicator = ({
 
   // Event listeners for removing animation.
   useEventListener({
+    // @ts-ignore TODO either fix the import to 'useEventListener' OR fix to fit 'useKeyEvent' OR remove entirely
```
YossiSaadi, line 119: needs to be discussed
it seems that as for today, since it uses wrong import (`useEventListener` is imported mistakenly from `useKeyListener`), the api is wrong.
`useKeyListener` does not have `eventName`, but rather `keyEventName`, causing `animationend` event to not be assigned to `componentRef`, and instead assigning it with `keydown` event.

This does not seem to have effect eventually as `disableStatusChangeAnimation` is never called (but weirdly, the component never gets the `withAnimation` class declared above 🥲 plz help 🥲)

src/components/MultiStepIndicator/MultiStepIndicator.tsx
```diff
@@ -12,32 +9,58 @@ import { SIZES, MULTI_STEP_TYPES, STEP_STATUSES, TEXT_PLACEMENTS } from "./Multi
 import { getTestId } from "../../tests/test-ids-utils";
 import { ComponentDefaultTestId } from "../../tests/constants";
 import styles from "./MultiStepIndicator.module.scss";
+import { withStaticProps, VibeComponentProps, VibeComponent, SubIcon } from "../../types";
+import { IconType } from "../Icon/IconConstants";
 
-const MultiStepIndicator = forwardRef(
+export type Step = {
```
YossiSaadi, line 15: Is it ok, according to your standards, to export the `Step` type like that? it makes sense since it's only relevant (at least for now) to the `MultiStepIndicator` scope (story + tests)

src/components/MultiStepIndicator/MultiStepConstants.ts
```diff
@@ -0,0 +1,22 @@
+export enum MULTI_STEP_TYPES {
```
hadasfa, line 1: We usually change the enum naming to be more like: MultiStepType

src/components/MultiStepIndicator/MultiStepConstants.ts
```diff
@@ -0,0 +1,22 @@
...
+  PRIMARY = "primary",
+  SUCCESS = "success",
+  DANGER = "danger",
+  DARK = "dark"
+}
+
+export enum STEP_STATUSES {
```
hadasfa, line 8: StepStatus

src/components/MultiStepIndicator/MultiStepIndicator.tsx
hadasfa: It's make sense. Did you see similar exports in other components? (Just to make sure we keep the API consistent as possible)

src/components/MultiStepIndicator/MultiStepIndicator.tsx
```diff
@@ -12,32 +9,58 @@ import { SIZES, MULTI_STEP_TYPES, STEP_STATUSES, TEXT_PLACEMENTS } from "./Multi
...
+  fulfilledStepIcon?: SubIcon;
+  fulfilledStepIconType?: IconType.SVG | IconType.ICON_FONT;
+  isFulfilledStepDisplayNumber?: boolean;
+  onClick?: (stepNumber: number) => void;
+  textPlacement?: TEXT_PLACEMENTS;
+  size?: SIZES;
+}
+
+const MultiStepIndicator: VibeComponent<MultiStepIndicatorProps> & {
+  types?: typeof MULTI_STEP_TYPES;
```
hadasfa, line 35: @SergeyRoyt I think that we can remove the ? after you merged the new use static props hook, am I right?

src/components/MultiStepIndicator/MultiStepIndicator.tsx
```diff
@@ -12,32 +9,58 @@ import { SIZES, MULTI_STEP_TYPES, STEP_STATUSES, TEXT_PLACEMENTS } from "./Multi
...
       steps,
+      type = MULTI_STEP_TYPES.PRIMARY,
       stepComponentClassName,
       dividerComponentClassName,
-      fulfilledStepIcon,
-      fulfilledStepIconType,
-      isFulfilledStepDisplayNumber,
-      onClick,
-      textPlacement,
+      fulfilledStepIcon = Check,
```
hadasfa, line 47: Just make sure you don't change here the default props values by mistake (for example undefined => false or Noop => undefined)

hadasfa: Because it can lead to unexpected behaviour

src/components/MultiStepIndicator/__tests__/multiStepIndicator-snapshot-tests.jest.tsx
```diff
@@ -23,7 +23,7 @@ const exampleSteps = [
 
 describe("MultiStepIndicator renders correctly", () => {
   it("with empty props", () => {
-    const tree = renderer.create(<MultiStepIndicator />).toJSON();
+    const tree = renderer.create(<MultiStepIndicator steps={[]} />).toJSON();
```
hadasfa, line 26: If steps can be rendered with empty array why steps is required?

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
```diff
@@ -18,48 +17,79 @@ import Clickable from "../../../../components/Clickable/Clickable";
...
-      iconLabel={STEP_STATUSES.FULFILLED}
-      iconType={fulfilledStepIconType}
-      ignoreFocusStyle
-      clickable={false}
-      ariaHidden={true}
-    />
-  ) : (
-    stepNumber
+  return (
+    <>
```
hadasfa, line 42: Why it was needed?

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
```diff
@@ -18,48 +17,79 @@ import Clickable from "../../../../components/Clickable/Clickable";
...
-  status,
-  titleText,
-  subtitleText,
-  type,
-  fulfilledStepIcon,
-  fulfilledStepIconType,
-  isFulfilledStepDisplayNumber,
-  onClick,
-  isFollowedByDivider,
+  stepNumber = 1,
```
hadasfa, line 79: same comment about double check that the default values of the props did not changed by mistake :)

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
hadasfa: Sound good to me.  Can you open a task for it in our backlog with link to here?

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
```diff
@@ -139,18 +170,18 @@ const StepIndicator = ({
         <div
           className={cx(...getClassNamesWithSuffix("__number-container"))}
           ref={componentRef}
-          tabIndex="0"
+          tabIndex={0}
           role="button"
         >
           <SwitchTransition mode="out-in">
-            <CSSTransition
+            <CSSTransition<undefined> /* https://stackoverflow.com/questions/64019519/typing-out-react-transition-group-handlers-in-typescript */
```
hadasfa, line 177: mmmmm Can you explain?

src/components/MultiStepIndicator/__tests__/multiStepIndicator-snapshot-tests.jest.tsx
YossiSaadi: exactly for the reason mentioned above, to not break/change existing API.
you want to change it anyway to be optional?

YossiSaadi: before my change:
```
steps: PropTypes.arrayOf(
  PropTypes.shape({
    titleText: PropTypes.string,
    subtitleText: PropTypes.string,
    status: PropTypes.oneOf([
      MultiStepIndicator.stepStatuses.PENDING,
      MultiStepIndicator.stepStatuses.ACTIVE,
      MultiStepIndicator.stepStatuses.FULFILLED
    ])
  })
).isRequired,
```

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: since I wanted `StepCircleDisplay` to always return JSX, not a JSX | number, as it causes TS issues on render

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: Yep checked everything again, seems correct :)
although the prop is set to be required, I wanted to keep the existing behavior (it is currently set as required but has default)
```
StepIndicator.propTypes = {
  ...
  stepNumber: PropTypes.number.isRequired,
  ...
}

StepIndicator.defaultProps = {
  ...
  stepNumber: 1,
  ...
}
```

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: CSSTransition needs to be specified with the generic parameter to decide type for callback (out of few possible)
We can add instead ts-ignore, but if we don't specifiy the generic param the callback `addEndListener` expects only one argument (of `done`, without `node`)

YossiSaadi: image

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: https://monday.monday.com/boards/3532714909/pulses/4817660390

src/components/MultiStepIndicator/MultiStepIndicator.tsx
YossiSaadi: Actually after checking it, yep, seems like it is a pattern when needed

src/components/MultiStepIndicator/MultiStepIndicator.tsx
YossiSaadi: @hadasfa can you elaborate on that please?

src/components/MultiStepIndicator/MultiStepIndicator.tsx
YossiSaadi: Add it to API_GUIDE

src/components/MultiStepIndicator/__tests__/multiStepIndicator-snapshot-tests.jest.tsx
YossiSaadi: Change API for steps to be optional

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: check if I can declare 'as' casting for the number without Fragments

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: remove required

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: Move the comment to be above the <CssTran..> and remove stackoverflow link - explain in plain text

src/components/MultiStepIndicator/MultiStepIndicator.tsx
YossiSaadi: https://monday.monday.com/boards/3532714909/pulses/4818659218

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
YossiSaadi: @hadasfa not possible unfortunately
https://stackoverflow.com/a/70895599/7363120

seems like a TS issue

Though React allows a component to return a number and not necessarily JSX/ReactNode, TS typing does not allow it.. :/

src/components/MultiStepIndicator/__stories__/helper.ts
```diff
@@ -1,6 +1,6 @@
-import MultiStepIndicator from "../MultiStepIndicator";
+import MultiStepIndicator, { type Step } from "../MultiStepIndicator";
```
SergeyRoyt, line 1: Maybe worth moving `Step` to constants

src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
```diff
@@ -15,51 +14,77 @@ import Divider from "../../../../components/Divider/Divider";
...
   fulfilledStepIconType,
-  stepNumber,
-  size
+  stepNumber
 }) => {
-  return status === STEP_STATUSES.FULFILLED && !isFulfilledStepDisplayNumber ? (
+  return status === StepStatus.FULFILLED && !isFulfilledStepDisplayNumber ? (
     <Icon
       icon={fulfilledStepIcon}
-      className={classNames(styles.numberContainerTextCheckIcon, { [styles.compact]: size === SIZES.COMPACT })}
```
SergeyRoyt, line 35: compact style is not needed?

src/components/MultiStepIndicator/__stories__/helper.ts
YossiSaadi: Thanks for the review @SergeyRoyt :)
@hadasfa and I discussed about it and decided that since it is coupled to MultiStepIndicator and it's relevant usages, it might be better to have this connection between them, since Step should never be used unless it's for MultiStepIndicator implementation

I'm not sure actually about industry standards best practice for it
I'll make further check

YossiSaadi: From what I've checked, in case Step is always used only when MultiStep is being imported and used, it makes more sense to export it directly from the component, it seems to be a pretty wide standard (which we use on other components also)
is that ok to leave it that way, @SergeyRoyt?

SergeyRoyt: Yeah, I think it's ok

hadasfa (approved):

PR #1405, opened 2023-07-11
Sergeyro/fix/remove builded icons

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4786437354


.github/workflows/chromatic.yml
```diff
@@ -1,22 +1,21 @@
...
   chromatic-deployment:
     # Operating System
     runs-on: ubuntu-latest
-    # Job steps
     steps:
       - uses: actions/checkout@v1
       - name: Install dependencies
         run: npm ci
-        # 👇 Adds Chromatic as a step in the workflow
+      - name: Build react icons
```
shlomitc, line 17: why is this needed? isn't it enough to to be ran as part of the regular build step in the package.json?

shlomitc (changes requested):

.github/workflows/chromatic.yml
SergeyRoyt: @shlomitc It isn't enough - it was crashing on all these steps before

shlomitc: this is the only one I see that doesn't have any build actions at all, what command is it running behind the scenes that build the storybook?

.github/workflows/test-workflow.yml
```diff
@@ -18,6 +18,8 @@ jobs:
           npm ci
         env:
           NODE_AUTH_TOKEN: ${{secrets.npm_token}}
+      - name: Build react icons
```
shlomitc, line 21: in the same context as the other comment - then it sounds to me like it should be added to each of the commands and not only for the CI.

As a best practice - the CI steps should only be familiar with only one command from the package.json and not try to have some logic composition multiple commands

SergeyRoyt: Combined it via `ci` script in `package.json`

SergeyRoyt:
Closed in favor of this solution

PR #1406, opened 2023-07-11
test(useSwitch): add tests for hook

YossiSaadi:


src/hooks/useSwitch/index.ts
```diff
@@ -5,17 +5,14 @@ enum SwitchRole {
...
-}: {
+export interface UseSwitchProps {
   isChecked?: boolean;
   defaultChecked?: boolean;
   onChange?: (value: boolean) => void;
   isDisabled?: boolean;
-}) {
+}
+
+export default function useSwitch({ isChecked, defaultChecked, onChange, isDisabled }: UseSwitchProps = {}) {
```
YossiSaadi, line 15: @hadasfa, please notice that I've added an empty object fallback for the hook (to allow using it without passing empty object in case no modifications should be done)
If you don't agree with it, and it goes against one of the Vibe principles, I'll revert

hadasfa (approved):
Awesome!

hadasfa (commented):
Just please check if the hook is exported before merging. Check Published components and hooks/index

PR #1409, opened 2023-07-11
fix: InteractionsTest - fix types

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4729201760


shlomitc (approved):
In general, if we're exposing `interaction-utils` and using `@testing-library/react` inside - then it should become a dependency of the project instead of devDependency

PR #1419, opened 2023-07-13
improve useState to only called once in useIsOverflowing

doronbrikman:
We saw some long task on the browser coming from `Heading` component, seems that the source of the issue is this styling accessing:

```
function checkOverflow(element: HTMLElement, ignoreHeightOverflow: boolean) {
...
  const isOverflowing =
    element.clientWidth < element.scrollWidth || (!ignoreHeightOverflow && element.clientHeight < element.scrollHeight);
...
}
```
We might can improve this, but first thing we can do is to reduce the times it calling this function by changing the `useState` function invocation to only call this once.

#### Basic
- [ ] PR has description.
#### Style
#### Storybook
#### Tests


hadasfa (approved):
Thank you for contributing :) !

PR #1422, opened 2023-07-17
fix(Chips): change border color token, do not limit chip-with-border's height

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4818138972

as it is using box-sizing of border-box, limiting the height to 22px causes the chip to be with total height of 22px instead of like regular chip which is 24px.

Before:
image

After:
image

---

Bug fix:
It caused this bug:
image

✅ Before, without border - correct (24px)
image

❌ Before, with border - wrong (22px)
image

✅ After, with border - correct (24px)
image

### DIFF
--- a/src/components/Chips/Chips.module.scss
+++ b/src/components/Chips/Chips.module.scss
@@ -18,9 +18,7 @@
   animation: chips-pop-in-emphasized var(--motion-productive-medium) var(--motion-timing-emphasize);
   @include font-general-text();
   &.border {
-    height: 22px;
-    border: 1px solid var(--text-color-on-primary);
-    line-height: 21px;
+    border: 1px solid var(--primary-background-color);
   }
   &.withUserSelect {
     user-select: text;


src/components/Chips/Chips.module.scss
```diff
@@ -18,9 +18,7 @@
   animation: chips-pop-in-emphasized var(--motion-productive-medium) var(--motion-timing-emphasize);
   @include font-general-text();
   &.border {
-    height: 22px;
-    border: 1px solid var(--text-color-on-primary);
-    line-height: 21px;
+    border: 1px solid var(--primary-background-color);
```
SergeyRoyt, line 21: I think `primary-background-color` is a mistake here. Have Meirav told you to use this one?
image
image

You know how to switch themes in Storybook? It's this button
image

SergeyRoyt (changes requested):
Doesn't look good in other themes than Light - need to use different color token

YossiSaadi:
> Doesn't look good in other themes than Light - need to use different color token

@SergeyRoyt this if from Figma, Meirav asked for this one
image

SergeyRoyt (approved):

PR #1425, opened 2023-07-18
PR #3 of replacing old token variants with mixins

hadasfa:
After this pr the only token which left is the font-family token. There is a reason why I did not removed it, let's talk about it


src/components/Combobox/Combobox.module.scss
```diff
@@ -3,10 +3,10 @@
 @import "../../styles/states";
 
 .combobox {
+  @include smoothing-text;
```
hadasfa, line 6: need to fix this before marge (design review)
Screen Shot 2023-07-19 at 14 45 49

src/components/ListTitle/ListTitle.module.scss
```diff
@@ -2,7 +2,7 @@
 @import "../../styles/typography";
 
 .listTitle {
-  @include font-paragraph-bold();
+  @include smoothing-text;
```
hadasfa, line 5: need to be discussed between Meirav and Yael

src/components/RadioButton/RadioButton.module.scss
```diff
@@ -59,13 +59,6 @@ input:checked:disabled + .control {
   @include theme-prop(background-color, secondary-text-color);
 }
 
-.label {
-  line-height: 21px;
```
hadasfa, line 63: fix align

src/components/Combobox/Combobox.module.scss
hadasfa: reason: default line height change. 
Added 4 pixel padding fix fix it (and extra 2 for make the sticky header positioned better)

src/components/IconButton/__stories__/iconButton.stories.module.scss
```diff
@@ -22,19 +23,19 @@
   }
 
   &BoardName{
-    font-weight: var(--font-weight-bold);
+    font-weight: 500;
```
shlomitc, line 26: Using a variable

src/components/TextField/TextField.module.scss
```diff
@@ -252,7 +253,7 @@
 }
 
 .textField .inputWrapper.wrapperSizeMedium .iconContainer .icon {
-  font-size: var(--font-size-paragraph);
+  font-size: 16px;
```
shlomitc, line 256: Same

shlomitc (changes requested):

src/components/Avatar/__stories__/avatar.stories.mdx
```diff
@@ -150,39 +150,6 @@ Use when a user’s image is not available, use their initials.
   </Story>
 </Canvas>
 
-### Initials avatar custom text format
```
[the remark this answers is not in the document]
hadasfa: In the design review with Yael we decided to remove this story because we do not want people to override the appearance of the letters inside the avatar (it should be consistent)

src/components/TextField/TextField.module.scss
hadasfa: As discussed we deiced to keep it because it's a font size for font icon and not text

PR #1430, opened 2023-07-21
Replace storybookComponents with vibe-storybook-components

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4803909254


package.json
```diff
@@ -47,7 +47,7 @@
     "build:esm": "rollup -c",
     "build:esm:mock-classnames": "mock_classnames=on rollup -c",
     "build:react-icons": "svg2react-icon --typescript --keep-colors node_modules/monday-ui-style/src/Icons src/components/Icon/Icons",
-    "storybook": "storybook=true start-storybook -p 7007",
+    "storybook": "storybook=true start-storybook -p 7008",
```
shlomitc, line 50: why?

.storybook/preview.js
```diff
@@ -30,23 +31,27 @@ addParameters({
...
+      ComponentName: ComponentNameDecorator,
       ComponentRules,
       UsageGuidelines,
       FunctionArguments,
       FunctionArgument,
-      RelatedComponents
+      RelatedComponents: RelatedComponentsDecorator,
+      Frame,
+      UnstyledList,
+      UnstyledListItem
```
shlomitc, line 54: what are these?

.storybook/preview-head.html
```diff
@@ -26,18 +26,6 @@
...
-  }
-  .main-story-title {
-    font-size: 32px;
-    color: #323338;
-  }
-  .story-title {
-    color: #323338;
-    font-size: 18px;
-    margin-bottom: 48px;
-  }
```
shlomitc, line 40: why?

src/storybook/functions/createStoryMetaSettingsDecorator.js
```diff
@@ -0,0 +1,21 @@
+import { iconsMetaData } from "monday-ui-style/src/Icons/iconsMetaData";
+import * as AllIcons from "../../components/Icon/Icons";
+import { createStoryMetaSettings } from "vibe-storybook-components";
+
+export function createStoryMetaSettingsDecorator({
```
shlomitc, line 5: 1. Can you explain why this change was done?
2. Worth adding a comment explaining the purpose of this function.
3. Should it affect the plop as well?

src/components/VirtualizedList/__stories__/VirtualizedList.stories.mdx
```diff
@@ -4,7 +4,8 @@ import { generateItems } from "./virtualizedList.components";
 import { useCallback } from "react";
 import { Flex } from "../../";
 import styles from "./virtualizedList.module.scss";
-import { Tip, UsageGuidelines } from "../../../storybook";
+import { Tip } from "../../../storybook";
```
shlomitc, line 7: what is Tip? why is it here and not in vibe-storybook-components?

shlomitc (changes requested):

package.json
SergeyRoyt: For convenience - it's running on same port a monday-ui-components does, so it makes it uncomfortable to run both packages at the same time

.storybook/preview.js
SergeyRoyt: Some storybook components, which we are using...
Putting them here makes it not required to import them in stories files. So, for consistency I've put all of used components here

.storybook/preview-head.html
SergeyRoyt: Not needed any more - it was used only in storybook-helpers components, which weren't used, so I've deleted them

src/components/VirtualizedList/__stories__/VirtualizedList.stories.mdx
SergeyRoyt: Tip is implemented via `AttentionBox` component. We can duplicate it in `vibe-storybook-components`, but it requires some efforts, so I've decided not to put it into the first version

src/storybook/functions/createStoryMetaSettingsDecorator.js
SergeyRoyt: 1,2. Added a comment there 
`Decorator on createStoryMetaSettings from vibe-storybook-components - adds icons metadata and icons components to function arguments`
3. Plop changes were made already

package.json
shlomitc: I think storybook asks you each time if you want to try a different port if another process already runs on it, no?

src/components/VirtualizedList/__stories__/VirtualizedList.stories.mdx
shlomitc: ok, but let's put it as a TODO somewhere, so it can be a good first task.

SergeyRoyt: Added it to our backlog

package.json
SergeyRoyt: It asks you and you have to input "y". It's not that a big deal, but still not really convenient. Especially, when it can be easily solved just by changing port on our side

shlomitc: ok

shlomitc (approved):

PR #1434, opened 2023-07-25
fix(AttentionBox): general improvements + allow ellipsis for overflowin attention box

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4183198730

- remove redundant div when no title: no need to render the wrapper flex div in case there is no title (if there's no title, now a div is rendered without content in it)
- add extra padding-right only if attention box is compact and dismissable
- make attention box stretched to container
- allow ellipsis for overflowing attention box


src/components/AttentionBox/AttentionBox.module.scss
```diff
@@ -6,7 +6,7 @@
   position: relative;
   padding: var(--spacing-medium);
   border-radius: var(--border-radius-medium);
-  width: fit-content;
+  width: 100%;
```
hadasfa, line 9: dangerous a bit (breaking change)

hadasfa (commented):
let's find a way to make it work without change the default width

src/components/AttentionBox/AttentionBox.module.scss
YossiSaadi: but we said that we are intentionally want it 👀

YossiSaadi: will try to solve with fit-content 👍🏼

YossiSaadi:
@hadasfa it seems to work ok with max-width: 100% and width of fit-content.
We're keeping the same behavior of fitting to content in case it does not fit to 100% of container, it will shrink to fit-content

src/components/AttentionBox/AttentionBox.module.scss
```diff
@@ -7,6 +7,7 @@
   padding: var(--spacing-medium);
   border-radius: var(--border-radius-medium);
   width: fit-content;
+  max-width: 100%;
```
hadasfa, line 10: Great solution!

hadasfa (approved):
in general looking good, just want to make sure how it behave in chromatic before merge

PR #1439, opened 2023-07-27
feat: <List> a11y improvements

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4882273102


src/components/List/List.tsx
```diff
@@ -41,13 +45,14 @@ const List: FC<ListProps> = forwardRef(
       ariaLabel,
       ariaDescribedBy,
       renderOnlyVisibleItems = false,
+      defaultFocusIndex = 0,
```
eladdo92, line 48: I think it should be `focusIndex` and also be able to get updates (in the current implementation if the focused item gets changed from outside the inner state wouldn't know it)

src/components/List/__stories__/List.stories.mdx
```diff
@@ -96,6 +96,29 @@ Lists is a group of actionable items containing primary and supplemental actions
...
+          {[0, 1, 2, 3, 4, 5].map(index => (
+            <ListItem selected={selectedItemIndex === index} onClick={() => updateSelectedItem(index)}>
+              {`List item ${index}`}
+            </ListItem>
+          ))}
+        </List>
+      );
+    }}
+  </Story>
+</Canvas>
```
SergeyRoyt, line 120: TODO remove after review - just for demo

src/components/ListItem/__tests__/__snapshots__/listItem.jest.js.snap
```diff
@@ -1,71 +1,76 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`renders correctly with disabled 1`] = `
-<div
+<li
   aria-disabled={true}
   className="typography primary start singleLineEllipsis small normal listItem small disabled"
   data-testid="list-item"
+  id="undefined-item-undefined"
```
shlomitc, line 8: mistake?

src/components/List/List.tsx
```diff
@@ -71,6 +81,23 @@ const List: FC<ListProps> = forwardRef(
       },
       [focusIndex]
     );
+
+    useEffect(() => {
+      if (!id) {
+        console.warn("List should have a valid id prop");
```
shlomitc, line 87: does this mean the component can't function without it, if so, then it should be more than just a warning

shlomitc (changes requested):
I'll be happy to see some test that covers this logic as it is very hard to dive into the technical details of this PR

src/components/ListItem/ListItem.tsx
```diff
@@ -72,57 +76,72 @@ const ListItem: FC<ListItemProps> & { sizes?: typeof SIZES } = forwardRef(
...
+        element="li"
         data-testid={dataTestId || getTestId(ComponentDefaultTestId.LIST_ITEM, id)}
         ref={mergedRef}
         className={cx(styles.listItem, className, getStyle(styles, camelCase(size)), {
           [styles.selected]: selected && !disabled,
           [styles.disabled]: disabled
         })}
-        id={id}
+        id={overrideId}
+        key={overrideId}
```
eladdo92, line 137: is `key` needed?

src/components/List/List.tsx
```diff
@@ -81,18 +108,20 @@ const List: FC<ListProps> = forwardRef(
...
-                ref: ref => (childrenRefs.current[index] = ref),
-                tabIndex: focusIndex === index ? 0 : -1
-              });
+
+          return React.cloneElement(child, {
+            // @ts-ignore not sure how to deal with ref here
+            ref: ref => (childrenRefs.current[index] = ref),
+            tabIndex: focusIndex === index ? 0 : -1,
+            updateSelectedItem: updateSelectedItem,
+            index: index,
```
eladdo92, line 117:
```suggestion
            updateSelectedItem,
            index,
```

eladdo92 (approved):
awesome! 🤩

src/components/ListItem/ListItem.tsx
SergeyRoyt: Yeah, list elements should have a key attribute - was receiving a warning about it

src/components/List/List.tsx
SergeyRoyt: @shlomitc 
It can function, but the id(s) won't be unique, so it can lead to some troubles potentially
The logic is the following here:
- `List` doesn't have default id
- `ListItem` have default id, which is equal to `${listId}-item-${index}` - it's needed to set aria-activedescendants e.g.

So, if user don't pass `id` to `List`, he'll have following ListItem(s) ids:
`undefined-item-0`
`undefined-item-1`
...

And if user has more than 1 list rendered, elements ids won't be unique
So user should give an `id` to `List` and it all be good, but it's not a must
These ugly-looking ids (`undefined-item-0`) and a `console.warn` are supposed to help user to come to the right solution - give `List` an `id`

Tell me, if you have other suggestions on how to resolve it - we had the same issue with `Menu` in the past and eventually we came to this compromise solution with @orrgottlieb

src/components/ListItem/__tests__/__snapshots__/listItem.jest.js.snap
SergeyRoyt: No, cause the formula for `ListItem` id is `${listId}-item-${index}` and in this snapshot we are rendering just ListItem, without a `List`, so it doesn't receive `listId` or `index`

SergeyRoyt:
> I'll be happy to see some test that covers this logic as it is very hard to dive into the technical details of this PR

Added some a11y tests

src/components/List/List.tsx
shlomitc: why do we need to set `id`s at all?
if we don't have to - just don't put it.

if we must for some reason, then I'd at least remove the `undefined` which looks like a mistake

SergeyRoyt: @shlomitc  For `ListItem` u mean?
We should always have them, cause we need to set some `ListItem` `id` to `List` `aria-activedescendant`

src/constants/keyCodes.js
```diff
@@ -2,11 +2,12 @@ export const keyCodes = {
   ENTER: "Enter",
   SPACE: " ",
   ESCAPE: "Escape",
-  DOWN_ARROW: 40,
-  UP_ARROW: 38,
+  DOWN_ARROW: "ArrowDown",
+  UP_ARROW: "ArrowUp",
```
shlomitc, line 6: just making sure no one else is using this so it won't break

src/components/ListItem/__tests__/__snapshots__/listItem.jest.js.snap
```diff
@@ -1,71 +1,73 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`renders correctly with disabled 1`] = `
-<div
+<li
   aria-disabled={true}
   className="typography primary start singleLineEllipsis smallNormal listItem small disabled"
   data-testid="list-item"
+  id="list-0-item-undefined"
```
shlomitc, line 8: why do we still get undefined then?

src/components/ListItem/ListItem.tsx
```diff
@@ -58,6 +58,10 @@ export interface ListItemProps extends VibeComponentProps {
    */
   tabIndex?: number;
   "data-testid"?: string;
+  /**
+   * The index of the item in the list
+   */
+  index?: number;
```
shlomitc, line 64: just use the `id` prop and pass `${listId}-item-${index}` instead of introducing the `index` prop

src/components/List/utils/ListContext.ts
```diff
@@ -0,0 +1,20 @@
...
+export let listIdCounter = 0;
+
+export const generateListId = () => `list-${listIdCounter++}`;
+
+type ListContextType = {
+  /**
+   * A callback function which is being called when the item is being focused: by keyboard navigation or by mouse hover
+   * @param ListItem index
+   */
+  updateFocusedItem: (index: number) => void;
```
shlomitc, line 13: is this `index` or `id` that you get as argument?

src/components/List/__stories__/List.stories.mdx
shlomitc: consider removing or putting this in the storybook under a category of `Tests`

shlomitc (changes requested):

src/constants/keyCodes.js
SergeyRoyt: I've checked, yes

src/components/List/utils/ListContext.ts
SergeyRoyt: Changed it to `id`

src/components/ListItem/__tests__/__snapshots__/listItem.jest.js.snap
SergeyRoyt: Fixed

src/tests/constants.ts
```diff
@@ -94,8 +94,6 @@ export enum ComponentDefaultTestId {
   ICON = "icon",
   RESPONSIVE_LIST = "responsive-list",
   LIST = "list",
-  LIST_ITEM = "list-item",
-  LIST_TITLE = "list-title",
```
shlomitc, line 98: just make sure no one relies on this

src/components/ListItem/ListItem.tsx
```diff
@@ -108,21 +109,22 @@ const ListItem: FC<ListItemProps> & { sizes?: typeof SIZES } = forwardRef(
 
     return (
       <Text
-        element="div"
-        data-testid={dataTestId || getTestId(ComponentDefaultTestId.LIST_ITEM, id)}
+        element="li"
+        data-testid={dataTestId || id}
```
SergeyRoyt, line 113: So default `data-testid` won't be like this: `list-item_list-item-0`
We already have a solid default id, so I've decided it's better just to reuse it here

src/components/ListItem/ListItem.tsx
```diff
@@ -108,21 +109,22 @@ const ListItem: FC<ListItemProps> & { sizes?: typeof SIZES } = forwardRef(
...
         ref={mergedRef}
         className={cx(styles.listItem, className, getStyle(styles, camelCase(size)), {
           [styles.selected]: selected && !disabled,
           [styles.disabled]: disabled
         })}
         id={id}
+        key={id}
```
shlomitc, line 120: try moving it to the `List` component

shlomitc (changes requested):

shlomitc (approved):

src/tests/constants.ts
SergeyRoyt: Checked - no one is using

src/components/ListItem/ListItem.tsx
SergeyRoyt: Eventually removed it completely as I don't see this warning anymore and Text component seems like it's not passing this key anyway

PR #1444, opened 2023-07-31
fix(SplitButton): make menu-opening a11y-compatible

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4329030751


src/components/SplitButton/__stories__/SplitButtonExampleDialog.jsx
```diff
@@ -1,6 +1,7 @@
 import React from "react";
```
shlomitc, line 1: the documentation doesn't show this code, so it makes it hard for consumers to understand how to use the component.

See how it looks like
image

I suggest to kill this file and make it inline in the code example instead

shlomitc (changes requested):
Regarding the behavior itself:
1. when clicking on "Enter" to open the menu, the window scroll changes, which is weird.
2. When I'm focusing on an element and clicking "Enter", the item isn't selected nor the menu closes, are we sure this is the correct behavior?
3. Pressing "Escape" loses the focus completely, I'd expect the focus to come back to the split button

src/components/SplitButton/__stories__/SplitButtonExampleDialog.jsx
YossiSaadi: hmm good point.
will do

YossiSaadi:
> Regarding the behavior itself:
> 
> 1. when clicking on "Enter" to open the menu, the window scroll changes, which is weird.
> 2. When I'm focusing on an element and clicking "Enter", the item isn't selected nor the menu closes, are we sure this is the correct behavior?
> 3. Pressing "Escape" loses the focus completely, I'd expect the focus to come back to the split button

1. You're right, I've now checked without assigning the focusOnMount to it and it seems to happen because of the prop. Seems like a bug with this prop in menu. Will check deeper.
2. The same happens today if you click with the mouse. I believe this is the desired behavior, and the consumer should decide what he wants to happen, but TBH I'm not sure what is the current behavior if the menu is used as uncontrolled, I'll search for usages in the app to better understand (from looking at the code, most are implementing split button with uncontrolled menu)
3. I agree, same as previous bullet - it is the default behavior today (regardless to my changes). You want it to be handled on this PR? I'm not sure if it's a fix that should be done on `Menu` or on `SplitButton`

src/components/SplitButton/SplitButton.tsx
```diff
@@ -207,7 +207,6 @@ const SplitButton: FC<SplitButtonProps> & {
     <div
       className={classNames}
       ref={ref}
-      role="button"
```
hadasfa, line 210: talya request this?

src/components/SplitButton/SplitButtonMenu/__tests__/splitButtonMenu-tests.jest.tsx
```diff
@@ -0,0 +1,41 @@
...
+  );
+};
+
+describe("SplitButtonMenu", () => {
+  it("renders correctly", () => {
+    const { container } = renderSplitButtonMenu();
+    expect(container).toBeInTheDocument();
+  });
+
+  it("should pass the correct props to the underlying Menu component", () => {
```
hadasfa, line 26:
```suggestion
  it("should set first item as aria-activedescendant", () => {
```

is more accurate, no?

src/components/SplitButton/SplitButtonMenu/__tests__/splitButtonMenu-tests.jest.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const { getByTestId } = renderSplitButtonMenu();
+    expect(getByTestId(menuDataTestId)).toHaveAttribute("aria-activedescendant", `${id}-0`);
+  });
+
+  it("should render the children of Menu", () => {
```
hadasfa, line 31: can be a snapshot test, no?

src/components/SplitButton/SplitButtonMenu/__tests__/splitButtonMenu-tests.jest.tsx
```diff
@@ -0,0 +1,41 @@
...
+    const { getByText } = renderSplitButtonMenu();
+    expect(getByText("Test 1")).toBeInTheDocument();
+    expect(getByText("Test 2")).toBeInTheDocument();
+  });
+
+  it("should focus first menu item on mount", () => {
+    const { getByTestId } = renderSplitButtonMenu();
```
hadasfa, line 38: super cool :)

src/components/SplitButton/__stories__/SplitButtonExampleDialog.jsx
hadasfa: +1 :)

src/components/SplitButton/__tests__/splitButton-tests.jest.js
```diff
@@ -47,6 +60,14 @@ describe("SplitButton tests", () => {
     expect(expectedSecondaryDialog).toBeFalsy();
   });
 
+  it("should move between buttons with tab", () => {
```
hadasfa, line 63: awesome

src/components/SplitButton/__tests__/splitButton-tests.jest.js
```diff
@@ -1,16 +1,29 @@
 import React from "react";
+import "@testing-library/jest-dom";
```
hadasfa, line 2: I did not see it here (if I missed it let me know) - let's add snapshot with split button menu and a snapshot with custom child (for making sure we are not breaking this ability now or in the fututre)

src/components/Menu/Menu/Menu.tsx
```diff
@@ -28,7 +28,7 @@ import { getTestId } from "../../../tests/test-ids-utils";
 import { ComponentDefaultTestId } from "../../../tests/constants";
```
hadasfa, line 28: I did not see any updates in the docs, let's think about it in our sync

hadasfa (changes requested):
Awesome job! wrote some notes, let's talk about the changes we will do in the docs.

src/components/SplitButton/SplitButtonMenu/__tests__/splitButtonMenu-tests.jest.tsx
YossiSaadi: agreed, will do

src/components/Menu/Menu/Menu.tsx
YossiSaadi: @hadasfa Can you plz remind me what's the context here? 🙏🏼

src/components/SplitButton/SplitButtonMenu/__tests__/splitButtonMenu-snapshot-tests.jest.tsx
```diff
@@ -0,0 +1,27 @@
...
+const SplitButtonMenuWithItems = (
+  <SplitButtonMenu id="menu">
+    <MenuItem title="Test 1" />
+    <MenuItem title="Test 2" />
+  </SplitButtonMenu>
+);
+
+describe("SplitButtonMenu", () => {
+  beforeEach(() => {
+    mockRequestAnimationFrame();
```
hadasfa, line 16: what is this and what the purpose of this?

src/components/SplitButton/SplitButtonMenu/__tests__/test-utils.ts
```diff
@@ -0,0 +1,12 @@
+let rafSpy: jest.SpyInstance;
+
+export const mockRequestAnimationFrame = () => {
+  rafSpy = jest.spyOn(window, "requestAnimationFrame").mockImplementation((cb: FrameRequestCallback) => {
+    cb(0);
```
hadasfa, line 5: why 0?

src/components/SplitButton/__stories__/splitButton.stories.mdx
```diff
@@ -52,7 +54,18 @@ export const splitButtonTemplate = createComponentTemplate(SplitButton);
 A split button is a dual-function menu button that offers a default action as well as the possibility of choosing a secondary action, by selecting from a set of alternatives.
 
 <Canvas>
-  <Story name="Overview" args={{ children: "Button", secondaryDialogContent: () => SplitButtonExampleDialog({}) }}>
+  <Story
+    name="Overview"
+    args={{
+      children: "Button",
+      secondaryDialogContent: () => (
```
hadasfa, line 61: great

src/components/SplitButton/__stories__/splitButton.stories.mdx
```diff
@@ -190,10 +252,11 @@ The split button has supports multiple sizes: small, medium and large.
         ),
         description: (
           <>
-            Don’t use split button if there’s only one option within the menu. Use
+            Don’t use split button if there’s only one option within the menu. Use{" "}
             <Link href="/?path=/docs/buttons-button--overview" withoutSpacing>
```
hadasfa, line 256: maybe you can just remove without spacing

hadasfa (commented):
in general looking good, I will see the story in chromatic and then approve

src/components/SplitButton/SplitButtonMenu/__tests__/splitButtonMenu-snapshot-tests.jest.tsx
YossiSaadi: remember we spoke about adding requestAnimationFrame for the focus in the component?
Then we need to mock it in the tests to be able to test the focus (we need to wait for the next frame, after the focus is being done)

src/components/SplitButton/SplitButtonMenu/__tests__/test-utils.ts
YossiSaadi: typescript-wise - it need to get a number, so I'm just sending 0, doesn't really matter

src/components/SplitButton/__stories__/splitButton.stories.mdx
YossiSaadi: oh 😓

hadasfa (approved):
please just add the split button menu to the props before merge

PR #1448, opened 2023-08-01
fix: remove console log

hadasfa:


src/components/Typography/Typography.tsx
```diff
@@ -59,7 +59,6 @@ const Typography: VibeComponent<TypographyProps, HTMLElement> = forwardRef(
     },
     ref
   ) => {
-    console.log("!", htmlAttributes);
```
shlomitc, line 62: let's add a lint rule to fail the build in case such cases occur

PR #1450, opened 2023-08-01
chore: prevent from style injection to explode in server side rendering

shlomitc:
Solve the errors thrown
https://github.com/mondaycom/monday-ui-react-core/issues/1203#issuecomment-1658912659

Errors that appeared before:
image

### DIFF
--- a/rollup/styleInject.ejs
+++ b/rollup/styleInject.ejs
@@ -1,4 +1,5 @@
 function styleInject(css) {
+    if(!document) return;
     const head = document.head || document.getElementsByTagName('head')[0];
     const id = "<%= hashValue %>";
     const styleExists = head.querySelector("#<%= hashValue %>");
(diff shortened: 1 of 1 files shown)


YossiSaadi (approved):
does it have any affect on non-SSR?

shlomitc:
@YossiSaadi - shouldn't do any harm since it's just validating it actually exists, but do you think of other ways to break it?

YossiSaadi:
> @YossiSaadi - shouldn't do any harm since it's just validating it actually exists, but do you think of other ways to break it?

Hmm..
I actually guess it does not have any affect, otherwise we would already see it break if document did not exist

shlomitc:
@YossiSaadi - not sure we would see it since this script injection logic isn't used in the storybook

YossiSaadi:
> @YossiSaadi - not sure we would see it since this script injection logic isn't used in the storybook

unfortunately, I'm not familiar enough with this so I'll might suggest @hadasfa and @SergeyRoyt to also give their input on this one

PR #1452, opened 2023-08-02
feat(Modal): css changes, add new footer layout

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4569985599


src/components/Modal/ModalFooter/ModalFooterButtons/ModalFooterButtons.tsx
```diff
@@ -0,0 +1,50 @@
...
+  secondaryButtonText?: string;
+  onPrimaryButtonClick?: () => void;
+  onSecondaryButtonClick?: () => void;
+}
+
+const ModalFooterButtons: VibeComponent<ModalFooterButtonsProps> = forwardRef(
+  (
+    {
+      primaryButtonText = "Confirm",
+      secondaryButtonText = "Cancel",
```
shlomitc, line 22: since texts are always translated, I suggest removing those default props as it can cause confusion in terms of usage.

src/components/Modal/ModalFooter/ModalFooterButtons/ModalFooterButtons.tsx
```diff
@@ -0,0 +1,50 @@
...
+      onPrimaryButtonClick = NOOP,
+      onSecondaryButtonClick = NOOP,
```
shlomitc, line 24: I assume it will work also without assigning a default NOOP action

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -348,15 +301,15 @@ clicking outside. Use this variant in case of sensitive or important messages, a
...
           <ModalExampleWrapper bestPractice modalTitle="Modal title" buttonTitle="Click here">
-            <p>Modal with a proper CTA example</p>
+            Modal with a proper CTA example
           </ModalExampleWrapper>
         ),
         description: "Allow the user to easily close the modal by CTA and a close button."
       },
       negative: {
         component: (
           <ModalExampleWrapper bestPractice={false} hideCloseButton modalTitle="Modal title" buttonTitle="Click here">
```
shlomitc, line 311: image

shlomitc (changes requested):
minor stuff

src/components/Modal/__stories__/modal.stories.mdx
YossiSaadi: Niceee catch 😅 Thanks!

src/components/Modal/ModalFooter/ModalFooterButtons/ModalFooterButtons.tsx
YossiSaadi: good point 💪🏼

src/components/Modal/ModalFooter/ModalFooterButtons/ModalFooterButtons.tsx
YossiSaadi: I can just go with () => {} instead, would you prefer it instead? or to give up entirely on a default?
If we are giving up on a default entirely we'll need to add a check before calling the function in the `onClick`.

Btw, I know it can work without the NOOP, but I wanted to make sure I follow existing practices

YossiSaadi: @shlomitc pushed fixes for the other comments, lmk what option you prefer here

shlomitc: the button component should not how to manage it already, so you can just remove the defaults completely

shlomitc (approved):

PR #1465, opened 2023-08-07
feat: <TextField/> add event as a param to onChange callback

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4940675694

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -31,7 +31,7 @@ interface TextFieldProps extends VibeComponentProps {
   /** See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete for all of the available options */
   autoComplete?: string;
   value?: string;
-  onChange?: (value: string, event: React.ChangeEvent) => void;
+  onChange?: (value: string, event: Pick<React.ChangeEvent, "target">) => void;
   onBlur?: (event: React.FocusEvent) => void;
   onFocus?: (event: React.FocusEvent) => void;
   onKeyDown?: (event: React.KeyboardEvent) => void;
@@ -137,7 +137,9 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
     const inputRef = useRef(null);
     const { inputValue, onEventChanged, clearValue } = useDebounceEvent({
       delay: debounceRate,
-      onChange,
+      onChange: value => {
+        onChange(value, { target: inputRef.current });
+      },
       initialStateValue: value,
       trim
     });
@@ -158,12 +160,9 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
         if (inputRef.current) {
           inputRef.current.focus();
         }
+        // Do it cause otherwise the value is not cleared in target object
         inputRef.current.value = "";
-        const syntheticChangeEvent = {
-          type: "change",
-          target: inputRef.current
-        } as React.ChangeEvent<Partial<HTMLInputElement> | Partial<HTMLTextAreaElement>>;
-        clearValue(syntheticChangeEvent);
+        clearValue();
       }
       onIconClick(currentStateIconName);
     }, [disabled, clearOnIconClick, onIconClick, currentStateIconName, clearValue]);
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -31,7 +31,7 @@ interface TextFieldProps extends VibeComponentProps {
   /** See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete for all of the available options */
   autoComplete?: string;
   value?: string;
-  onChange?: (value: string, event: React.ChangeEvent | React.MouseEvent) => void;
+  onChange?: (value: string, event: React.ChangeEvent) => void;
   onBlur?: (event: React.FocusEvent) => void;
   onFocus?: (event: React.FocusEvent) => void;
   onKeyDown?: (event: React.KeyboardEvent) => void;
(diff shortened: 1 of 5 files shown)


src/components/TextField/__stories__/TextField.stories.mdx
```diff
@@ -71,6 +71,25 @@ An input field includes a label and a text field users can type text into. They
...
+        <div className="monday-storybook-text-field_column-wrapper">
+          <TextField iconName={Clear} clearOnIconClick onChange={onChange} placeholder="Small" name="small" />
+          <TextField onChange={onChange}  placeholder="Medium" name="medium" size={TextField.sizes.MEDIUM} />
+          <TextField onChange={onChange}  placeholder="Large" name="large" size={TextField.sizes.LARGE} />
+        </div>
+      );
+    }}
+  </Story>
+</Canvas>
+
```
SergeyRoyt, line 92: **TODO remove - demo story**

src/components/TextField/TextField.tsx
```diff
@@ -31,7 +31,7 @@ interface TextFieldProps extends VibeComponentProps {
   /** See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete for all of the available options */
   autoComplete?: string;
   value?: string;
-  onChange?: (value: string) => void;
+  onChange?: (value: string, event: React.ChangeEvent | React.MouseEvent) => void;
```
YossiSaadi, line 34: on which cases it can be sent as a mouse event?

src/components/TextField/TextField.tsx
```diff
@@ -158,10 +158,15 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
         if (inputRef.current) {
           inputRef.current.focus();
         }
-        clearValue();
+        inputRef.current.value = "";
```
YossiSaadi, line 161: is that intentional? it wasn't there before, I guess `clearValue` should handle it by itself on the returned `inputValue`?

src/components/TextField/TextField.tsx
```diff
@@ -158,10 +158,15 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
...
+        const syntheticChangeEvent = {
+          type: "change",
+          target: inputRef.current
```
YossiSaadi, line 164: btw, is this what Lior on ask-vibe meant? It sounded like he wanted to send a name to the `TextField` component and receive it back from the `onChange`

src/components/TextField/TextField.tsx
SergeyRoyt: Fixed this one

src/components/TextField/TextField.tsx
SergeyRoyt: Yeah, without it it was sending `target=inputRef.current` was containing not empty value, but the previous one

YossiSaadi: So what if you were to move the logic of contructing the event sent back on the debounce hook?

src/components/TextField/TextField.tsx
```diff
@@ -158,10 +158,15 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
...
+        } as React.ChangeEvent<Partial<HTMLInputElement> | Partial<HTMLTextAreaElement>>;
+        clearValue(syntheticChangeEvent);
       }
       onIconClick(currentStateIconName);
-    }, [clearValue, currentStateIconName, inputRef, clearOnIconClick, disabled, onIconClick]);
+    }, [disabled, clearOnIconClick, onIconClick, currentStateIconName, clearValue]);
```
YossiSaadi, line 169: your are suppose to be dependent on `inputRef` changes I believe

YossiSaadi:
Ok I have an idea
what if THE ONLY changes in this PR would be this:

change:
```
onChange?: (value: string) => void;
```
to:
```
onChange?: (value: string, event: Partial<React.ChangeEvent>) => void;
```
or even more specific:
```
onChange?: (value: string, event: Pick<React.ChangeEvent, "type" | "target">) => void;
```

AND manipulate the `onChange` being supplied to `useDebounceEvent`:
```
onChange,
```
to:
```
onChange: value => {
  onChange(value, { type: "change", target: inputRef.current });
},
```

This will also make it more consistent, so the consumer of `onChange` will know that the only fields he's getting back are the `type` and `target`, regardless to the event that occurs.

WDYT?

src/components/TextField/__tests__/textField-tests.jest.js
```diff
@@ -34,7 +36,16 @@ describe("TextField tests", () => {
     act(() => {
       fireEvent.change(input, { target: { value } });
     });
-    expect(onChangeStub).toHaveBeenCalledWith(value);
+
+    expect(onChangeStub).toHaveBeenCalledWith(
+      value,
+      expect.objectContaining({
```
YossiSaadi, line 42: this can be removed now I think

src/hooks/__tests__/useDebounceEvent.jest.ts
```diff
@@ -73,8 +73,11 @@ describe("useDebounceEvent", () => {
     it("should clear the value", () => {
       const { clearValue } = hookResult.result.current;
 
+      const syntheticChangeEvent = {
+        type: "change"
+      } as ChangeEvent<Partial<HTMLInputElement> | Partial<HTMLTextAreaElement>>;
       act(() => {
-        clearValue();
+        clearValue(syntheticChangeEvent);
       });
```
YossiSaadi, line 81: probably can be reverted? or am I missing something?

YossiSaadi (approved):
looks amazing 🚀

src/components/TextField/__tests__/textField-tests.jest.js
SergeyRoyt: Why? I think it's better to test it

src/hooks/__tests__/useDebounceEvent.jest.ts
SergeyRoyt: Yeah, just haven't pushed the changes after our talk - it's done now

PR #1472, opened 2023-08-10
fix(Menu): `focusIndexOnMount` can focus on an unfocusable item

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4961287518

There's a bug in `Menu` component where `focusIndexOnMount` can focus on `Title`/`Divider`, etc.


src/components/SplitButton/__stories__/splitButton.stories.mdx
```diff
@@ -344,6 +359,29 @@ When there’s already a primary button in the view, use a secondary split butto
   </Story>
 </Canvas>
 
+### Custom menu
+
+The split button can accept a custom Menu in case there's a need to override the default behavior.
+
+Notice to always include `focusItemIndexOnMount` prop for accessibility reasons when using custom menus.
```
SergeyRoyt, line 366: We don't have one standard for it, but I like it better when props names are referenced using `<code>` tag. You can try it out, but the decision is up to you

SergeyRoyt: We can maybe try to make it through `preview` file - to assign `<code>` tag to replace this `. But it's definitely out of scope here

SergeyRoyt:
Looks great!
The only thing I've found is that `focustItemIndexOnMount` is still counting other Menu children, though not allowing to focus on them. Not sure if it's important to fix it, but if we wanted to we could have filtered MenuItem's and use this index instead. Would probably look better regarding the API, but again not sure if it's worth the efforts

SergeyRoyt (approved):

YossiSaadi:
> Looks great!
> The only thing I've found is that `focustItemIndexOnMount` is still counting other Menu children, though not allowing to focus on them. Not sure if it's important to fix it, but if we wanted to we could have filtered MenuItem's and use this index instead. Would probably look better regarding the API, but again not sure if it's worth the efforts

Thanks man!
I thought about it, but then it might be confusing for someone who has let's say a title and two MenuItems. 
And he chooses focusOnMount=1.
He would have expect, I guess, the first MenuItem to be selected.
But if we filter menuItems, it would select the first menu item.
Wdyt? Is that a scenario that makes sense?

SergeyRoyt:
> Thanks man! I thought about it, but then it might be confusing for someone who has let's say a title and two MenuItems. And he chooses focusOnMount=1. He would have expect, I guess, the first MenuItem to be selected. But if we filter menuItems, it would select the first menu item. Wdyt? Is that a scenario that makes sense?

Scenario makes sense, but I'd probably expect first item to be chosen here)
Anyway I'd say it's up to you to decide here) Both options have a right to exist
I'd only vouch for adding this decision to the prop description

Btw, let's fix the controls of the `Menu` story, while we are here? You will only need to change 
`<ArgsTable of={Menu} />` to `<ArgsTable story="Overview" />`

src/components/SplitButton/__stories__/splitButton.stories.mdx
YossiSaadi: I think they are rendered the same way in mdx (also on md in general I think) (meaning `hello` is converted to `<code>hello</code>`)

YossiSaadi: yep

image

PR #1485, opened 2023-08-16
fix: <List> id generation

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4999466126

Now run only once instead of running on each rerender

### DIFF
--- a/src/components/List/utils/ListUtils.ts
+++ b/src/components/List/utils/ListUtils.ts
@@ -1,12 +1,17 @@
-import { MutableRefObject, useState } from "react";
+import { MutableRefObject, useEffect, useState } from "react";
 import { ListWrapperComponentStringType, ListWrapperComponentType } from "../ListConstants";
 import { ListItemComponentType } from "../../ListItem/ListItemConstants";
 
 let listIdCounter = 0;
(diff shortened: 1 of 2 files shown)


src/components/List/utils/ListUtils.ts
```diff
@@ -1,10 +1,15 @@
...
 import { ListWrapperComponentStringType, ListWrapperComponentType } from "../ListConstants";
 import { ListItemComponentType } from "../../ListItem/ListItemConstants";
 
 let listIdCounter = 0;
 export const generateListId = () => `list-${listIdCounter++}`;
 
+export const useListId = (id: string) => {
+  const [listId] = useState(id || generateListId());
+  return listId;
+};
```
YossiSaadi, line 11: Why does it using state? Can't it be a regular util function?

SergeyRoyt: Cause if it will be regular function it will run on each rerender and return a new id each time

SergeyRoyt: Cause if it will be regular function it will run on each rerender and return a new id each time

hadasfa (approved):

PR #1498, opened 2023-08-22
fix: <AlertBannerLink> pass textColor

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5030142278


hadasfa (approved):
nice!

PR #1511, opened 2023-08-28
fix(Dropdown): when used inside overflowed container with transform, consider container scroll

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4990237068

### DIFF
--- a/src/components/Dropdown/Dropdown.jsx
+++ b/src/components/Dropdown/Dropdown.jsx
@@ -74,7 +74,6 @@ const Dropdown = forwardRef(
       isOptionSelected,
       insideOverflowContainer,
       insideOverflowWithTransformContainer,
-      containerRef,
       tooltipContent,
       onKeyDown,
       isLoading,
@@ -128,7 +127,6 @@ const Dropdown = forwardRef(
         rtl,
         insideOverflowContainer,
         controlRef,
-        containerRef,
         insideOverflowWithTransformContainer
       });
 
@@ -617,10 +615,6 @@ Dropdown.propTypes = {
    * For display the drop down menu in overflow hidden/scroll container which contains transform css function usage.
    */
   insideOverflowWithTransformContainer: PropTypes.bool,
-  /**
-   * For when insideOverflowWithTransformContainer is used, this is the ref of the container dialog.
-   */
-  containerRef: PropTypes.func,
   /**
    * When content is passed, the dropdown will include a tooltip on the dropdown's value.
    */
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.jsx
```diff
@@ -615,6 +617,10 @@ Dropdown.propTypes = {
    * For display the drop down menu in overflow hidden/scroll container which contains transform css function usage.
    */
   insideOverflowWithTransformContainer: PropTypes.bool,
+  /**
+   * For when insideOverflowWithTransformContainer is used, this is the ref of the container dialog.
+   */
+  containerRef: PropTypes.func,
```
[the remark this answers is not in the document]
YossiSaadi: Thanks for the suggestions!
I pushed a fix to independently find the scrollable parent, without passing the `containerRef`.
I had 2nd thoughts after we spoke and I think we might not want to allow user to "stick" the menu to things other than the dropdown:
1. It might break how we meant the component to render
2. There's not necessarily a use case where user would like to "stick" it to stuff other than the dropdown

With saying that, we should really consider - on another dropdown iteration - making this behaviors by default without hacks as this prop. I believe this is something that should be achievable

src/components/Dropdown/Dropdown.styles.js
```diff
@@ -315,8 +316,10 @@ const menu =
      * start when the menu position is fixed, and this is why in this case we define top:auto.
      */
     let top = insideOverflowWithTransformContainer ? "auto" : parentPositionData.bottom;
+    const translateY = getScrollableParent(controlRef?.current)?.scrollTop;
```
shlomitc, line 319: 1. run it only when `insideOverflowWithTransformContainer` is applied
2. when applying the scrollable container logic, the `insideOverflowWithTransformContainer` can simply apply `insideOverflowContainer` instead of applying the transform logic
3. in the example - make the dropdown scrollable in the middle (to explain scrolling before opening)

src/storybook/patterns/dropdown-in-modals/dropdown-in-modals.stories.mdx
```diff
@@ -90,7 +90,9 @@ overflow container has one parent or more with conflicting overflow definitions.
       return (
         <Flex gap={Flex.gaps.LARGE}>
           <DialogContentContainer style={dialogStyle}>
+            <div style={{ height: "100px" }} />
```
hadasfa, line 93: this for testing purpose or we want to keep it in the docs?

YossiSaadi:
@shlomitc after some discussions and experiences with Hadas, we believe that (currently) we cannot merge both usages, as because of the transform it requires different `top` value.
We've spent some time on it last week and I believe that since this area is a black-box for the consumer, and it does not require any additional thing from it, we can tackle it later as part of the new react-select migration (it might even be solved there by design)

shlomitc (approved):

YossiSaadi:
for future reference: we need to think of a better solution to not ask the consumer use one of those props
the "magic" should happen behind the scenes

PR #1514, opened 2023-08-31
feat(EmptyState): new component

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5075443870


src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+import styles from "./EmptyState.module.scss";
+import Flex from "../Flex/Flex";
+import Heading from "../Heading/Heading";
+import Text from "../Text/Text";
+import Button from "../Button/Button";
+
+export interface EmptyStateProps extends VibeComponentProps {
+  imgSrc: string;
+  title: string;
+  body: string;
```
SergeyRoyt, line 17: Not sure if `body` is the best name for it... Maybe just `text`, `message` or smth else?

src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onPrimaryActionClick?: () => void;
+  primaryActionLabel?: string;
+  onSecondaryActionClick?: () => void;
+  secondaryActionLabel?: string;
+  imgClassName?: string;
```
SergeyRoyt, line 22: Some description comments would be nice

src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+        align={Flex.align.CENTER}
+        justify={Flex.justify.CENTER}
+        gap={Flex.gaps.LARGE}
+        id={id}
+        data-testid={dataTestId || getTestId(ComponentDefaultTestId.EMPTY_STATE, id)}
+      >
+        <img src={imgSrc} alt={title} className={cx(styles.image, imgClassName)} />
+        <Flex direction={Flex.directions.COLUMN} gap={Flex.gaps.SMALL} align={Flex.align.CENTER}>
+          <Heading type={Heading.types.H2}>{title}</Heading>
+          <Text type={Text.types.TEXT1}>{body}</Text>
```
SergeyRoyt, line 59: Always one line here by design?

SergeyRoyt: I have a concern about what's gonna happen if people put a lot of text here - attaching screenshots to demonstrate
image
image
image

We should think about some max-width limitations, so the text will just go ellipsis with a tooltip when it's reaching the limits

src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+        </Flex>
+        <Flex gap={Flex.gaps.SMALL} align={Flex.align.CENTER}>
+          {secondaryActionLabel && (
+            <Button kind={Button.kinds.TERTIARY} onClick={onSecondaryActionClick}>
+              {secondaryActionLabel}
+            </Button>
+          )}
+          {primaryActionLabel && <Button onClick={onPrimaryActionClick}>{primaryActionLabel}</Button>}
```
SergeyRoyt, line 67: Nice, but people might use secondaryButton without the primary one, which won't be good. Perhaps, we can protect from this from the beginning

SergeyRoyt:
Maybe also worth adding RelatedComponents, if u can think of any - just to be consistent

SergeyRoyt:
I'd also consider what's needed to be done here (same story in LegacyHeading also)
image

Maybe we can just add some text and reference the new component there

src/components/EmptyState/EmptyState.tsx
shlomitc: content?

src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { getTestId } from "../../tests/test-ids-utils";
+import { ComponentDefaultTestId } from "../../tests/constants";
+import styles from "./EmptyState.module.scss";
+import Flex from "../Flex/Flex";
+import Heading from "../Heading/Heading";
+import Text from "../Text/Text";
+import Button from "../Button/Button";
+
+export interface EmptyStateProps extends VibeComponentProps {
+  imgSrc: string;
```
shlomitc, line 15: an empty state would also need support `SVG`, `Lottie` and maybe at some point different assets. why not keep this as a general node? this would also make `imgClassName` redundant

src/components/EmptyState/EmptyState.tsx
shlomitc, line 59: general suggestion - apply `Heading` and `Text` only when `string` type is passed, so in cases where other content will be passed, it will not always be wrapped with the typography component constaraints

shlomitc, line 59: this would apply ellipsis by default if I understand correctly, not sure this is the expected behavior for the empty state

YossiSaadi:
> Maybe also worth adding RelatedComponents, if u can think of any - just to be consistent

great suggestion, for the meantime Yael and I agreed that since we need to "run" with it quick, the story would be very minimal

src/components/EmptyState/EmptyState.tsx
```diff
@@ -0,0 +1,74 @@
...
+  title: string;
+  body: string;
+  onPrimaryActionClick?: () => void;
+  primaryActionLabel?: string;
```
shlomitc, line 19: I like the abstraction to have very simple actionable button props, but take into consideration that if in the future some a11y props will be needed, we might want to pass the entire props easily `primaryActionProps`

shlomitc (changes requested):
worth adding also an example in the `Table` component page and see it in a real use case

YossiSaadi:
@shlomitc @SergeyRoyt you had great suggestions, I'll go over the content/design-related also with Yael/Meirav
Thanks! :)

talkor:
@YossiSaadi can be closed, right?

YossiSaadi:
> @YossiSaadi can be closed, right?

yep

PR #1521, opened 2023-09-03
chore(browserslist): Use browserslist-config-monday on Vibe

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5097839339

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,7 @@
     "a11y-dialog": "^7.5.2",
     "autosize": "^5.0.1",
     "body-scroll-lock": "^4.0.0-beta.0",
+    "browserslist-config-monday": "^1.0.6",
     "classnames": "^2.3.2",
     "lodash-es": "^4.17.21",
     "monday-ui-style": "0.1.186",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -249,16 +250,7 @@
...
-    "Firefox ESR",
-    "not IE 11",
-    "Edge 18",
-    "not dead",
-    "not Samsung > 0",
-    "not op_mini all",
-    "not and_ff > 0",
-    "not and_qq > 0",
-    "not and_uc > 0"
+    "extends browserslist-config-monday"
```
orrgottlieb, line 253: this is a breaking change....

shlomitc: If I understand correctly - this is the exact same configuration we are using in the Monday production, so it's basically just an alignment

YossiSaadi: @orrgottlieb I agree with @shlomitc
It means this is a breaking change only for consumers that aren't using vibe to build monday apps

shlomitc (approved):

package.json
orrgottlieb: this is still an open source library, which we have other users... @shlomitc @YossiSaadi 

open source is open source, you should go by the community standards for semver

PR #1526, opened 2023-09-05
feat: <ThemeProvider>

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5065724987


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,67 @@
...
+      document.head.appendChild(styleElement);
+    } catch (error) {
+      console.error("Error loading dynamic CSS:", error);
+    }
+
+    // Cleanup: Remove the style element when the component unmounts
+    return () => {
+      document.head.removeChild(styleElement);
+    };
+  }, [theme]);
```
SergeyRoyt, line 59: Issue with that - that's not working synchronously. Meaning when u refresh u first see unthemed version and then in half of the second themed one
Didn't yet understand how to solve it

src/components/ThemeProvider/__stories__/ThemeProvider.stories.mdx
```diff
@@ -0,0 +1,79 @@
...
+            [Color.primaryColor]: "green"
+          },
+          [SystemTheme.DARK]: {
+            [Color.primaryColor]: "red"
+          },
+          black: {
+            "primary-color": "orange"
+          },
+          hacker: {
+            "primary-color": "yellow"
```
SergeyRoyt, line 53: Showing that both versions are available: enum with autocomplete or just a value. It's type safe to use both

src/components/ThemeProvider/__stories__/ThemeProvider.stories.mdx
```diff
@@ -0,0 +1,166 @@
...
+      <code>body</code> and render everything else inside it
+    </>,
+    <>
+      In most common case ThemeProvider should be rendered only once on the root level of the application - below the{" "}
+      <code>body</code>
+    </>,
+    <>
+      ThemeProvider is populating theme name <code>className</code> to it's child, so don't put{" "}
+      <code>{"\<Fragment\>"}</code> (<code>{"\<\>"}</code>) inside - it's not accepting <code>className</code> prop
+    </>
```
SergeyRoyt, line 98: Currently there are problems with displaying this - @talkor will take care to fix it UsageGuidelines component styles

src/components/ThemeProvider/__tests__/themeProvider-tests.jest.js
```diff
@@ -0,0 +1,102 @@
+import { generateThemeCssOverride } from "../ThemeProviderUtils";
+
+const THEME_NAME = "test-theme-name";
+
+describe("ThemeProvider tests", () => {
+  it("theme with no colors", () => {
+    expect(true).toBe(true);
```
shlomitc, line 7:
```suggestion

```

src/components/ThemeProvider/__tests__/themeProvider-tests.jest.js
```diff
@@ -0,0 +1,102 @@
...
+          "primary-color": "#d14900"
+        }
+      }
+    };
+    const css = generateThemeCssOverride(theme);
+    expect(css).toBe(".light-app-theme .test-theme-name {--primary-color: #d14900;}\n");
+  });
+
+  it("basic theme", () => {
+    expect(true).toBe(true);
```
shlomitc, line 40:
```suggestion

```

src/components/ThemeProvider/__tests__/themeProvider-tests.jest.js
```diff
@@ -0,0 +1,102 @@
...
+    const theme = {
+      name: THEME_NAME
+    };
+    const css = generateThemeCssOverride(theme);
+    expect(css).toBeNull();
+  });
+
+  it("theme with empty colors", () => {
+    expect(true).toBe(true);
```
shlomitc, line 16:
```suggestion

```

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    try {
+      // Set the innerText property of the style element with the dynamic CSS
+      styleElement.appendChild(document.createTextNode(themeCssOverride));
+
+      // Append the style element to the document head
+      document.head.appendChild(styleElement);
+      setStylesLoaded(true);
+    } catch (error) {
+      console.error("Error loading dynamic CSS:", error);
```
shlomitc, line 37: that's actually not loading dynamic CSS but rather generating a theme.
And maybe we can add some prefix like `"vibe: "` or something similar so we can monitor such errors if they get caught sentry

src/components/ThemeProvider/ThemeProviderUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (!theme.colors) {
+    return null;
+  }
+
+  let css = "";
+  for (const systemTheme of Object.keys(theme.colors) as SystemTheme[]) {
+    css += generateCss(theme.colors[systemTheme], "", `.${SystemThemeClassMap[systemTheme]} .${theme.name}`) + "\n";
+  }
+
+  console.log("### css", css);
```
shlomitc, line 35:
```suggestion

```

src/components/ThemeProvider/__tests__/themeProvider-snapshot-tests.jest.js
```diff
@@ -0,0 +1,22 @@
...
+        <ThemeProvider
+          theme={{
+            name: THEME_NAME
+          }}
+        >
+          <div />
+        </ThemeProvider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
shlomitc, line 20: I think we can ditch the snapshot testing. we use chromatic which is much stronger, and besides it will always render `null`.

Instead, you should render the component and make sure the original children are eventually rendered there

src/components/ThemeProvider/ThemeProviderConstants.ts
```diff
@@ -0,0 +1,51 @@
...
+  textColorOnPrimary = "text-color-on-primary",
+  brandColor = "brand-color",
+  brandHoverColor = "brand-hover-color",
+  brandSelectedColor = "brand-selected-color",
+  brandSelectedHoverColor = "brand-selected-hover-color",
+  textColorOnBrand = "text-color-on-brand"
+}
+
+export type Theme = {
+  name: string;
```
shlomitc, line 24: I'm missing the documentation of this property and testing what happens if not passed

src/components/ThemeProvider/__stories__/ThemeProvider.stories.tsx
```diff
@@ -0,0 +1,179 @@
...
+        <Button>Themed</Button>
+      </ThemeProvider>
+    </>
+  );
+};
+
+export const ThemeProviderThemingScopeTemplate = (_args: JSX.IntrinsicAttributes & ThemeProviderProps) => {
+  return (
+    <>
+      <ThemeProvider
```
shlomitc, line 29: This happens when I clear the theme via the storybook toggle. which means we might need to consider a fallback regarding how to handle the case a specific theme is not applied
image

src/components/ThemeProvider/__stories__/ThemeProvider.stories.mdx
```diff
@@ -0,0 +1,166 @@
...
+<Canvas>
+  <Story name="Folded theming">{ThemeProviderFoldedThemingTemplate.bind({})}</Story>
+</Canvas>
+
+### Product theming
+
+These are theme-definitions, which are used in monday.com products.
+
+<Canvas>
+  <Story name="Product theming">{ThemeProviderProductThemingTemplate.bind({})}</Story>
```
shlomitc, line 140: image
not working

shlomitc (changes requested):

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+    styleElement.type = "text/css";
+    styleElement.id = theme.name;
+    const themeCssOverride = generateThemeCssOverride(theme);
+
+    try {
+      // Set the innerText property of the style element with the dynamic CSS
+      styleElement.appendChild(document.createTextNode(themeCssOverride));
+
+      // Append the style element to the document head
+      document.head.appendChild(styleElement);
```
orrgottlieb, line 34: Major major major performance hits, please please don't use it like this.

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  useEffect(() => {
+    if (!theme) {
+      return;
+    }
+
+    // Create a new style element
+    const styleElement = document.createElement("style");
+    styleElement.type = "text/css";
+    styleElement.id = theme.name;
```
orrgottlieb, line 26: Name must be unique...., please enforce it

src/components/ThemeProvider/ThemeProvider.tsx
SergeyRoyt: @orrgottlieb  You have any ideas how to do it in an other way?

src/components/ThemeProvider/__stories__/ThemeProvider.stories.mdx
SergeyRoyt: Fixed

src/components/ThemeProvider/__tests__/themeProvider-snapshot-tests.jest.js
SergeyRoyt: I think it'll come to us via sentry anyway - cause it's in our component
But I'll add it, np

src/components/ThemeProvider/ThemeProvider.tsx
SergeyRoyt: done

src/components/ThemeProvider/__stories__/ThemeProvider.stories.mdx
talkor: Fixed

src/components/ThemeProvider/ThemeProviderConstants.ts
SergeyRoyt: Added docs and tests

src/components/ThemeProvider/__stories__/ThemeProvider.stories.tsx
SergeyRoyt: What kind of fallback you are thinking about? When systemTheme is not defined, ThemeProvider sysyemTheme dependent color overrides are not applying - that's the design
In the Monolith PR, though, I've implemented a separate logic, which works for default-system-theme as well

src/components/ThemeProvider/ThemeProvider.tsx
orrgottlieb: i have multiple approaches

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+   */
+  theme?: Theme;
+  /**
+   * The children to render with the theme
+   */
+  children: ReactElement;
+  /**
+   * Random string which adds up to theme name to make it more specific (in case if theme.name is colliding with some other class name)
+   */
+  randomStringSelector?: string;
```
SergeyRoyt, line 19: Do we want this prop, @shlomitc?
Apart from the tests, I guess I can imagine cases where it might be useful, though don't really like to garbage ThemeProviderProps with it... WDYT?

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+  colors?: typeof ThemeColor;
+} = ({ theme, children, randomStringSelector: customRandomStringSelector }) => {
+  const [stylesLoaded, setStylesLoaded] = useState(false);
+  // Using random string selector to secure selector by making it more specific - if theme name is not unique
+  const randomStringSelector = useState(customRandomStringSelector || generateRandomAlphaString())[0];
+
+  useEffect(() => {
+    if (!theme || !theme?.name) {
+      return;
+    }
```
shlomitc, line 33: no value means not rendering anything? why not just render the default `light` theme?

src/components/ThemeProvider/ThemeProvider.tsx
shlomitc: 1. it's not random, so the name might be misleading
2. if not mandatory, let's start without it and add it only when needed

src/components/ThemeProvider/ThemeProvider.tsx
SergeyRoyt: Changed it to 
```
    if (!shouldGenerateTheme(theme)) {
      return;
    }
```
If the theme is not defined correctly then this useEffect shouldn't do anything - and user will get just default color according to the theme he has

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+  randomStringSelector?: string;
+}
+
+const ThemeProvider: FC<ThemeProviderProps> & {
+  systemThemes?: typeof SystemTheme;
+  colors?: typeof ThemeColor;
+} = ({ theme, children, randomStringSelector: customRandomStringSelector }) => {
+  const [stylesLoaded, setStylesLoaded] = useState(false);
+  // Using random string selector to secure selector by making it more specific - if theme name is not unique
+  const randomStringSelector = useState(customRandomStringSelector || generateRandomAlphaString())[0];
```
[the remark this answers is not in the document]
SergeyRoyt: @orrgottlieb wanted to add this to be more secure in cases where e.g. when theme.name is the same as some other class name - so it would still work
Let's rename it to `additionalStringSelector`? wdyt @shlomitc ?

src/components/ThemeProvider/ThemeProvider.tsx
SergeyRoyt: @orrgottlieb wanted to add this to be more secure in cases where e.g. when theme.name is the same as some other class name - so it would still work
Let's rename it to `additionalStringSelector`? wdyt @shlomitc ?

src/components/ThemeProvider/ThemeProvider.tsx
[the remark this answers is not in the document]
SergeyRoyt: Done

src/components/ThemeProvider/ThemeProvider.tsx
shlomitc: 1. `additionalStringSelector` still sounds a bit vague, maybe `specifyThemeClass` or something that can explain it without looking at the prop description
2. if we already provide the way to control it via `theme.name`, why do we need an additional prop for that?

SergeyRoyt: 1. Renamed it to `themeClassSpecifier`
2. `theme.name` might not be unique and collide with other classes leading to issues... This additional `specifier` adds by default is random string, which makes sure that css selector is unique. We are allowing to provide it as a prop, so it won't hurt e.g. tests and this kind of thing - to grant consistency, if it's needed. for any reason

shlomitc (approved):

PR #1527, opened 2023-09-05
accessibility: add aria attributes to button, list & list-item

eladdo92:
Add aria attributes to button, list & list-item

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/button-has-type */
-import React, { forwardRef, useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, { AriaAttributes, forwardRef, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { camelCase } from "lodash-es";
 import cx from "classnames";
 import { SIZES } from "../../constants";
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.tsx
```diff
@@ -22,76 +22,77 @@ import styles from "./Button.module.scss";
...
-  defaultTextColorOnPrimaryColor?: string;
-  dataTestId?: string;
-  "data-testid"?: string;
-  /** Change the focus indicator from around the button to within it */
-  insetFocus?: boolean;
-  /** Specifies the tab order of an element */
-  tabIndex?: number;
-}
+export type ButtonProps = VibeComponentProps &
+  AriaAttributes & {
```
YossiSaadi, line 26: @eladdo92 I'm not sure we should allow all arias for Button

src/components/List/List.tsx
```diff
@@ -20,26 +29,27 @@ import {
...
-  ariaDescribedBy?: string;
-  children?: ReactElement<ListItemProps | ListTitleProps> | ReactElement<ListItemProps | ListTitleProps>[];
-  /**
-   * Using virtualized list for rendering only the items which visible to the user in any given user (performance optimization)
-   */
-  renderOnlyVisibleItems?: boolean;
-  style?: CSSProperties;
-}
+export type ListProps = VibeComponentProps &
+  AriaAttributes & {
```
YossiSaadi, line 33: Same, not sure we should allow all

src/components/ListItem/ListItem.tsx
```diff
@@ -12,57 +12,58 @@ import { ListContext } from "../List/utils/ListContext";
...
-   * Selected indication
-   */
-  selected?: boolean;
-  /**
-   * The size of the list item
-   */
-  size?: typeof SIZES[keyof typeof SIZES];
-  /**
+export type ListItemProps = VibeComponentProps &
+  AriaAttributes & {
```
YossiSaadi, line 16: Same

YossiSaadi (commented):
@eladdo92 By extending aria on all three we are allowing arias that not necessarily should be allowed.
Might be better to Pick the ones needed

src/components/Button/Button.tsx
shlomitc: @YossiSaadi - any good reason for not opening all arias? if not, I let's just allow all

YossiSaadi: @shlomitc Button for example should get stuff like "aria-pressed", it shouldn't get props like "aria-activedescendant" or "aria-checked", and this change would allow it

I've read once on W3C that a bad aria is worse than no aria, if we'll want to revert it would be impossible, as this would be a breaking change..

oribarak: @shlomitc @YossiSaadi any conclusion? We need to move forward with this

YossiSaadi: @oribarak I've DMed @eladdo92 about it
We were wondering if you know the exact arias needed for now, as a temp solution, and from our side we will make sure to handle it asap to be able to pass all

YossiSaadi:
After a talk to @eladdo92 to understand exactly what needed, because we can't allow all arias for all components, we decided to allow only the needed props for now.

resolved at #1535, as part of ticket: https://monday.monday.com/boards/3532714909/pulses/5121462835

PR #1529, opened 2023-09-05
docs: Change <LegacyHeading> name in the story

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5070102183

### DIFF
--- a/src/components/LegacyHeading/__stories__/LegacyHeading.stories.mdx
+++ b/src/components/LegacyHeading/__stories__/LegacyHeading.stories.mdx
@@ -1,7 +1,7 @@
 import Heading from "../LegacyHeading";
 import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
 import { createStoryMetaSettingsDecorator, DeprecatedWarning } from "../../../storybook";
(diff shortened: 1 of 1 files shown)


src/components/LegacyHeading/__stories__/LegacyHeading.stories.mdx
```diff
@@ -52,8 +52,8 @@ export const headingTemplate = createComponentTemplate(Heading);
 Heading components are used to title any page sections or sub-sections in top-level page sections.
 
 <Canvas>
-  <Story name="Overview" args={{ value: "Hello world" }}>
-    {headingTemplate.bind({})}
+  <Story name="Overview" args={{}}>
+    <Heading value="Hello world" />
```
YossiSaadi, line 56: It means that we can't play with the component in Storybook, isn't it?

SergeyRoyt: Yeah, unfortunately :(
But I wasn't able to remove LegacyHeading from the show code other way

YossiSaadi: got ya
actually it might not be that important in a legacy one

PR #1537, opened 2023-09-07
feat(Tipseen): in floating variation do not show tip, make docs for variation clearer

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5121927498


YossiSaadi:
@SergeyRoyt added a commit with a test after your approve (FYI)

PR #1540, opened 2023-09-11
fix(TipseenTitle): font size should be TEXT1 (16px)

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5127607534


SergeyRoyt (approved):
Need to update the snapshots

PR #1542, opened 2023-09-12
feat(Table): table leftovers

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5134983370


src/components/Table/Table/Table.tsx
```diff
@@ -1,18 +1,22 @@
...
 import { ComponentDefaultTestId } from "../../../tests/constants";
 
+export type TableContentSkeletonType = "long-text" | "medium-text" | "circle" | "rectangle";
+
 export interface ITableColumn {
   id: string;
   title: string;
   infoContent?: string;
   width?: number | { min: number; max: number };
+  icon?: SubIcon;
```
shlomitc, line 18: I'm missing a code example. the Do and Don't don't show how it is used
image

src/components/Table/Table/__stories__/table.stories.module.scss
```diff
@@ -0,0 +1,15 @@
...
+.virtualizedScrollTable {
+  display: flex;
+  flex-direction: column;
+  height: 250px;
+  flex: 1;
+}
+
+.iconRule {
+  width: 95%;
+}
```
shlomitc, line 15: since classnames aren't available in the story (and aren't copy-pastable), they should be inlined.
I'm also not sure why `.virtualizedScrollTable` and `. scrollTable` should be set manually and not built-in in the table

src/components/Table/TableHeader/TableHeader.module.scss
```diff
@@ -1,12 +1,12 @@
 .tableHeader {
   display: grid;
   grid-template-columns: var(--table-grid-template-columns);
+  position: sticky;
+  top: 0;
```
shlomitc, line 5: Is this always the case or only in certain scrolling logic (like full page scroll)?

src/components/Table/TableHeader/TableHeader.module.scss
```diff
@@ -1,12 +1,12 @@
...
+  z-index: 1;
```
shlomitc, line 6: z-index always makes me shake a little bit. I'm just making sure this is necessary.

src/components/Table/TableHeader/TableHeader.module.scss
```diff
@@ -1,12 +1,12 @@
...
+  background-color: inherit;
```
shlomitc, line 7: isn't inherit the default behavior?

src/components/Table/Table/__stories__/helpers.js
```diff
@@ -32,16 +82,127 @@ export const emailColumns = [
   {
     id: "sentOn",
     title: "Sent on",
-    width: 150
+    width: 150,
+    contentSkeleton: "long-text"
```
shlomitc, line 86: this can't be seen in the story, which makes it hard to understand how to use
image

src/components/Table/Table/__stories__/helpers.js
```diff
@@ -32,16 +82,127 @@ export const emailColumns = [
...
   },
   {
     id: "subject",
-    title: "Subject"
+    title: "Subject",
+    contentSkeleton: ["long-text", "medium-text"]
```
shlomitc, line 91: why both are needed? I think it's enough to state `medium-text` and then internally decide on long and medium based on the row index

src/components/Table/Table/Table.tsx
```diff
@@ -1,18 +1,22 @@
...
+  contentSkeleton?: TableContentSkeletonType | TableContentSkeletonType[];
```
shlomitc, line 19: name-wize, I'd use something that indicates `loading`. the fact that it's currently a skeleton doesn't really matter.
I suggested `loadingStateType` in the [API RFC](https://github.com/mondaycom/monday-ui-react-core/issues/1471)

shlomitc (changes requested):

src/components/Table/Table/Table.tsx
YossiSaadi: @shlomitc I'll talk to Yael for maybe adding a story for this case

src/components/Table/Table/__stories__/table.stories.mdx
```diff
@@ -145,43 +157,89 @@ Using skeleton
...
+Table with both vertical and horizontal scroll
 
 <Canvas>
   <Story name="Scroll">
+    {
+      {
+        /* styles.scrollTable = height: 220px; width: 800px to enforce scrolling container */
+      }
+    }
+    <TableContainer className={styles.scrollTable}>
```
shlomitc, line 203: I see that this is the only place that we're actually placing something on the `TableContainer` container.
All the rest are empty components such as `<TableContainer> ... </TableContainer>`

Are we sure we need to wrap the Table with another component?

shlomitc (changes requested):

src/components/Table/Table/__stories__/table.stories.mdx
YossiSaadi: If we're not wrapping it, and we're styling Table (and assigning it overflow) straight, the Table cannot fit under no container.
I checked with more libraries and didn't find one that do not wrap it inside their design system.

The reason is that the overflow with border-radius and others should be on the wrapper, and if we're not wrapping it ourselves we are asking people to implement it themselves

src/components/Table/Table/Table.tsx
YossiSaadi: wasn't aware of the RFC

src/components/Table/TableHeader/TableHeader.module.scss
YossiSaadi: As far as I understand the header should always stick to top (we can opt-out with a prop in case a demand will arise)

src/components/Table/TableHeader/TableHeader.module.scss
YossiSaadi: Since we're sticking the header to top, and this z-index is stacked inside the table header, I don't think we have a lot to be worried here (+ I've used the minimum)

src/components/Table/TableHeader/TableHeader.module.scss
YossiSaadi: nope, defaults to transparent (makes sense, it is the broader use case, since otherwise we might have to configure transparent background for each subelement where we define background for parent)

src/components/Table/Table/__stories__/helpers.js
YossiSaadi: hmm I get what you mean.
We need to think of a technical solution for that, maybe add some "assets" feature to stories for cases like this where we want to show api usage.
I'll think of a median solution for now

src/components/Table/Table/__stories__/helpers.js
YossiSaadi: I thought about it, but this allow more flexibility. you prefer the way where if user chooses medium/long we are alternating between them?

src/components/Table/Table/__stories__/helpers.js
shlomitc: I'd just keep the data inside the render function

src/components/Table/Table/__stories__/helpers.js
shlomitc: this was the design guideline

PR #1547, opened 2023-09-18
Feat/talko/label clickable

talkor:
1. Add a clickable variation to Label
2. Add a `onClick` prop to Label and apply clickable variation when it's provided
3. Nest styles in Label

https://monday.monday.com/boards/5113955565/pulses/5114222278

### DIFF
--- a/src/components/Label/Label.module.scss
+++ b/src/components/Label/Label.module.scss
@@ -10,6 +10,10 @@
   border-radius: var(--border-radius-small);
   padding: 1px var(--spacing-small);
   position: relative;
+
+  &.clickable {
+    cursor: pointer;
+  }
 }
 
 .withLeg {
@@ -29,77 +33,114 @@
   transform-origin: bottom left;
 }
 
-.kindFill.colorPrimary {
-  background-color: var(--primary-color);
-}
+.kindFill {
+  &.colorPrimary {
+    background-color: var(--primary-color);
 
-.kindFill.colorPrimary svg path {
-  fill: var(--primary-color);
-}
+    svg path {
+      fill: var(--primary-color);
+    }
 
-.kindFill.colorDark {
-  background-color: var(--inverted-color-background);
-}
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--primary-hover-color);
+    }
+  }
 
-.kindFill.colorDark svg path {
-  fill: var(--inverted-color-background);
-}
+  &.colorDark {
+    background-color: var(--inverted-color-background);
 
-.kindFill.colorNegative {
-  background-color: var(--negative-color);
-}
+    svg path {
+      fill: var(--inverted-color-background);
+    }
+  }
 
-.kindFill.colorNegative svg path {
-  fill: var(--negative-color);
-}
+  &.colorNegative {
+    background-color: var(--negative-color);
 
-.kindFill.colorPositive {
-  background-color: var(--positive-color);
-}
+    svg path {
+      fill: var(--negative-color);
+    }
 
-.kindFill.colorPositive svg path {
-  fill: var(--positive-color);
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--negative-color-hover);
+    }
+  }
+
+  &.colorPositive {
+    background-color: var(--positive-color);
+
+    svg path {
+      fill: var(--positive-color);
+    }
+
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--positive-color-hover);
+    }
+  }
 }
 
 .kindLine {
   border: 1px solid currentColor;
-}
 
-.kindLine .legWrapper {
-  left: -1px;
-  top: calc(100% + 1px);
-}
+  .legWrapper {
+    left: -1px;
+    top: calc(100% + 1px);
+  }
 
-.kindLine.colorPrimary {
-  color: var(--primary-color);
-}
+  &.colorPrimary {
+    color: var(--primary-color);
 
-.kindLine.colorPrimary svg path {
-  fill: var(--primary-color);
-}
+    svg path {
+      fill: var(--primary-color);
+    }
 
-.kindLine.colorDark {
-  color: var(--inverted-color-background);
-}
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--primary-background-hover-color);
+    }
+  }
 
-.kindLine.colorDark svg path {
-  fill: var(--inverted-color-background);
-}
+  &.colorDark {
+    color: var(--inverted-color-background);
 
-.kindLine.colorNegative {
-  color: var(--negative-color);
-}
+    svg path {
+      fill: var(--inverted-color-background);
+    }
 
-.kindLine.colorNegative svg path {
-  fill: var(--negative-color);
-}
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--primary-background-hover-color);
+    }
+  }
 
-.kindLine.colorPositive {
-  color: var(--positive-color);
-}
+  &.colorNegative {
+    color: var(--negative-color);
 
-.kindLine.colorPositive svg path {
-  fill: var(--positive-color);
+    svg path {
+      fill: var(--negative-color);
+    }
+
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--negative-color-selected);
+    }
+  }
+
+  &.colorPositive {
+    color: var(--positive-color);
+
+    svg path {
+      fill: var(--positive-color);
+    }
+
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--positive-color-selected);
+    }
+  }
 }
 
 @include keyframe(label-spin-in-emphasized) {
--- a/src/components/Label/Label.tsx
+++ b/src/components/Label/Label.tsx
@@ -2,13 +2,14 @@ import { camelCase } from "lodash-es";
 import { ComponentDefaultTestId, getTestId } from "../../tests/test-ids-utils";
 import cx from "classnames";
 import { getStyle } from "../../helpers/typesciptCssModulesHelper";
-import React, { FC, useMemo } from "react";
+import React, { FC, useMemo, useRef } from "react";
 import { backwardCompatibilityForProperties } from "../../helpers/backwardCompatibilityForProperties";
 import Text from "../Text/Text";
 import Leg from "./Leg";
 import { LabelColor, LabelKind } from "./LabelConstants";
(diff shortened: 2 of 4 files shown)


src/components/Label/__stories__/label.stories.mdx
```diff
@@ -138,5 +140,18 @@ In case of visual overload, use the secondary label in order to create hirarchy.
   </Story>
 </Canvas>
 
+### Clickable
+Default prop is set to false. In case you want to activate, change boolean to true.
```
talkor, line 144: This description might be misleading, we might want to change it because it's not a boolean prop

src/components/Label/__tests__/Label-tests.jest.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+describe("Label", () => {
+  const defaultTestId = getTestId(ComponentDefaultTestId.LABEL);
+
+  it("should not call onClick callback when label doens't have an onClick callback", () => {
+    const onClick = jest.fn();
+    render(<Label text="Label" />);
+    fireEvent.click(screen.getByTestId(defaultTestId));
+    expect(onClick).not.toBeCalled();
+  });
```
shlomitc, line 14: this test looks a bit redundant, you're not passing the `onClick` at all so it won't be ever called

shlomitc (changes requested):

src/components/Label/__stories__/label.stories.mdx
```diff
@@ -138,5 +140,18 @@ In case of visual overload, use the secondary label in order to create hirarchy.
...
 </Canvas>
 
+### Clickable
+Labels with an onClick prop change appearance on hover, focus, and click.
+<Canvas>
+  <Story name="Clickable">
+    <div className="monday-storybook-label_block">
+      <div className="monday-storybook-label_inline-block">
+        <Label text="New" kind={Label.kinds.LINE} onClick={NOOP} />
+        <Label text="Beta" onClick={NOOP} />
```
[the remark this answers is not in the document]
talkor: This is happening because the filled label is 22px in height and the lined one is 24px in height, they both should be 22px according to spec. I can fix the height and it will resolve this issue as well, or change the clickable element to be the inner element instead and resolve the height in a different PR. Which one sounds better?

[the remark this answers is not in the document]
talkor: I fixed the height difference, now the focus works well. Moving the onClick handler to the inner element might not be intuitive IMO, this can also make tests less intuitive. 

But, it will fail Chromatic, though it will now have the correct height

Screenshot 2023-09-18 at 16 38 38
Screenshot 2023-09-18 at 16 38 48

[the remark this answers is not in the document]
YossiSaadi: If this is the behavior Meirav/Yael wants then it's ok it will fail Chromatic, just accept the new baseline on your branch's build

YossiSaadi (approved):
Other than Shlomi's comments (that you already addressed, looks cool :))
Congrats on your first PR 🎉 !

src/components/Label/Label.tsx
```diff
@@ -24,6 +25,7 @@ interface LabelProps extends VibeComponentProps {
   text?: string;
   isAnimationDisabled?: boolean;
   isLegIncluded?: boolean;
+  onClick?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void;
```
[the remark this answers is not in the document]
talkor: Ohh yeah good catch, BTW it has a Text component inside which is a div, so we have a block element inside an inline element which is useless already. Does Text has an option to be an inline element? Couldn't find such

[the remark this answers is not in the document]
YossiSaadi: discussed with @SergeyRoyt about it, we should map all those cases, as it is an illegal html

You can pass `component` prop to `Text`

src/components/Label/Label.tsx
```diff
@@ -53,12 +57,30 @@ const Label: FC<LabelProps> & {
...
           [styles.withLeg]: isLegIncluded
         },
-        labelClassName
+        labelClassName,
+        { [styles.clickable]: isClickable }
       ),
-    [kind, color, isAnimationDisabled, isLegIncluded, labelClassName]
+    [kind, color, isAnimationDisabled, isLegIncluded, labelClassName, isClickable]
   );
+  const labelRef = useRef<HTMLSpanElement>(null);
```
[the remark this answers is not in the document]
YossiSaadi: @talkor 😎 awesome

src/components/Label/__stories__/label.stories.mdx
```diff
@@ -138,5 +140,18 @@ In case of visual overload, use the secondary label in order to create hirarchy.
   </Story>
 </Canvas>
 
+### Clickable
```
SergeyRoyt, line 143: Have you talked with the designers about the section this example should be? 
It's arguable, but I think it fits `Variants` better than `Use cases and examples` - like we have in Chips. I'd make sure with one of designers

src/components/Label/Label.module.scss
```diff
@@ -10,6 +10,10 @@
   border-radius: var(--border-radius-small);
   padding: 1px var(--spacing-small);
   position: relative;
+
+  &.clickable {
+    cursor: pointer;
```
SergeyRoyt, line 15: Also our `focus-style` is needed here - all together can be achieved by importing Clickable mixin as done in `Chips`
```
@use "../Clickable/Clickable.module" as clickable;
@include clickable.clickable;
```

src/components/Label/Label.tsx
```diff
@@ -53,12 +57,30 @@ const Label: FC<LabelProps> & {
...
+  const clickableProps = useClickableProps(
+    {
+      onClick,
```
SergeyRoyt, line 68: Let's consider adding `preventDefault` to the `useClickableProps` `onClick` callback - it would remove the jump when the element is pressed by Space button. As far as I know, it's the best practice

src/components/Label/Label.module.scss
talkor: Oh that's important 🙏🏼

src/components/Label/Label.module.scss
```diff
@@ -29,77 +33,115 @@
...
+    &.clickable:active {
+      background-color: var(--primary-hover-color);
+    }
+  }
 
-.kindFill.colorDark svg path {
-  fill: var(--inverted-color-background);
-}
+  &.colorDark {
+    background-color: var(--inverted-color-background);
```
SergeyRoyt, line 51: That's not related to your changes, but I spotted that we have a color token issue here in dark themes 
image
In our storybook we have a theming switcher, which is always nice to check when adding usage of new tokens. Here is how it can be found
image

I propose we tackle it in a separate task. Could you please add a corresponding task to the board and solve it in a separate PR? Once it's merged to master, let's just update from master here and make sure it's working fine with clickable ability. Or we can do the other way around - whatever you prefer

src/components/Label/Label.module.scss
```diff
@@ -29,77 +33,115 @@
...
+    svg path {
+      fill: var(--primary-color);
+    }
 
-.kindFill.colorDark {
-  background-color: var(--inverted-color-background);
-}
+    &.clickable:hover,
+    &.clickable:active {
+      background-color: var(--primary-hover-color);
```
SergeyRoyt, line 46: image
Hovered outlined version doesn't look too good in dark theme. Though token seems correct... I'd consult with the designers just to be sure it's okay - maybe we need to change text color as well e.g.

src/components/Label/Label.module.scss
SergeyRoyt, line 51:
```
&.clickable:hover,
&.clickable:active
```
are missing here on purpose?

talkor: Yes, they have the same colors whatsoever

SergeyRoyt (changes requested):
Let's fix the issues 🙏

src/components/Label/__stories__/label.stories.mdx
talkor: Moved to Variant 👍🏼

src/components/Label/Label.module.scss
talkor: Fixed in this PR, changed it to the correct token 🙏🏼

src/components/Label/Label.module.scss
talkor: Talked to Yael, nothing to do at the moment, this is by design although contrast is not that great, I will raise an issue in the team meeting about color contrast

src/components/Label/__tests__/__snapshots__/label-snapshot-tests.jest.tsx.snap
```diff
@@ -5,8 +5,8 @@ exports[`Label renders correctly fill dark 1`] = `
   className=""
   data-testid="label"
 >
-  <div
-    className="typography onPrimary start singleLineEllipsis text text2Normal label kindFill colorDark withAnimation"
+  <span
+    className="typography onInverted start singleLineEllipsis text text2Normal label kindFill colorDark withAnimation"
```
talkor, line 9: Fixed to the correct color token

src/components/Label/__tests__/__snapshots__/label-snapshot-tests.jest.tsx.snap
```diff
@@ -5,8 +5,8 @@ exports[`Label renders correctly fill dark 1`] = `
   className=""
   data-testid="label"
 >
-  <div
-    className="typography onPrimary start singleLineEllipsis text text2Normal label kindFill colorDark withAnimation"
+  <span
```
talkor, line 8: Internal element is now an inline element and not a block

src/components/Label/Label.tsx
```diff
@@ -53,13 +56,42 @@ const Label: FC<LabelProps> & {
...
+
+  const onClickCallback = useCallback(
+    (event: React.MouseEvent<HTMLSpanElement, MouseEvent>) => {
+      if (onClick) {
+        event.preventDefault();
+        onClick(event);
+      }
+    },
+    [onClick]
+  );
```
SergeyRoyt, line 73: Why not do this inside `useClickableProps`?

talkor: What do you mean?

talkor: Oh you mean always preventDefault inside `useClickableProps`?

SergeyRoyt:
Now the color is not changing when hovering clickable labels, cause of clickable selector change

SergeyRoyt (approved):
Great job! Congrats with first Vibe PR! 🥳

PR #1549, opened 2023-09-19
fix: <Chips> text overflow Tooltip issue - replace internal Tooltip logic with Text overflow Tooltip logic

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5178445486

But loosing Tooltip on focus


src/components/Chips/__stories__/chips.stories.mdx
```diff
@@ -205,6 +205,40 @@ When a chip appears on a background color identical to its color, use `showBorde
   </Story>
 </Canvas>
 
+### Overflowing chip
```
shlomitc, line 208: As this is not proper documentation, I really think we should consider having a section in the storybook that is dedicated to tests incl. interaction tests (i.e. Tests/Chips).
Another option would be to create another dedicated storybook just for tests, but I'd go with the quickest one for now

shlomitc (approved):

PR #1552, opened 2023-09-20
feat(TableRow): add highlighted styling for TableRow

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5186000360

Without `onClick` functionality - will be added in a separate PR https://github.com/mondaycom/monday-ui-react-core/pull/1755


src/components/Table/Table/__stories__/helpers.js
```diff
@@ -259,5 +259,6 @@ export const TableAvatar = ({ text }) => (
     size={Avatar.sizes.SMALL}
     ariaLabel={text}
     backgroundColor="dark_purple"
+    tabIndex={-1}
```
shlomitc, line 262: This smells bad.
The fact that we're not using the Avatar as is and we need to trick it with tabIndex means every consumer will need to do it.

Also, the fact that it's a helper component means it won't appear in the code example as copy-pastable

src/components/Table/TableRow/TableRow.tsx
```diff
@@ -1,16 +1,46 @@
...
+import React, { FC, useCallback, useRef } from "react";
 import styles from "./TableRow.module.scss";
+import { VibeComponentProps } from "../../../types";
 import { ITableCellProps } from "../TableCell/TableCell";
+import { useKeyEvent } from "../../../hooks";
+import { SELECTION_KEYS } from "../../../constants";
+import { NOOP } from "../../../utils/function-utils";
 
 export interface ITableRowProps extends VibeComponentProps {
+  highlight?: boolean;
```
shlomitc, line 10: 1. add documentation to the prop
2. add a story example

src/components/Table/TableRow/TableRow.tsx
```diff
@@ -1,16 +1,46 @@
...
+  onClick?: (event: React.MouseEvent | React.KeyboardEvent) => void;
   children?: React.ReactElement<ITableCellProps> | React.ReactElement<ITableCellProps>[];
   style?: React.CSSProperties;
 }
 
-const TableRow: FC<ITableRowProps> = ({ children, style = {} }) => {
+const TableRow: FC<ITableRowProps> = ({ highlight = false, onClick = NOOP, children, style = {} }) => {
```
shlomitc, line 16: This should be fine as well without the default values
```suggestion
const TableRow: FC<ITableRowProps> = ({ highlight, onClick, children, style }) => {
```

src/components/Table/TableRow/TableRow.tsx
```diff
@@ -1,16 +1,46 @@
...
+  const ref = useRef(null);
+  useKeyEvent({
```
shlomitc, line 18: this is a bit tricky since our table rows aren't clickable. I didn't see it in the storybook nor in tests so I can't tell how it works

shlomitc (changes requested):

src/components/Table/Table/__stories__/helpers.js
YossiSaadi: I agree, it's currently for the purpose of future-story in storybook, to not focus on it.
currently, we don't have story for the highlight, Yael wants to come up with one and then I'll add it.

As we apply tabIndex 0 to avatar by default, it causes this headache and it would cause others the headache as well (and I guess even today it is, I don't see a reason to focus on an Avatar if it has no onClick event).
But removing this tabIndex from avatar would cause breaking change, so I'm pretty much blocked here :/

src/components/Table/TableRow/TableRow.tsx
YossiSaadi: It's our implementation for handling keyboard press on elements, it works the same as doing:
```
<div
  ...
  onKeyDown={someCallback}
```

without the overhead for checking specific keycodes

src/components/Table/TableRow/TableRow.tsx
YossiSaadi: having no default for highlight would force me do two things:
```aria-selected={highlight || false}```
for onclick it makes sense, I'll add a check before calling the method itself

Comment (shlomitc) [reply src/components/Table/TableRow/TableRow.tsx:16]: cool, so whatever you prefer

File: src/components/Table/TableRow/TableRow.tsx (hunk 4 above)
Comment (shlomitc) [reply src/components/Table/TableRow/TableRow.tsx:18]: but do we allow focusing on a row? because you probably need some "focus" indication before clicking

Comment (YossiSaadi) [reply src/components/Table/TableRow/TableRow.tsx:18]: yep, it's happening on
```
tabIndex={onClick ? 0 : -1}
```

File: src/components/Table/Table/__stories__/helpers.js (hunk 1 above)
Comment (SergeyRoyt) [reply src/components/Table/Table/__stories__/helpers.js:262]: We recently removed tabIndex = 0 from Avatar. I didn't feel like it's super breaking change actually, so we did it as a part of Hacktoberfest - https://github.com/mondaycom/monday-ui-react-core/pull/1594
Instead we introduced logic which makes tabIndex = 0, if there is `onClick` or tooltip content is not empty
image

It doesn't solve this issue though! Cause tooltip content isn't empty, so the Avatar is focusable by default

File: src/components/Table/TableRow/TableRow.tsx (hunk 2 above)
Comment (SergeyRoyt) [reply src/components/Table/TableRow/TableRow.tsx:14]: Done ✅
Don't understand why prop name isn't `selected` or smth... Highlighted is a bit not-straightforward to understand. Let's think twice about the naming before we release this

Comment (SergeyRoyt) [reply src/components/Table/TableRow/TableRow.tsx:14]: Checkout the story btw, I've done some experimental stuff. Will also consult with Meirav, if you'll like it
image

File: src/components/Table/TableRow/TableRow.tsx (hunk 4 above)
Comment (SergeyRoyt) [reply src/components/Table/TableRow/TableRow.tsx]: Looks like this
image
A bit cut from the sides

File: src/components/Table/TableRow/TableRow.tsx (hunk 2 above)
Comment (shlomitc) [reply src/components/Table/TableRow/TableRow.tsx:14]: There will be another feature for "selection" which is adding checkboxes per row, so we didn't want to mix both.
We find `highlight` accurate as it highlights the row but

File: src/components/Table/TableRow/TableRow.module.scss [hunk 5]
```diff
@@ -1,9 +1,21 @@
...
+
+  &:hover {
+    > * {
+      background-color: var(--allgrey-background-color);
+    }
+
+    &[aria-selected="true"] > * {
+      background-color: var(--primary-selected-hover-color);
+    }
   }
```
shlomitc, line 20: I'm not a big fan of `*`, I'd consider targeting specific classes when possible

src/components/Table/TableRow/TableRow.tsx
shlomitc: let's remove this tip, it's not relevant

src/components/Table/Table/__stories__/helpers.js
SergeyRoyt: Done ✅

src/components/Table/TableRow/TableRow.tsx
SergeyRoyt: So in the future it will be another `selected` prop or what? If row will be selected via checkbox it will be highlighted or not always? Could rows be highlighted, but not selected? 
We are using highlight in `aria-selected` attribute and it's also a bit confusing
What should I write to the description comments?

Btw, maybe the name should be `highlighted` then and not `highlight`?

src/components/Table/Table/__stories__/table.stories.mdx
```diff
@@ -349,6 +348,48 @@ This is an example of a table with 5000 rows
   </Story>
 </Canvas>
 
+## [Clickable-highlighted rows - demo story]
```
SergeyRoyt, line 351: **TODO remove demo story**

shlomitc: the story should be replaced with a simple table with a few lines highlighting its first row.
no clicking

src/components/Table/TableRow/TableRow.tsx
shlomitc: according to the latest discussions, we should not introduce a click functionality on the row.
the only behavior added is to allow a row to become highlighted. the way to trigger it would be by composing some clickable item in a cell.

shlomitc (changes requested):

shlomitc (approved):

PR #1558, opened 2023-09-27
sergeyro/docs/useClickOutside event name arg

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5231359395


YossiSaadi (approved):
💪🏼 LGTM

funny we exported from a different name than it really is

PR #1559, opened 2023-09-27
docs: sidebar tags

talkor:
https://monday.monday.com/boards/3532714909/pulses/4794600988

Screenshot 2023-09-27 at 14 18 30

There are basically two possible ways of passing a status to a component or a story:
1. By adding the status in square brackets to the component/story name (e.g. `Text [new]`, `LegacyHeading [deprecated]` etc).
2. By passing the status as a param - works only with stories (not working with components). 

Regarding `#2`, in the `renderLabel` function we get additional params only for stories and not for components, so it won't work in the component level. This is why I used `#1` as well, as a workaround. Though it's not as manageable as we'd want (since it's providing a string in the component's name), I think it's good enough for now as [Storybook are planning on releasing light tagging in the near future ](https://discord.com/channels/486522875931656193/1102958649455955999/1103335874139279364) (planned for v7.1 but not released yet unfortunately), and it can be used on both component and story level, so only a single API.

The `StatusTag` and `SidebarItem` components will be obviously re-written with proper css, tokens and design in `vibe-storybook-components`, this is just for demonstration and shouldn't be merged. 

Would love to hear your thoughts :)


src/components/LegacyHeading/__stories__/LegacyHeading.stories.mdx
```diff
@@ -22,7 +22,7 @@ export const metaSettings = createStoryMetaSettingsDecorator({
 });
 
 <Meta
```
talkor, line 24: Passing `parameters: { status: 'deprecated' }` to the `Meta` component won't work in the Component level

.storybook/addons.jsx
```diff
@@ -22,5 +24,22 @@ const theme = create({
 });
 
 addons.setConfig({
-  theme
+  theme,
+  sidebar: {
+    renderLabel: ({ name, parameters = {} }) => {
+      const statusRegex = /\[([^)]+)\]/gi;
```
talkor, line 30: Using the square brackets is obviously just a suggestion, we can choose anything we want

src/components/Text/__stories__/Text.stories.mdx
```diff
@@ -138,7 +138,7 @@ When the text is longer than its container and the ellipsis flag is on, the end
 We support two kinds of ellipsis: single-line ellipsis with a tooltip displayed in hover or ellipsis after multiple lines. You can see examples for both use cases below.
 
 <Canvas>
-  <Story name="Overflow" play={textOverflowSuite}>
+  <Story name="Overflow" play={textOverflowSuite} parameters={{ status: "new" }}>
```
talkor, line 141: This is the #1 of passing the status, but doing `name="Overflow [new]"` would do the same

SergeyRoyt:
Like it!
As we discussed, I think for now we should use brackets for component level (it won't change too often) and parameters for story level

shlomitc (commented):
image

talkor:
Applied only to LegacyHeading for now:

Screenshot 2023-10-10 at 13 38 51

src/components/Text/__stories__/Text.stories.mdx
```diff
@@ -204,7 +204,11 @@ A Text component with a link skin can be used to create a link within running te
   <Story name="Links inside running text">
     <Flex direction={Flex.directions.COLUMN} align={Flex.align.START} gap={Flex.gaps.SMALL}>
       <Text>
-        This is the story of a <Link href="/?path=/docs/foundations-typography--page">link</Link> inside running text.
+        This is the story of a
```
talkor, line 207: These were broke in a different PR in Chromatic

SergeyRoyt (approved):
Great stuff!

PR #1561, opened 2023-09-27
fix(Label): apply preventDefault when calling onClick event

talkor:
https://monday.monday.com/boards/3532714909/pulses/5232208575


YossiSaadi (approved):

SergeyRoyt (approved):
Other option can be to add `preventDefault` param to `useClickableProps`, which then will be propagated to `useKeyboardButtonPressedFunc`
But this solution is also fine - whatever you prefer

Btw, I've added a task to backlog to add interaction tests for EditableHeading to test spacing issue if it ever occurs again

PR #1564, opened 2023-10-01
docs: component-name-decorator- text color = fixed-dark

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5253754106


src/storybook/components/component-name/component-name-decorator.scss
```diff
@@ -1,4 +1,5 @@
 .vibe-storybook-component-name {
+  color: var(--sb-text-color-fixed-dark);
```
YossiSaadi, line 2: is this a new token?

YossiSaadi (approved):

src/storybook/components/component-name/component-name-decorator.scss
SergeyRoyt: Yes

PR #1581, opened 2023-10-02
fix: avatar remove default tabIndex = 0

pipliya:


SergeyRoyt:
Hey, @pipliya!
Not exactly what we need - we do want to persist tabIndex = 0 for `ListItem` and `Menu` - we want to remove `Avatar` default `tabIndex = 0`, instead it should be equal to 0 only if `!!onClick` prop or if it has any kind of a tooltip (please check this page https://style.monday.com/?path=/docs/media-avatar-avatar--avatar-with-tooltip)

pipliya:
Thanks @SergeyRoyt for your clarification. Working on it!

pipliya:
Closing this for now as it seems a little complex for me at a moment. Thanks!

PR #1596, opened 2023-10-03
chore: enforce using npm

talkor:
https://monday.monday.com/boards/3532714909/pulses/5265659631

To prevent issues in dependency resolution when using different package managers

When using `yarn install` for example it will throw an error:
Screenshot 2023-10-03 at 16 07 28

### DIFF
--- a/package.json
+++ b/package.json
@@ -262,5 +262,8 @@
       "PR: Breaking Changes 💥": "Breaking Changes",
       "PR: Icon 💎": "New Icons"
     }
+  },
+  "engines": {
+    "yarn": "please-use-npm"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -262,5 +262,8 @@
       "PR: Breaking Changes 💥": "Breaking Changes",
       "PR: Icon 💎": "New Icons"
     }
+  },
+  "engines": {
```
talkor, line 266: We can also specify npm and node version if we want

YossiSaadi: coool

sahariko: Node version should be handled by `.nvmrc`.

sahariko (approved):
Super cool stuff!

YossiSaadi (approved):

PR #1603, opened 2023-10-03
infra: allow running chromatic workflow on pull_request

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4526273740

Allow running `chromatic` workflow on `pull_request` - for public forks after approval from the maintainers

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -3,7 +3,6 @@
 name: 'Chromatic'
 
 on:
-  push:
   pull_request:
     types: [opened, synchronize]
 
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,26 +1,22 @@
 # .github/workflows/chromatic.yml
 
-# Workflow name
 name: 'Chromatic'
 
-# Event for the workflow
-on: push
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize]
 
-# List of jobs
 jobs:
   chromatic-deployment:
-    # Operating System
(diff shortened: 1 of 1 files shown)


.github/workflows/chromatic.yml
```diff
@@ -1,26 +1,22 @@
...
 
-# Workflow name
 name: 'Chromatic'
 
-# Event for the workflow
-on: push
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize]
```
YossiSaadi, line 8: So maybe only on push to master?
Since it will run twice for both opened and pushed when a PR is created & twice for both synchronize and push when a commit is pushed

YossiSaadi (commented):
We can also instead have a pull request labeled event with specific label. 
A lot of open source does that. And only maintainers are able to add labels.

.github/workflows/chromatic.yml
SergeyRoyt: `push` is not running for forks that's reason this workflow currently doesn't run on the PRs of external contributors

About the labels - good trick, but allowing it on `pull_request` seems secure and same time easier, no? You think it might be a security breach?

YossiSaadi: I don't think it's a security breach, it would just run twice like many of our actions that we should anyway fix

YossiSaadi (approved):
I know it's for Hacktoberfest (btw, amazing work in the past two days from your side!), so approving for now, but it would most definitely run twice for each push so we need to mind these and fix all of our flows that run twice for every push (e.g. tests)

Issue #1607, opened 2023-10-04
feat: <AttentionBox> change default icon for Primary type to Info icon


SergeyRoyt:
[Monday ticket](https://monday.monday.com/boards/3532714909/pulses/5075672010)

Currently `AttentionBox` have a default icon = `AlertIcon`
<img width="362" alt="image" src="https://github.com/mondaycom/monday-ui-react-core/assets/104433616/8e38cc64-295f-41a1-979e-e8d05e846dc7">

We want to replace it with `Info` icon, but only when `type` is `primary`:
<img width="389" alt="image" src="https://github.com/mondaycom/monday-ui-react-core/assets/104433616/a521aa4c-75f8-46b5-bdfb-dfb0be31d62f">

Kritik-J:
@SergeyRoyt please assign this to me

SergeyRoyt:
You got it, champ! ;)

SergeyRoyt:
Issue solved!
Thanks a lot, @Kritik-J!

PR #1611, opened 2023-10-04
feat: <AttentionBox> - update default icon for Primary type to Info icon

Kritik-J:
Resolves: #1607

### DIFF
--- a/src/components/AttentionBox/AttentionBox.tsx
+++ b/src/components/AttentionBox/AttentionBox.tsx
@@ -7,6 +7,7 @@ import Icon from "../Icon/Icon";
 import IconButton from "../IconButton/IconButton";
 import CloseSmall from "../Icon/Icons/components/CloseSmall";
 import AlertIcon from "../Icon/Icons/components/Alert";
+import InfoIcon from "../Icon/Icons/components/Info";
 import { IconType } from "../Icon/IconConstants";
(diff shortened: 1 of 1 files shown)


src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -72,6 +73,14 @@ const AttentionBox: React.FC<AttentionBoxProps> & {
 
   const overrideClassName = backwardCompatibilityForProperties([className, componentClassName]);
 
+  const overrideIcon = useMemo(() => {
+    if (type === AttentionBox.types.PRIMARY) {
+      return InfoIcon;
+    }
+
+    return icon;
+  }, [type, icon]);
```
SergeyRoyt, line 82:
```suggestion
  const defaultIcon = useMemo(() => {
    return type === AttentionBox.types.PRIMARY ? InfoIcon : AlertIcon;
  }, [type]);
  const overrideIcon = icon ?? defaultIcon;
```

Cause currently you are overriding any icon if `type === AttentionBox.types.PRIMARY` and you need to only provide a default value for this case

SergeyRoyt (changes requested):
Little fix is needed

SergeyRoyt (approved):
Great! Thanks for the contribution! 🥳

Issue #1577, opened 2023-10-02
docs: <Checkbox> Usage guidelines - fix font


SergeyRoyt:
Fix font in the first usage - should be the same as everywhere else
![image](https://github.com/mondaycom/monday-ui-react-core/assets/104433616/566152bc-ca05-4778-a3c6-7aa6d28fbde6)

Probably include `vibe-text` is missing in the custom class

AlisanCavus:
I would like to work on it.

SergeyRoyt:
Thanks, @AlisanCavus ! It's yours ;)

Krishnakalani111:
If it's still addressable i would like to work on it

SergeyRoyt:
Thanks for the contribution, @SoumyadiptoPal!
Will assign the issue to you, once you write smth here

PR #1621, opened 2023-10-06
docs: <Checkbox> Usage guidelines - fix font #1577

SoumyadiptoPal:
Fixes issue #1577 
#### Basic

- [x] PR has description.
- [x] Component is written in TypeScript.

#### Style

- [x] Styles are added to `NewComponent.modules.scss` file inside the `NewComponent` folder.
- [x] Styles are defined using [monday-ui-style](https://github.com/mondaycom/monday-ui-style) tokens
- [x] Displays correctly in all the themes

#### Storybook

- [x] Stories were added to `/src/components/NewComponent/__stories__/NewComponent.stories.mdx` file.
- [x] Stories implemented using [vibe-storybook-components](https://github.com/mondaycom/vibe-storybook-components) components and tokens.

#### Tests

#### Accessibility

- [x] Component is accessible.
- [x] Component is compatible with [Vibe Accessibility Guidelines](https://style.monday.com/?path=/docs/foundations-accessibility--page)
image

### DIFF
--- a/src/components/Checkbox/__stories__/checkbox.stories.scss
+++ b/src/components/Checkbox/__stories__/checkbox.stories.scss
@@ -1,4 +1,5 @@
 @import "~vibe-storybook-components/dist/mixins";
+@import "~monday-ui-style/dist/mixins";
 
 .monday-storybook-checkbox {
   &_border {
@@ -37,12 +38,13 @@
   &_list {
     display: flex;
     flex-direction: column;
+    font-family :var(--font-family) !important;
   }
 
   &_list-item {
     display: inline-block;
     width: 100%;
-    @include sb-content-text();
+    font-family :var(--font-family) !important;
   }
 
   &_link {
--- a/src/components/Checkbox/__stories__/checkbox.stories.scss
+++ b/src/components/Checkbox/__stories__/checkbox.stories.scss
@@ -1,5 +1,4 @@
 @import "~vibe-storybook-components/dist/mixins";
-@import "~monday-ui-style/dist/mixins";
 
 .monday-storybook-checkbox {
   &_border {
@@ -38,13 +37,17 @@
   &_list {
     display: flex;
     flex-direction: column;
-    font-family :var(--font-family) !important;
+    span {
+      @include sb-content-text-without-color;
+    }
   }
 
   &_list-item {
     display: inline-block;
     width: 100%;
-    font-family :var(--font-family) !important;
+    span {
+      @include sb-content-text-without-color;
+    }
   }
 
   &_link {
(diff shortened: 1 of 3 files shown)


SoumyadiptoPal:
@SergeyRoyt I have fixed issue #1577

src/components/Checkbox/__stories__/checkbox.stories.scss
```diff
@@ -1,4 +1,5 @@
 @import "~vibe-storybook-components/dist/mixins";
+@import "~monday-ui-style/dist/mixins";
```
SergeyRoyt, line 2: It's not needed here - no mixins from monday-ui-style are used

src/components/Checkbox/__stories__/checkbox.stories.scss
```diff
@@ -37,12 +38,13 @@
   &_list {
     display: flex;
     flex-direction: column;
+    font-family :var(--font-family) !important;
```
SergeyRoyt, line 41: `font-family` is not the only thing to define the font. Let's instead use `sb-content-text` mixin and increase the specificity of selector, so all the styles would apply. Btw, even thought, it's only documentation using `!important` in general is a bad practice

I propose we'd do smth like this:
```suggestion
    span {
      @include sb-content-text;
    }
```


And also we'll need to wrap `Use checkboxes to:` text with `span` tag
This way all out mixin styles will be applied both to the header title and to the list items without using `!important`

src/components/Checkbox/__stories__/checkbox.stories.scss
```diff
@@ -37,12 +38,13 @@
...
   }
 
   &_list-item {
     display: inline-block;
     width: 100%;
-    @include sb-content-text();
+    font-family :var(--font-family) !important;
```
SergeyRoyt, line 47: Then this won't be needed too

SergeyRoyt (changes requested):
Hey, @SoumyadiptoPal!
Thanks for the efforts! Left a few notes needed to finish the job! 💪

SergeyRoyt:
Btw, it's always better to name the branches with some semantic names, e.g. in this case it could have been smth like
`docs/checkbox-usage-guidelines-font`
Doesn't really matter in this case though, no worries, just a tip for the future :)

SoumyadiptoPal:
> Btw, it's always better to name the branches with some semantic names, e.g. in this case it could have been smth like `docs/checkbox-usage-guidelines-font` Doesn't really matter in this case though, no worries, just a tip for the future :)

Thank out for your advice. Will try to follow this from next time.👍👍

SoumyadiptoPal:
@SergeyRoyt I have updated the code.

SergeyRoyt (approved):
Looks great, thanks!!
Please, also write a comment under the issue itself, so I can assign it to you and you'll get your Hactoberfest reward :)

PR #1623, opened 2023-10-08
fix: <Toast> use color - text-color-fixed-light

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4776199392


SergeyRoyt:
image

PR #1624, opened 2023-10-08
fix: <Counter> use color - text-color-fixed-light

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/4776199392


SergeyRoyt:
Tested in monolith - looks good in all themes and products

talkor (approved):

PR #1628, opened 2023-10-09
docs: Replace tip alpha warning deprecated warning with vibe storybook components

talkor:
https://monday.monday.com/boards/3532714909/pulses/5296950950


webpack/published-storybook-components.js
```diff
@@ -1,10 +0,0 @@
-const publishedStorybookComponents = {
-  // Components
-  "storybook/AlphaWarning": "storybook/components/alpha-warning/alpha-warning.tsx",
-  "storybook/Tip": "storybook/components/tip/tip.jsx"
```
talkor, line 4: Apparently these are the only two components that were exported as `storybookComponents`, the rest is commented-out and I guess they will be moved to `vibe-storybook-components` as well which makes the build scripts redundant. I'm not sure who's consuming them, I looked up in GH search and couldn't find but I'm not 100% sure

package.json
```diff
@@ -17,7 +17,6 @@
     },
     "./interactionsTests": "./dist/esm/src/tests/interactionsTests.js",
     "./testIds": "./dist/esm/src/tests/testIds.js",
-    "./storybookComponents": "./dist/storybook/index.js",
```
SergeyRoyt, line 20: We probably shouldn't remove this endpoint - it's a breaking change
I think we should consider doing it in the next major version

scripts/build-utils.js
```diff
@@ -50,18 +46,7 @@ function buildComponentsTypesIndexFile() {
   convertExportsToFile(exportsWithTypescript.concat(exportsWithJavascript), "types.d.ts");
 }
 
-function buildStorybookComponentsIndexFile() {
-  const imports = ["import 'vibe-storybook-components/index.css';"];
```
SergeyRoyt, line 54: This line should remain I guess

scripts/build-utils.js
```diff
@@ -50,18 +46,7 @@ function buildComponentsTypesIndexFile() {
...
-  const exports = Object.entries(publishedStorybookComponents).map(([name, _path]) => {
-    const fileName = name.split("/").slice(-1);
-    return buildComponentExport(fileName, `./${fileName}`);
-  });
-  exports.push("export * from 'vibe-storybook-components'");
```
SergeyRoyt, line 59: As well as this reexport from vibe-storybook-components

SergeyRoyt (changes requested):
Let's not do the breaking change for now and leave reexport of `vibe-storybook-components` on `/storybookComponents` endpoint. The rest can be removed, but, please, make sure that /storybookComponents is still building and working correctly

talkor:
> Let's not do the breaking change for now and leave reexport of `vibe-storybook-components` on `/storybookComponents` endpoint. The rest can be removed, but, please, make sure that /storybookComponents is still building and working correctly

@SergeyRoyt 
Correct me if I'm mistaken but looks like those all scripts are for publishing just the Tip and AlphaWarning components. Including the export and import statements in the script. If we do not want to deploy breaking changes it means that those existing components should remain as they are, not sure it's being used though

SergeyRoyt:
> > Let's not do the breaking change for now and leave reexport of `vibe-storybook-components` on `/storybookComponents` endpoint. The rest can be removed, but, please, make sure that /storybookComponents is still building and working correctly
> 
> @SergeyRoyt Correct me if I'm mistaken but looks like those all scripts are for publishing just the Tip and AlphaWarning components. Including the export and import statements in the script. If we do not want to deploy breaking changes it means that those existing components should remain as they are, not sure it's being used though

One of the scripts is also adding reexport of vibe-storybook-components - this should be remain as it is, everything else can be removed
And regarding Tip and AlphaWarning and its not breaking cause they will be replaced with reexport from vibe-storybook-components

SergeyRoyt (approved):
Looks great now!

PR #1637, opened 2023-10-10
feat: add Toast warning variation

talkor:
https://monday.monday.com/boards/3532714909/pulses/4333991946
Screenshot 2023-10-10 at 11 25 44


SergeyRoyt (approved):
Looks great!

SergeyRoyt (commented):
Now saw that it's not with the latest version, let's merge from master to see how it'd look

If the override will work just fine, I'm ok with leaving this in CSS and fixed_light in JS - but let me know what you think

Issue #1648, opened 2023-10-11
chore: resolve lint warnings - < Menu />


SergeyRoyt:
[Monday ticket](https://monday.monday.com/boards/3532714909/pulses/5311541029)

<img width="923" alt="image" src="https://github.com/mondaycom/monday-ui-react-core/assets/104433616/ffd37a2f-ce28-4ba9-ac83-e97ab9b68f47">

Kritik-J:
@SergeyRoyt Please assign this to me

SergeyRoyt:
Closed thanks to @Kritik-J !

PR #1653, opened 2023-10-11
chore: resolve lint warnings - < Menu />

Kritik-J:
fixes: #1648

### DIFF
--- a/src/components/Menu/Menu/Menu.tsx
+++ b/src/components/Menu/Menu/Menu.tsx
@@ -34,7 +34,7 @@ import styles from "./Menu.module.scss";
 export interface MenuProps extends VibeComponentProps {
   /** Backward compatibility for props naming **/
   classname?: string;
-  size?: typeof SIZES[keyof typeof SIZES];
+  size?: (typeof SIZES)[keyof typeof SIZES];
   tabIndex?: number;
   ariaLabel?: string;
   ariaDescribedBy?: string;
@@ -87,15 +87,20 @@ const Menu: VibeComponent<MenuProps> & {
 
     const children = useMemo(() => {
       const allChildren = React.Children.toArray(originalChildren);
-      return allChildren.filter(child => {
-        // @ts-ignore
-        if (child.type.isMenuChild) return true;
-        console.error(
-          "Menu child must be a menuChild item (such as MenuItem, MenuDivider, MenuTitle, etc). This child is not supported: ",
-          child
-        );
-        return false;
-      });
+      return allChildren.filter(
+        (
+          child: ReactElement & {
+            type: Record<string, unknown>;
+          }
+        ) => {
+          if (child.type.isMenuChild) return true;
+          console.error(
+            "Menu child must be a menuChild item (such as MenuItem, MenuDivider, MenuTitle, etc). This child is not supported: ",
+            child
+          );
+          return false;
+        }
+      );
     }, [originalChildren]);
 
     const updateActiveItemIndex = useCallback(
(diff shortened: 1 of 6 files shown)


src/components/Menu/MenuItemButton/MenuItemButton.tsx
```diff
@@ -16,7 +16,7 @@ import styles from "./MenuItemButton.module.scss";
 interface MenuItemButtonProps extends VibeComponentProps {
   /** Backward compatibility for props naming **/
   classname?: string;
-  kind?: typeof MenuItemButton.kinds[keyof typeof MenuItemButton.kinds];
+  kind?: (typeof MenuItemButton.kinds)[keyof typeof MenuItemButton.kinds];
```
SergeyRoyt, line 19: Lint issue

src/components/Menu/Menu/Menu.tsx
```diff
@@ -34,7 +34,7 @@ import styles from "./Menu.module.scss";
 export interface MenuProps extends VibeComponentProps {
   /** Backward compatibility for props naming **/
   classname?: string;
-  size?: typeof SIZES[keyof typeof SIZES];
+  size?: (typeof SIZES)[keyof typeof SIZES];
```
SergeyRoyt, line 37: Lint issue

src/components/Menu/Menu/Menu.tsx
```diff
@@ -87,15 +87,20 @@ const Menu: VibeComponent<MenuProps> & {
 
     const children = useMemo(() => {
       const allChildren = React.Children.toArray(originalChildren);
```
SergeyRoyt, line 89: I propose we cast the type here for all the array, like this
```suggestion
      const allChildren = React.Children.toArray(originalChildren) as MenuChild[];
```
And let's define MenuChild in `MenuConstants` and reuse it, also let's make the type field optional to avoid the errors, should be smth like this:
```
type MenuChild = ReactElement & {
  type?: Record<string, unknown>;
};
```

SergeyRoyt (changes requested):
Wrote a few comments. Let's make sure that all the checks are green - test and lint

src/components/Menu/Menu/utils/utils.ts
```diff
@@ -1,7 +1,10 @@
 import { ReactElement } from "react";
 
-export function isMenuChildSelectable(child: ReactElement): boolean {
-  // @ts-ignore
+export function isMenuChildSelectable(
+  child: ReactElement & {
+    type: Record<string, unknown>;
+  }
```
SergeyRoyt, line 6:
```suggestion
  child: MenuChild;
```
+ lint fixes

src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -117,9 +117,12 @@ const MenuItem: VibeComponent<MenuItemProps> & {
     const isSubMenuOpen = !!children && isActive && hasOpenSubMenu;
     const hasChildren = !!children;
     const shouldShowSubMenu = hasChildren && isParentMenuVisible && isSubMenuOpen;
-    const submenuChild = children && React.Children.only(children);
+    const submenuChild =
+      children &&
+      (React.Children.only(children) as ReactElement & {
+        type: Record<string, unknown>;
+      });
```
SergeyRoyt, line 124: Let's revert this change and instead change type of `children` prop:
```diff
-   children?: ReactElement | ReactElement[];
+   children?: MenuChild | MenuChild[];
```

SergeyRoyt:
```suggestion
    const submenuChild: MenuChild = children && React.Children.only(children);
```
Cause still React.ChildrenOnly will return ReactElement

Kritik-J:
@SergeyRoyt please review

SergeyRoyt (changes requested):
Looks good, but some tests are still not green, let's try to fix them 🙏

Kritik-J:
@SergeyRoyt all checks are green now

SergeyRoyt (approved):
Nice! Thanks a lot! 🙂

PR #1662, opened 2023-10-16
feat(AlertBanner): add warning variation

talkor:
Screenshot 2023-10-16 at 11 57 56
https://monday.monday.com/boards/3532714909/pulses/4333987287

### DIFF
--- a/src/components/AlertBanner/AlertBanner.module.scss
+++ b/src/components/AlertBanner/AlertBanner.module.scss
@@ -25,6 +25,15 @@
   background-color: var(--inverted-color-background);
 }
 
+.warning {
+  background-color: var(--warning-color);
+  color: var(--text-color-fixed-dark);
+
+  .closeBtn {
+    color: var(--text-color-fixed-dark);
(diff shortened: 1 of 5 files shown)


src/components/AlertBanner/AlertBanner.module.scss
```diff
@@ -25,6 +25,15 @@
   background-color: var(--inverted-color-background);
 }
 
+.warning {
+  background-color: var(--warning-color);
+  color: var(--text-color-fixed-dark);
+
+  .closeBtn {
+    color: var(--text-color-fixed-dark);
+  }
```
talkor, line 34: I preferred to override this since the Button (and IconButton) are too complex to change for this, and if we go towards controlling colors with CSS rather than JS I think it's a good direction anyway

YossiSaadi: agree

YossiSaadi (approved):
seems cool to me 👍🏼

PR #1664, opened 2023-10-16
useMergeRefs - fix return type

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5335082110

Idea: change useMergeRefs return type to be RefObject, so we could have used it normally afterwards without type issue

It's a breaking change, so I'll do it in a separate hook - just testing for now


SergeyRoyt:
Tests are failing, but couldn't figure why so far - giving it a pause for now

SergeyRoyt:
Fixed in another PR

PR #1667, opened 2023-10-17
feat(InfoIcon): add InfoIcon component

talkor:
There's no spec for this component so I did the most basic usage, but anyway I don't think there are many use cases.
Screenshot 2023-10-17 at 17 03 59

We might want to consider making it a bit more generic by having types of "info", "help" etc because I saw some usages of the same component with a question mark, and clickable (probably be better to rename it to `Guidance` or anything else)

https://monday.monday.com/boards/3532714909/pulses/4711066643


src/components/InfoIcon/InfoIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+export interface InfoIconProps extends VibeComponentProps {
+  children: ReactNode;
+}
+
+const InfoIcon: VibeComponent<InfoIconProps, HTMLElement> = forwardRef(
+  ({ children, className, id, "data-testid": dataTestId }, ref) => {
+    const componentRef = useRef(null);
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    return (
+      <span
```
talkor, line 20: Tooltip is not a vibe component and doesn't get a ref so I had to wrap it in a span

plop/component/index.js
```diff
@@ -58,7 +58,7 @@ module.exports = plop => {
         path: "webpack/published-ts-components.js",
         pattern: /(\/\/ plop_marker:published-components)/g,
         template:
-          '  {{properCase componentName}}: "/src/components/{{properCase componentName}}/{{properCase componentName}}",'
+          '  {{properCase componentName}}: "components/{{properCase componentName}}/{{properCase componentName}}",'
```
SergeyRoyt, line 61: Nice

src/components/InfoIcon/InfoIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+import VibeComponentProps from "../../types/VibeComponentProps";
+import VibeComponent from "../../types/VibeComponent";
+import { getTestId } from "../../tests/test-ids-utils";
+import { ComponentDefaultTestId } from "../../tests/constants";
+import Tooltip from "../Tooltip/Tooltip";
+import Icon from "../Icon/Icon";
+import { Info } from "../Icon/Icons";
+
+export interface InfoIconProps extends VibeComponentProps {
+  children: ReactNode;
```
SergeyRoyt, line 12: `ElementContent`?

src/components/InfoIcon/InfoIcon.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const componentRef = useRef(null);
+    const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
+    return (
+      <span
+        ref={mergedRef}
+        className={className}
+        data-testid={dataTestId || getTestId(ComponentDefaultTestId.INFO_ICON, id)}
+      >
+        <Tooltip content={children}>
+          <Icon icon={Info} clickable={false} />
```
SergeyRoyt, line 26: Icon looks very small to me - need to check with Meirav
Also I think it should be focusable with tooltip popping up on focus and ariaLabel which corresponds to Tooltip content, if it's string

talkor:
Closing for now

PR #1670, opened 2023-10-18
feat(types): export general types

YossiSaadi:
pre-step to export all types from vibe

https://monday.monday.com/boards/3532714909/pulses/5342878337

### DIFF
--- a/src/types/positions.ts
+++ b/src/types/positions.ts
@@ -1,0 +1,3 @@
+import { DialogPosition as DialogPositions } from "../constants";
+
+export type DialogPosition = DialogPositions;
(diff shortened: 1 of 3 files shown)


src/types/positions.ts
```diff
@@ -0,0 +1,3 @@
+import { DialogPosition as DialogPositions } from "../constants";
+
+export type DialogPosition = DialogPositions;
```
talkor, line 3: Not sure about the effort or the risk but would it be possible to define the type in the Dialog and export the type and not the const?

talkor (approved):

PR #1681, opened 2023-10-18
feat(Button): allow disabling resize observer

talkor:
Button is using a resize observer because the loading state makes the button shrink. In order to make the width change smoother we use transition on `min-width` which is controlled by a CSS variable (`--element-width/height`).
The problem is that since we apply `min-width` on the button, when shrinking the container which the button is in - the resize observer won't work (it takes the min-width value into consideration). 

I created a [ticket](https://monday.monday.com/boards/3532714909/pulses/5350787985) to fix it properly though it might be quite tricky.

As a quick solution I added a boolean (opt-in) that disables the resize observer and added validation to prevent using both the loading state and the disbableResizeObserver.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -393,6 +393,14 @@ const Button: VibeComponent<ButtonProps, unknown> & {
   }
 );
 
+Button.propTypes = {
+  disableResizeObserver: function (props) {
+    if (props.loading && props.disableResizeObserver) {
+      return new Error('Do not use "disableResizeObserver" with loading state');
+    }
+  }
+};
+
 Button.defaultProps = {
   className: undefined,
   name: undefined,
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -92,8 +92,8 @@ export interface ButtonProps extends VibeComponentProps {
   insetFocus?: boolean;
   /** Specifies the tab order of an element */
   tabIndex?: number;
-  /** Controls resize observer for smooth transition between states. Do not use with loading state */
-  disableResizeObserver?: boolean;
+  /** Make the button width dynamic according to container width. */
+  dynamicWidth?: boolean;
 }
 
 const Button: VibeComponent<ButtonProps, unknown> & {
@@ -145,16 +145,15 @@ const Button: VibeComponent<ButtonProps, unknown> & {
       "data-testid": dataTestId,
       insetFocus,
       tabIndex,
-      disableResizeObserver
+      dynamicWidth
     },
     ref
   ) => {
     const overrideDataTestId = backwardCompatibilityForProperties([dataTestId, backwardCompatabilityDataTestId]);
     const buttonRef = useRef<HTMLButtonElement>(null);
-    const [hasSizeStyle, setHasSizeStyle] = useState(false);
 
     const updateCssVariables = useMemo(() => {
-      if (disableResizeObserver) {
+      if (dynamicWidth) {
         return NOOP;
       }
       return ({ borderBoxSize }: { borderBoxSize: { blockSize: number; inlineSize: number } }) => {
@@ -164,9 +163,8 @@ const Button: VibeComponent<ButtonProps, unknown> & {
         if (!buttonRef.current) return;
         buttonRef.current.style.setProperty("--element-width", `${width}px`);
         buttonRef.current.style.setProperty("--element-height", `${height}px`);
-        setHasSizeStyle(true);
       };
-    }, [buttonRef, disableResizeObserver]);
+    }, [buttonRef, dynamicWidth]);
 
     useResizeObserver({
       ref: buttonRef,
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.tsx
```diff
@@ -92,6 +92,8 @@ export interface ButtonProps extends VibeComponentProps {
   insetFocus?: boolean;
   /** Specifies the tab order of an element */
   tabIndex?: number;
+  /** Controls resize observer for smooth transition between states. Do not use with loading state */
```
YossiSaadi, line 95: maybe describe a bit more why it is needed

YossiSaadi (approved):
looks fine by me, but we should maybe back this up with an extra line of why this was added
and add this to our imaginary Vibe 3.0 "to remove" list 😅

src/components/Button/Button.tsx
shlomitc: 1. I’d name the prop in a meaningful way which doesn’t explain the technology (resize observer) but rather what is the expected behavior (dynamic width?)
 2. It’s worth documenting such things in stories even though they are mostly technical, developers want to see how the code affects the design
 3. If we wish to make this the default behavior, we should start introducing a deprecation mechanism the console-logs the updated behavior. For example, if in the next major version we would like to eliminate it completely - a depredation log should say that in the next major version this prop will be turned on by default and we suggest migrating already to the new behavior. If the case is that it can be turned on/off - then the message should say what would be it’s default value (true?) and ask to pass explicitly the wanted behavior (true/false).

talkor: > I’d name the prop in a meaningful way which doesn’t explain the technology (resize observer) but rather what is the expected behavior (dynamic width?)

I agree the prop name should be meaningful, but it aims to fix an existing bug that is related to the resize observer. For example either way the button will have a dynamic width, but in combination with the loading state there is the bug described (where the button doesn't shrink), maybe `disableDynamicResizing`? It might also mislead I think

> It’s worth documenting such things in stories even though they are mostly technical, developers want to see how the code affects the design

I agree, I added a JSDoc comment so it will appear in the props table, but I'm wondering how would we be able to describe such case in a story, do you have any idea? Because having the new prop as true in addition to `loading` will result in a bad example of the UI

> If we wish to make this the default behavior, we should start introducing a deprecation mechanism the console-logs the updated behavior. For example, if in the next major version we would like to eliminate it completely - a depredation log should say that in the next major version this prop will be turned on by default and we suggest migrating already to the new behavior. If the case is that it can be turned on/off - then the message should say what would be it’s default value (true?) and ask to pass explicitly the wanted behavior (true/false).

Sounds cool, actually we wanted to remove the resize observer completely in the next major version and solve this case somehow with CSS only, so this prop will be deprecated hopefully

shlomitc: I’ve looked into your PR description again and suggest to name it “disableLoaderSmoothing” in case we wish to keep the bad behavior as default, or “loaderSmoothing” if we want to break the current behavior.

I kind of tending towards removing this logic completely if it makes more problems than benefits

src/components/Button/Button.tsx
```diff
@@ -330,6 +330,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
         <button {...buttonProps}>
           <span className={styles.loader}>
             <Loader svgClassName={styles.loaderSvg} />
+            <span className={styles.textPlaceholder}>{children}</span>
```
talkor, line 333: To keep the loader size the same before the loading state. It gets `opacity: 0`

src/components/Button/Button.tsx
```diff
@@ -142,25 +144,27 @@ const Button: VibeComponent<ButtonProps, unknown> & {
...
+      dynamicWidth
     },
     ref
   ) => {
     const overrideDataTestId = backwardCompatibilityForProperties([dataTestId, backwardCompatabilityDataTestId]);
     const buttonRef = useRef<HTMLButtonElement>(null);
-    const [hasSizeStyle, setHasSizeStyle] = useState(false);
 
     const updateCssVariables = useMemo(() => {
+      if (dynamicWidth) {
```
talkor, line 156: I think the prop name now is more meaningful

talkor:
> Button is using a resize observer because the loading state makes the button shrink. In order to make the width change smoother we use transition on `min-width` which is controlled by a CSS variable (`--element-width/height`). The problem is that since we apply `min-width` on the button, when shrinking the container which the button is in - the resize observer won't work (it takes the min-width value into consideration).
> 
> I created a [ticket](https://monday.monday.com/boards/3532714909/pulses/5350787985) to fix it properly though it might be quite tricky.
> 
> As a quick solution I added a boolean (opt-in) that disables the resize observer and added validation to prevent using both the loading state and the disbableResizeObserver.
> 
> https://monday.monday.com/boards/3532714909/pulses/5350721960

Apparently the transition applied only to the loading state since it shrunk but it wasn't the required behavior. I changed the loading state to be the same width as the original state of the button, so the transition is redundant now. Now the `dynamicWidth` can disable the resizeObserver so that (when needed, opt-in), button will be able to grow/shrink according to the container size. We still need the resize observer to keep the button size when text changes, but it's not relevant for cases where the button is in full width for example. @orrgottlieb WDYT?

src/components/Button/Button.tsx
talkor: Wondering whether we should add `aria-hidden` here

PR #1686, opened 2023-10-19
fix(Button): fix grow/shrink behavior with loading state

talkor:
As requested by design, the loader remains the same size when switched to loading state instead of shrinking. Same for text changes in the button, if it changes with longer text, the button will grow in size, if it changes with shorter text it will remain the same size.
But, container width won't affect the button size whatsoever anymore. The Resize Observer was removed and now we simply keep the button original size in a CSS variable. This is also by design as it shouldn't be affected by the container. 

Note: this will not fix the original issue of the button being responsive to the container as it's a wrong usage for the button. We will suggest using the MenuItem for that case.

Screenshot 2023-10-19 at 14 54 38

This replaces #1681

### DIFF
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -5,7 +5,6 @@
 $disabled-on-primary-color-opacity: 0.5;
 
 .button {
-  --loader-padding: 8px;
   outline: none;
   border: none;
   height: auto;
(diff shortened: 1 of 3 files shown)


src/components/Button/Button.module.scss
```diff
@@ -25,7 +24,18 @@ $disabled-on-primary-color-opacity: 0.5;
   user-select: none;
   /* Will be updated dynamically after render by js */
   --element-width: 32;
-  --element-height: 32;
+  min-width: var(--element-width);
+  width: fit-content;
```
YossiSaadi, line 28: is there really need in min-width and initial 32 width? it's also without `px` so it probably has no affect and it is immediately overriden.

YossiSaadi (approved):
looks cool, I had one comment

talkor:
Please ignore, will continue in https://github.com/mondaycom/monday-ui-react-core/pull/1681

PR #1693, opened 2023-10-22
chore: fix < Tip /> imports to be from vibe-storybook-components

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5370191036


talkor (approved):
Awesome!

Issue #1679, opened 2023-10-18
feat: < Dialog /> should throw console.error if containerSelector is not found


SergeyRoyt:
[Monday ticket](https://monday.monday.com/boards/3532714909/pulses/5350072279)

arianxq:
Hi, I would like to work on this.

SergeyRoyt:
Hey, @arianxq !
Thanks, good luck with that!

Kritik-J:
@SergeyRoyt As the assignee has not created a PR since last week, can you assign me this issue?

arianxq:
@Kritik-J
I had intended to work on this issue, but I see that you've already submitted a pull request. In light of that, I respect your efforts and contributions, you can take this issue. Best of luck with your PR!

SergeyRoyt:
Closed thanks to @Kritik-J !

Sorry, @arianxq 🙏

PR #1703, opened 2023-10-25
feat: < Dialog /> should throw console.error if containerSelector is not found

Kritik-J:
fixes: #1679

### DIFF
--- a/src/components/Dialog/Dialog.tsx
+++ b/src/components/Dialog/Dialog.tsx
@@ -280,14 +280,8 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
       console.error("Dialog: containerSelector is not provided or is falsy. Dialog may not be correctly positioned.");
       return document.body;
     }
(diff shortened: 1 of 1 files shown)


src/components/Dialog/Dialog.tsx
```diff
@@ -277,9 +277,17 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
   getContainer() {
     const { containerSelector } = this.props;
     if (!containerSelector) {
+      console.error("Dialog: containerSelector is not provided or is falsy. Dialog may not be correctly positioned.");
```
SergeyRoyt, line 280: This should be reverted - it's fine if containerSelector is undefined or null
The error should be thrown only if it's not null, but such element doesn't exist on the page - as you did below

SergeyRoyt:
You did the opposite of what I expected) Please, read what I wrote once more

SergeyRoyt (approved):
Looks good! Thanks!

Issue #1673, opened 2023-10-18
fix: < Text / > - when more than one row ellipsis is applied - the tooltip doesn't show


SergeyRoyt:
[Monday ticket](https://monday.monday.com/boards/3532714909/pulses/5070129086)

https://style.monday.com/?path=/docs/text-text--overflow

viditagrawal56:
Hey @SergeyRoyt I was looking into the codebase to see if I could find a fix for this issue, and I think i found one.

![image](https://github.com/mondaycom/monday-ui-react-core/assets/52532308/57b97cf4-b3ca-43ee-8729-b4dd2be51f04)

In the function useTooltipProps, the ignoreHeightOverflow is set to true, as shown in the above screenshot.
From what I understand ignoreHeightOverflow ignores the vertical oveflow of the text inside the Text component which I think is the issue.
So setting the ignoreHeightOverflow to false should resolve the error I think.
Am I missing something, or the solution is as simple as setting the ignoreHeightOverflow flag to false?

SoumyadiptoPal:
I want to work on this issue. Please assign this to me!!

viditagrawal56:
> Hey @SergeyRoyt I was looking into the codebase to see if I could find a fix for this issue, and I think I found one.
> 
> ![image](https://user-images.githubusercontent.com/52532308/276364483-57b97cf4-b3ca-43ee-8729-b4dd2be51f04.png)
> 
> In the function useTooltipProps, the ignoreHeightOverflow is set to true, as shown in the above screenshot. From what I understand ignoreHeightOverflow ignores the vertical overflow of the text inside the Text component which I think is the issue. So setting the ignoreHeightOverflow to false should resolve the error I think. Am I missing something, or the solution is as simple as setting the ignoreHeightOverflow flag to false?

@SergeyRoyt

viditagrawal56:
@SergeyRoyt Ok I'll give it a try :)

viditagrawal56:
Hey @SergeyRoyt Just wanted to confirm that I am on the right track.....

1.   I need to create a new context say `TextContext` with `ignoreHeightOverflow` set to `false` and export it from a new file say 
`TextContext.ts` which is placed in a new folder `utils` inside the root `Text` component folder.

2.   After that import the context in `Text.tsx` and wrap the `Typography` component inside the code with `<TextContext.Provider value={{ updateFocusedItem }}>`

Am I on the write track or am I missing something?

[...]

PR #1706, opened 2023-10-25
Fixes: #1673 BUG Tooltip not visible when more than one row ellipsis is applied to Text component 

viditagrawal56:
Fixes : #1673 

#### Basic

- [X] PR has description.
- [X] Component is written in TypeScript.

#### Style

- [ ] Styles are added to `NewComponent.modules.scss` file inside the `NewComponent` folder.
- [ ] Styles are defined using [monday-ui-style](https://github.com/mondaycom/monday-ui-style) tokens
- [ ] Displays correctly in all the themes

#### Storybook

- [ ] Stories were added to `/src/components/NewComponent/__stories__/NewComponent.stories.mdx` file.
- [ ] Stories implemented using [vibe-storybook-components](https://github.com/mondaycom/vibe-storybook-components) components and tokens.

#### Tests

#### Accessibility

- [ ] Component is accessible.
- [ ] Component is compatible with [Vibe Accessibility Guidelines](https://style.monday.com/?path=/docs/foundations-accessibility--page)


viditagrawal56:
Hey @SergeyRoyt, I understand that you must be busy,
I made this draft PR
Please let me know if any changes are required! :)

viditagrawal56:
> Hey @SergeyRoyt, I understand that you must be busy that's why you were not replying to my messages 😅. I made this draft PR, Please let me know if any changes are required whenever you are free! :)

@SeregyRoyt

src/components/Text/utils/TextContext.ts
SergeyRoyt, line 1: Let's call this one `TypographyContext` and move it `Typography` folder, cause it's related to Typography, not just to Text

SergeyRoyt: Let's fix lint issues here

src/components/Text/Text.tsx
```diff
@@ -32,9 +33,11 @@ const Text: VibeComponent<TextProps, HTMLElement> & {
       ...typographyProps
     },
     ref
-  ) => {
+    ) => {
+    const { ignoreHeightOverflow } = useContext(TextContext);
```
SergeyRoyt, line 37: That's not right. You should call `useContext` in the `Typography` component, then pass this value into `useTooltipProps` which should pass it further to `useIsOverflowing` hook

src/components/Text/Text.tsx
```diff
@@ -32,9 +33,11 @@ const Text: VibeComponent<TextProps, HTMLElement> & {
...
     const overrideEllipsis = ellipsis ?? element !== "p";
     return (
+      <TextContext.Provider value={{ignoreHeightOverflow}}>
```
SergeyRoyt, line 40:
```suggestion
      <TypographyContext.Provider value={{ignoreHeightOverflow: false}}>
```
And to make it explicit let's also provide TypographyContext value in `Heading` component, but there it should be 
`{ignoreHeightOverflow: true}` for now

SergeyRoyt (changes requested):
Hey, @viditagrawal56 !
Sorry for not a very quick respond!
Wrote you a few notes, feel free to ask if you need help with anything

Basically what I want us to achieve here is passing different kind of context from `Text` (`{ignoreHeightOverflow: false}`) and `Heading` (`{ignoreHeightOverflow: true}`). Then consuming this context in `Typography` component and make difference in overflowing logic due to it. 

Basically you could have achieved the same just by using props, but what we do here is most likely a temp fix (some work will still need to be done in `Heading` component to fix overflow for the case of multiple lines), so we don't want to spoil our components API by adding new props - so we just pass them kinda "behind the scenes"

viditagrawal56:
@SergeyRoyt Sorry for the mistakes 🙏 
I have made all the changes as per your notes
Let me know if more changes are needed!
Forgot to turn on prettier last time 😅, but all the lint issues are fixed now.

I think I also need to update the snapshots by running `npm run test:update` right?

src/components/Text/Text.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { forwardRef, ReactNode } from "react";
+import React, { forwardRef, ReactNode, useContext } from "react";
```
SergeyRoyt, line 1: Only one lint issue left to fix

SergeyRoyt:
Looks great, exactly what's needed to be done! 💪
Let me just test it on our production environment to make sure we are not breaking anything 🙏

SergeyRoyt (approved):
Approved! Thanks a lot for the great job, @viditagrawal56!! 👑👑

PR #1707, opened 2023-10-25
fix(Button): adjust loading state to initial button size

talkor:
Screenshot 2023-10-25 at 20 00 40

### DIFF
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -23,7 +23,8 @@ $disabled-on-primary-color-opacity: 0.5;
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
-  min-width: var(--element-width, auto);
+  --element-width: auto;
+  min-width: var(--element-width);
 
   .loader {
     height: 100%;
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -23,10 +23,7 @@ $disabled-on-primary-color-opacity: 0.5;
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
-  /* Will be updated dynamically after render by js */
-  --element-width: 32px;
-  --element-height: 32px;
-  min-width: var(--element-width);
+  min-width: var(--element-width, 6px);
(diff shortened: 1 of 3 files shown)


shlomitc:
Should the button have a min-width? In very short text it makes sense the button will look a bit larger
image

talkor:
> Should the button have a min-width? In very short text it makes sense the button will look a bit larger image

The min-width hasn't changed in this PR but I will verify with Meirav

shlomitc:
also - what should happen if the success text is smaller than the children text?
From my test - it shrinks to the success text, but I think we said it should stay the same size

talkor:
> also - what should happen if the success text is smaller than the children text? From my test - it shrinks to the success text, but I think we said it should stay the same size

We have a story for this, the text becomes shorter and the button remains the same size, this also hasn't changed
Screenshot 2023-10-26 at 9 21 42

src/components/Button/Button.module.scss
```diff
@@ -26,6 +26,21 @@ $disabled-on-primary-color-opacity: 0.5;
   /* Will be updated dynamically after render by js */
   --element-width: 32;
   --element-height: 32;
+  min-width: var(--element-width);
```
talkor, line 29: Moved from line 50

shlomitc: I don't see the effect of this. are you sure it's working? it's missing the `px`

shlomitc: in addition, the `element-width` will overridden after being calculated in the js

talkor: Oh yeah your'e right, the definition is missing the `px`, it is being overridden with JS (`Button.tsx:162-163`) so it's used as a default/fallback value, I will add `px` to the definition of the variable

talkor: It did not work before as the fallback value of 32px as the value was illegal, now when I changed it to be a valid value if messes up some component that use Button with a smaller width than 32px, so I think we should remove the default value at all

shlomitc (approved):

PR #1708, opened 2023-10-26
fix: <Dropdown> multi single line - counter options display

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5397103322

Issue: one of the options wasn't displaying not in input, not under counter
Solution: recalculate useHiddenOptionsData AFTER counter rerender, not before


src/components/Dropdown/__stories__/dropdown.stories.mdx
SergeyRoyt, line 1: **_TODO remove demo story_**

src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.jsx
```diff
@@ -35,9 +36,13 @@ export default function Container({ children, selectProps, ...otherProps }) {
     ref,
     chipClassName,
     chipWrapperClassName,
-    selectedOptionsCount: selectedOptions.length
+    selectedOptionsCount: selectedOptions.length,
+    isCounterShown
   });
-  const isCounterShown = hiddenOptionsCount > 0;
+
+  useEffect(() => {
```
talkor, line 43: It feels a bit weird to me, because there's an external state that is being sent to the hook and is dependent on it, and based on the result we use an effect on this value
I think the root cause of this issue is the CSS, there's a margin on the field which doesn't let another chip to be viewed, it's in wrapped in the next line, so the calculation is supposed to be ok. If you remove the margin it works well, maybe worth taking a look at the CSS first, WDYT?

SergeyRoyt: The problem as I understand it, is that it calculates amount of Chips to hide before rendering the Counter so not taking it in consideration, which makes the issue in this specific case, this trick makes it recalculate again after everything was rendered fixing the previous result
So I have doubts that it's a CSS issue, though it might fix it for this specific case, as the Chips just fits after the change

SergeyRoyt: Which margin you remove anyway? I'll try to play with it too

talkor (approved):

PR #1711, opened 2023-10-26
chore: replace mixins with exported mixins from monday-ui-style

talkor:
- Use exported mixins from `monday-ui-style`
- Remove unused mixins
- Remove redundant mixins import

There are some other mixins that can be a bit optimized but will be created in a different PR.

https://monday.monday.com/boards/3532714909/pulses/4982541657


src/styles/typography.scss
SergeyRoyt, line 1: I believe functions like `capitalize`, `camelize` and the rest are exported from monday-ui-style/dist/functions, so we can refactor this even better 💪

SergeyRoyt (approved):
Great job! 🤩
It warms my heart to see such cleanup ^^

src/styles/typography.scss
talkor: Ohh nice thanks, missed that, removing

PR #1716, opened 2023-10-29
docs: extract AlertBanner template to stories file

talkor:
The issue was that the condition in `AlertBanner.tsx:109` where the condition identified the children as MDX element and not AlertBanner Text, the ellipsis actually works well, there's even a story for that.
https://monday.monday.com/boards/3532714909/pulses/5382950355
Screenshot 2023-10-29 at 16 47 11


SergeyRoyt (approved):
Nice catch

PR #1726, opened 2023-11-01
feat: new EditableHeading with new typography

talkor:
https://monday.monday.com/boards/3532714909/pulses/5266220954


shlomitc (changes requested):
image
This should stay `EditableHeading` and not `LegacyEditableHeading`

shlomitc (approved):

PR #1727, opened 2023-11-01
feat: create a new EditableHeading component (with new typography styles)

talkor:
Screenshot 2023-10-31 at 19 58 25
https://monday.monday.com/boards/3532714909/pulses/5266220954


src/components/EditableHeading/HiddenInputPlaceholder.module.scss
```diff
@@ -0,0 +1,7 @@
+.hiddenInputPlaceholder {
+  position: absolute;
```
talkor, line 2: I can't use the `hidden-element` mixin here because I need the width

src/components/EditableHeading/__stories__/EditableHeading.stories.mdx
```diff
@@ -0,0 +1,66 @@
...
+  </Story>
+</Canvas>
+
+## Props
+
+<ArgsTable story="Overview" />
+
+## Usage
+
+<UsageGuidelines guidelines={[]} />
```
talkor, line 54: The old guidelines are not relevant anymore, I'll talk to Meirav

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,168 @@
...
+      <div
+        ref={mergedRef}
+        id={id}
+        aria-label={ariaLabel}
+        data-testid={dataTestId || getTestId(ComponentDefaultTestId.EDITABLE_HEADING, id)}
+        className={cx(styles.editableHeading, className)}
+      >
+        {isEditing && editable ? (
+          <>
+            <HiddenInputPlaceholder className={inputClassNames} value={inputValue} onChange={setInputWidth} />
```
talkor, line 137: The main difference between this and the old one is that the input is now an inline element and not a black element, so since html inputs cannot have a dynamic width that fits the text we use this hack of a hidden placeholder that defines the width for the input. This is a very common hack and it's being used in libraries such as `react-input-autosize`.

src/components/EditableHeading/EditableHeading.module.scss
```diff
@@ -0,0 +1,36 @@
+@import "../../styles/typography";
+@import "~monday-ui-style/dist/mixins";
+
+@include create-title-classes;
+
+.editableHeading {
+  @include ellipsis;
```
shlomitc, line 7: why is this needed? aren't you getting it from the `Heading` component?

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+export interface EditableHeadingProps extends VibeComponentProps {
+  value: string;
+  type?: HeadingType;
+  weight?: HeadingWeight;
+  editable?: boolean;
+  onChange?: (value: string) => void;
+  focusOnMount?: boolean;
+  selectOnMount?: boolean;
+  ariaLabel?: string;
```
shlomitc, line 24: add description per prop for its documenation

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,168 @@
...
+    {
+      id,
+      className,
+      "data-testid": dataTestId,
+      value,
+      type = Heading.types.H1,
+      weight = HeadingWeight.NORMAL,
+      editable = true,
+      onChange = () => {},
+      focusOnMount = true,
```
shlomitc, line 41: should probably be `false`

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,168 @@
...
+  (
+    {
+      id,
+      className,
+      "data-testid": dataTestId,
+      value,
+      type = Heading.types.H1,
+      weight = HeadingWeight.NORMAL,
+      editable = true,
+      onChange = () => {},
```
shlomitc, line 40: I guess you can keep it unassigned

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,168 @@
...
+} = forwardRef(
+  (
+    {
+      id,
+      className,
+      "data-testid": dataTestId,
+      value,
+      type = Heading.types.H1,
+      weight = HeadingWeight.NORMAL,
+      editable = true,
```
shlomitc, line 39: I usually prefer props that are "off" by default. meaning that if I want to turn something on, I activate it instead of making something inactive

for example, prefer:
`<EditableHeading readOnly/>`
instead of
`<EditableHeading editable={false}/>`

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,168 @@
...
+        onClick: toggleEditMode,
+        id,
+        role: "button"
+      },
+      mergedRef
+    );
+
+    function focus() {
+      if (inputRef.current) {
+        requestAnimationFrame(() => {
```
shlomitc, line 102: why is this needed?

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    useLayoutEffect(() => {
+      if (focusOnMount) {
+        focus();
+      }
+      if (selectOnMount) {
+        select();
+      }
+    }, [focusOnMount, isEditing, selectOnMount]);
```
shlomitc, line 121: this means the hook will run even if `focusOnMount` changes, which isn't the expected behavior.
You might want to split it into two effects, one that runs once on mounting, and another only when the `isEditing` changes

src/components/EditableHeading/EditableHeading.module.scss
```diff
@@ -0,0 +1,36 @@
+@import "../../styles/typography";
+@import "~monday-ui-style/dist/mixins";
+
+@include create-title-classes;
```
shlomitc, line 4: what is this exactly?

src/components/Modal/__stories__/modal.stories.mdx
```diff
@@ -410,7 +410,7 @@ clicking outside. Use this variant in case of sensitive or important messages, a
             {/** Please follow the design guidelines when implementing custom title **/}
             <ModalHeader
               description={"Description text goes here"}
-              title={<LegacyEditableHeading type={LegacyEditableHeading.types.h2} placeholder={"Modal title"} />}
+              title={<EditableHeading type={EditableHeading.types.H2} value={"Modal title"} />}
```
shlomitc, line 413: it behaves differently from the previous component. It should span to 100%

now:
image

before:
image

src/components/EditableHeading/EditableHeading.module.scss
```diff
@@ -0,0 +1,36 @@
...
+
+.editableHeading {
```
shlomitc, line 6: I took this example from the modal with the editable header. 
It should be aligned the same as the description text, and when hovered, the border should be applied outside this line. you can talk with Meirav if it is not clear

image

shlomitc (changes requested):
once PR #1726 is merged - please add also the deprecation to the legacy component, the same as you can see in the `LegacyHeading` page
image

src/components/EditableHeading/EditableHeading.module.scss
talkor: This creates the new typography classes

src/components/Modal/__stories__/modal.stories.mdx
talkor: That's intentional, the new spec defines it as an inline element
Screenshot 2023-11-01 at 12 22 26

See https://github.com/mondaycom/monday-ui-react-core/pull/1727#discussion_r1378548622

src/components/EditableHeading/EditableHeading.module.scss
talkor: I'll make sure with Meirav, the old component behaves the same, if we shift it to the left it means the the input will overflow

src/components/EditableHeading/EditableHeading.tsx
talkor: That's the default in the old component, I'll make sure with Meirav though

src/components/EditableHeading/EditableHeading.module.scss
shlomitc: Ok, I see. It’s for the input element. Just add a comment there to make it more clear as it looks like a side affect that no one uses. Maybe even using it inside some class that you use to explain the connecting

src/components/Modal/__stories__/modal.stories.mdx
shlomitc: As long as it’s verified with Meirav, we’re good

src/components/EditableHeading/EditableHeading.module.scss
shlomitc: The input should probably contain some padding so the text inside would be aligned, but the input frame will overflow

talkor: Yes but it might be a big deal to shift it in `overflow: hidden` containers like this one, we'll have to use a portal for this
Screenshot 2023-11-01 at 12 45 58

shlomitc: Or we can solve this with some padding games. Give the content some padding and reduce it from the container

talkor:
Accidentally deleted your comment about the focus hook :( Will resolve it

src/components/EditableHeading/EditableHeading.module.scss
talkor: Wouldn't that be too hacky? It means I'll have to adjust the `Modal*` components' width just for this case. And it would have to be adjusted as well in other overflowed containers

shlomitc: Yes, it is an issue that will appear in any container that is overflowing, and that the content sticks to the side.
But I don't think it's that hacky... it's just how you split the padding. it might get complex if all the rendered elements (header, description, content) aren't wrapped with any wrapper from the component.

But let's give it a shot and see if it goes bad or not

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -0,0 +1,177 @@
...
+import useClickableProps from "../../hooks/useClickableProps/useClickableProps";
+
+export interface EditableHeadingProps extends VibeComponentProps {
+  /** Heading text */
+  value: string;
+  /** Heading type */
+  type?: HeadingType;
+  /** Heading weight */
+  weight?: HeadingWeight;
+  /** onChange callback */
```
shlomitc, line 23: I meant here to something that explains better what it does. for example in this case, `will be called whenever the current value changes to a non-empty value, `

src/components/EditableHeading/EditableHeading.tsx
shlomitc: It's more of a business-decision, so I'd suggest keeping the component non-opinionated and allow the consumer to decide if he wants to focus

src/components/EditableHeading/__tests__/EditableHeading-interactions.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+function getHeading(canvas: Canvas) {
+  return getByTestId(canvas, ComponentDefaultTestId.CLICKABLE);
+}
+
+function getInput(canvas: Canvas) {
+  return getByRole(canvas, "input");
+}
+
+async function testBasicEditing(canvas: Canvas) {
```
shlomitc, line 28: I try to name the tests by what they should do.
For example, `editAndChangeToValidText`

That forces me to write tests that are simple and doesn't mix multiple tests in one

src/components/EditableHeading/__tests__/EditableHeading-interactions.tsx
```diff
@@ -0,0 +1,73 @@
...
+  clickElement(compponent);
+
+  const input = getInput(canvas);
+  await clearText(input);
+  await typeText(input, text);
+  expect(input).toHaveAttribute("value", text);
+
+  await resetFocus();
+  const heading = getHeading(canvas);
+  expect(heading).toHaveTextContent(text);
```
shlomitc, line 39: worth splitting to another test, just so chromatic will catch both

shlomitc (changes requested):

src/components/EditableHeading/EditableHeading.tsx
talkor: Talked to Meirav about this and she said that editing the text is the purpose so it should be focused by default, and also she suggested to keep it the same as the old component's default

src/components/EditableHeading/__tests__/EditableHeading-interactions.tsx
talkor: What do you mean? I want to check that the text is changing when you blur after you change it, and the next test also checks the case where you just clear the text

src/components/EditableHeading/EditableHeading.tsx
shlomitc: None of the html elements are focused by default. This is always an opt-in feature and I can think of product cases that doesn’t require editing by default (board header). So let’s avoid from making it on by default

src/components/EditableHeading/__tests__/EditableHeading-interactions.tsx
shlomitc: So the first text will be the blur only. The second test would be a component which already has a value set and then clear it

src/components/EditableHeading/EditableHeading.module.scss
talkor: I added an opt-out prop (`alignWithText` - if you have a better name LMK) that adds negative margin to the component so it will be aligned with the text. In case of hidden overflowing containers users can disable it. Specifically the `ModalHeader` has a different issue when providing a custom title but the might be a breaking change, I suggest discussing it further

src/components/EditableHeading/EditableHeading.tsx
talkor: Just to be clear - the `onMount` refers to the editing mode, not the view mode. E.g. you can on the component, the input is shown, and only then you focus on the input

shlomitc: So why do you need to props. If it’s in edit mode - isn’t it always focused?

talkor: Which two props? I'm referring only to the focusOnMount prop, it can be disabled when you want to control the focus externally for example, or if you want to select the text on mount which is also a feature we provide

src/components/Modal/ModalHeader/ModalHeader.tsx
```diff
@@ -82,19 +87,20 @@ const ModalHeader: FC<ModalHeaderProps> = ({
...
-        id={id}
-        maxLines={2}
-        data-testid={dataTestId || getTestId(ComponentDefaultTestId.MODAL_HEADER, id)}
-        className={titleClassName}
-      >
-        {icon && (
-          <span className={cx(styles.icon, iconClassName)}>
-            <Icon icon={icon} iconType={Icon.type.SVG} iconSize={iconSize} ignoreFocusStyle clickable={false} />
-          </span>
+      <div data-testid={dataTestId || getTestId(ComponentDefaultTestId.MODAL_HEADER, id)}>
```
shlomitc, line 90: can't you simply set the data-testid on the root div of the component? do we really need this extra div?

src/components/EditableHeading/__stories__/EditableHeading.stories.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+## Usage
+
+<UsageGuidelines
+  guidelines={[
+    "Use when the title (Heading) is editable and you want to edit it",
+    "The component takes 100% of width and height of its container"
+  ]}
+/>
+## Variants
```
shlomitc, line 60: probably needs an empty line before
image

src/components/EditableHeading/__stories__/EditableHeading.stories.mdx
```diff
@@ -0,0 +1,70 @@
...
+## Props
+
+<ArgsTable story="Overview" />
+
+## Usage
+
+<UsageGuidelines
+  guidelines={[
+    "Use when the title (Heading) is editable and you want to edit it",
+    "The component takes 100% of width and height of its container"
```
shlomitc, line 57: is it? 
image

talkor: I agree but I asked Meirav specifically about this and she said that even though the header itself is an inline element, the entire component takes the entire block which is true

shlomitc (approved):
looks good

PR #1729, opened 2023-11-02
fix(MultiValueContainer): align items vertically

talkor:
https://monday.monday.com/boards/3532714909/pulses/5392143623
Screenshot 2023-11-02 at 14 26 56

### DIFF
--- a/src/components/Dropdown/__stories__/dropdown.stories.mdx
+++ b/src/components/Dropdown/__stories__/dropdown.stories.mdx
@@ -8,7 +8,7 @@ import {
   MENU,
   SPLIT_BUTTON
 } from "../../../storybook/components/related-components/component-description-map";
-import { overviewPlaySuite } from "../__tests__/Dropdown.interactions";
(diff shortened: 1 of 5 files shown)


shlomitc (commented):
image
looks like the arrow is misplaced.
Also - let's add a visual test for that. we need also to simulate what happens when there are multiple items (scrolling)

talkor:
> image looks like the arrow is misplaced. Also - let's add a visual test for that. we need also to simulate what happens when there are multiple items (scrolling)

Fixed
About the visual test, the dialog is opened only with a click, we can't trigger it pragmatically, do you have any idea how to get a visual test anyway?

shlomitc:
> > image looks like the arrow is misplaced. Also - let's add a visual test for that. we need also to simulate what happens when there are multiple items (scrolling)
> 
> Fixed About the visual test, the dialog is opened only with a click, we can't trigger it pragmatically, do you have any idea how to get a visual test anyway?

Interaction testing should do the trick, no? I think chromatic should catch this as well

shlomitc (approved):

PR #1733, opened 2023-11-06
feat: <Dropdown> add filterOption prop for custom filter

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5457990824

Provides ability to override default filterOption behaviour - https://react-select.com/advanced#custom-filter-logic

### DIFF
--- a/src/components/Dropdown/Dropdown.jsx
+++ b/src/components/Dropdown/Dropdown.jsx
@@ -3,7 +3,7 @@ import { ComponentDefaultTestId, getTestId } from "../../tests/test-ids-utils";
 import cx from "classnames";
 import { SIZES } from "../../constants/sizes";
 import React, { forwardRef, useCallback, useMemo, useRef, useState } from "react";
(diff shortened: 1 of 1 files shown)


src/components/Dropdown/Dropdown.jsx
```diff
@@ -396,7 +399,8 @@ Dropdown.defaultProps = {
   readOnly: false,
   isLoading: false,
   loadingMessage: undefined,
-  ariaLabel: undefined
+  ariaLabel: undefined,
+  filterOption: undefined
```
talkor, line 403: It's already `undefined` by default, no?

talkor (approved):

src/components/Dropdown/Dropdown.jsx
SergeyRoyt: Yeah, should be. I guess just for explicitly...
We'll refactor when we do TS migration

PR #1739, opened 2023-11-09
fix(Button): differentiate different states of the button with a key

talkor:
This [issue](https://github.com/facebook/react/issues/11538) was a very big pointer and helped reproduce this issue because it gives the same error message (and might be also the same issue - need further investigation to verify).

How to reproduce:
1. Install Google Translate plugin in [Chrome](https://chrome.google.com/webstore/detail/google-translate/aapbdbdomjkkjkaonfhkkikfgjllcleb/related).
2. Translate a page with a button (that changes to a loading state on click) to any language.
3. Click on the button, notice the page crushes.

Conjecture: 
Having three different root nodes that are returned from the Button confuses React understanding the changes in the component since all these root nodes get the same props.

The simplest solution (suggested by @orrgottlieb 👑 ) is to add a `key` for the different states.
Moving forward we might want to merge the different root nodes into a single one to make the component more coherent.
Also, since it caused the button to re-render and lose the CSS variable with the saved button width, I used a text placeholder instead, so the resize observer is redundant now. This is the same technique that is used in the loading state.

https://monday.monday.com/boards/3532714909/pulses/5303882985

### DIFF
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -32,10 +32,20 @@ $disabled-on-primary-color-opacity: 0.5;
       height: 100%;
       margin: 0;
     }
+  }
 
-    .textPlaceholder {
-      opacity: 0;
-    }
+  .textPlaceholder {
+    opacity: 0;
+    height: 0;
+    visibility: hidden;
+    white-space: pre;
+  }
+
+  .successContent {
+    position: absolute;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
   }
 }
 
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -41,11 +41,17 @@ $disabled-on-primary-color-opacity: 0.5;
     white-space: pre;
   }
 
-  .successContent {
-    position: absolute;
+  &.success {
     display: inline-flex;
+    flex-direction: column;
     align-items: center;
     justify-content: center;
+
+    .successContent {
+      display: inline-flex;
+      align-items: center;
+      justify-content: center;
+    }
   }
 }
 
--- a/src/components/Button/Button.module.scss
+++ b/src/components/Button/Button.module.scss
@@ -22,8 +22,7 @@ $disabled-on-primary-color-opacity: 0.5;
   -moz-user-select: none;
   -ms-user-select: none;
   user-select: none;
-  --element-width: auto;
-  min-width: var(--element-width);
+  min-width: auto;
 
   .loader {
     height: 100%;
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -358,7 +358,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
     }
 
     return (
-      <button {...buttonProps}>
+      <button {...buttonProps} key={`${id}-button`}>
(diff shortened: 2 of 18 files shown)


orrgottlieb (approved):

src/components/Button/Button.tsx
```diff
@@ -149,17 +149,16 @@ const Button: VibeComponent<ButtonProps, unknown> & {
   ) => {
     const overrideDataTestId = backwardCompatibilityForProperties([dataTestId, backwardCompatabilityDataTestId]);
     const buttonRef = useRef<HTMLButtonElement>(null);
-    const [hasSizeStyle, setHasSizeStyle] = useState(false);
```
talkor, line 152: This was unused, related to old CSS of the loader before...

src/components/Button/Button.tsx
```diff
@@ -265,7 +262,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
         className: classNames,
         name,
         onMouseUp,
-        style,
+        style: { minWidth: `${buttonWidth.current}px`, ...style },
```
talkor, line 265: I intentionally put the extra style after in case someone overrode it or intending to override it, WDYT?

src/components/Button/Button.tsx
```diff
@@ -339,26 +317,31 @@ const Button: VibeComponent<ButtonProps, unknown> & {
...
+                className={cx({
+                  [styles.leftIcon]: !!successText
+                })}
+                ignoreFocusStyle
+              />
+            ) : null}
+            {successText}
+          </span>
+          <span aria-hidden className={styles.textPlaceholder}>
+            {children}
```
talkor, line 337: Used this text placeholder to keep the success button the same size, same technique that was used in the loading state, it saves us the need to use the resize observer

src/components/Button/Button.module.scss
```diff
@@ -33,10 +32,20 @@ $disabled-on-primary-color-opacity: 0.5;
...
-    }
+  .textPlaceholder {
+    opacity: 0;
+    height: 0;
+    visibility: hidden;
+    white-space: pre;
+  }
+
+  .successContent {
+    position: absolute;
```
orrgottlieb, line 45: This assumes that the text size of the success is shorter than the original content

src/components/Button/Button.tsx
orrgottlieb: I have a comment about it in the CSS file

orrgottlieb (approved):

src/components/Button/Button.module.scss
talkor: Good point, I flexed it instead, thanks!

src/components/Button/Button.tsx
```diff
@@ -339,26 +317,31 @@ const Button: VibeComponent<ButtonProps, unknown> & {
...
+                iconSize={successIconSize}
+                className={cx({
+                  [styles.leftIcon]: !!successText
+                })}
+                ignoreFocusStyle
+              />
+            ) : null}
+            {successText}
+          </span>
+          <span aria-hidden className={styles.textPlaceholder}>
```
orrgottlieb, line 336:
```suggestion
          <span aria-hidden="true" className={styles.textPlaceholder}>
```

PR #1742, opened 2023-11-12
fix: <Dropdown> multi single line use hidden options - for all the sizes

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5495680889

Should contain fix for this, cause it leading to infinite cycle in some cases (with not medium size) https://github.com/mondaycom/monday-ui-react-core/pull/1708

image


talkor:
Already fixed :)

PR #1744, opened 2023-11-13
fix(Button): fix button crash when changing to loading state

talkor:
Revert mondaycom/monday-ui-react-core#1739

https://monday.monday.com/boards/3532714909/pulses/5303882985


src/components/Button/Button.tsx
```diff
@@ -354,7 +371,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
             ignoreFocusStyle
           />
         ) : null}
-        {children}
+        <span>{children}</span>
```
talkor, line 374: Just this fixes the issue of crash when changing to loading state

PR #1745, opened 2023-11-14
Editable text

talkor:
https://monday.monday.com/boards/3532714909/pulses/3950034088

Screenshot 2023-11-14 at 12 17 52


plop/component/index.js
```diff
@@ -58,7 +58,7 @@ module.exports = plop => {
         path: "webpack/published-ts-components.js",
         pattern: /(\/\/ plop_marker:published-components)/g,
         template:
-          '  {{properCase componentName}}: "/src/components/{{properCase componentName}}/{{properCase componentName}}",'
+          '  {{properCase componentName}}: "components/{{properCase componentName}}/{{properCase componentName}}",'
```
talkor, line 61: The path here was wrong

src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -25,143 +18,34 @@ export interface EditableHeadingProps extends VibeComponentProps {
...
-          <Heading
-            {...clickableProps}
-            className={cx(styles.heading, { [styles.disabled]: readOnly })}
-            type={type}
-            weight={weight}
-          >
-            {inputValue}
+        renderTypography={({ value, className, ...clickableProps }) => (
+          <Heading type={type} weight={weight} className={className} {...clickableProps}>
+            {value}
```
talkor, line 43: I was wondering about a couple of ways of composing `EditableText` and `EditableTypography` out of the generic `EditableTypography`, using render props seems most reasonable, this will allow us (when needed) to easily make any typography component editable. Other alternatives might be providing a component as a prop (but this means prop drilling - context will work but it will be less explicit), or passing the typography component as children which will probably require using `cloneElement` to inject additional props but [it's not recommended now](https://react.dev/reference/react/cloneElement)

src/components/EditableText/EditableText.tsx
```diff
@@ -0,0 +1,56 @@
...
+      type = Text.types.TEXT2,
+      weight = Text.weights.NORMAL,
+      "data-testid": dataTestId,
+      id,
+      ...editableTypographyProps
+    },
+    ref
+  ) => {
+    return (
+      <EditableTypography
```
talkor, line 38: It's cool that this composition makes this component super lean and minimal, dealing only with the typography itself

src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -0,0 +1,34 @@
+@import "~monday-ui-style/dist/mixins";
```
talkor, line 1: All `EditableTypography` files are previously the new `EditableHeading` component that became more generic

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -0,0 +1,148 @@
...
+  onChange?: (value: string) => void;
+  /** Disables editing mode - component will be just a <Heading /> */
+  readOnly?: boolean;
+  /** Shown in edit mode when the text value is empty */
+  placeholder?: string;
+  /** ARIA Label */
+  ariaLabel?: string;
+}
+
+interface EditableTypographyProps extends VibeComponentProps, EditableTypographyImplementationProps {
```
talkor, line 24: These props are only require for the composition and should not be exposed from the composed components, this is why the types are separated

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -0,0 +1,148 @@
...
+              onKeyDown={handleKeyDown}
+              onBlur={handleBlur}
+              aria-label={ariaLabel}
+              placeholder={placeholder}
+              style={{ width: inputWidth }}
+              role="input"
+            />
+          </>
+        ) : (
+          renderTypography({ value: inputValue, className: styles.typograpy, ...clickableProps })
```
talkor, line 141: This is the major change

src/components/EditableText/EditableText.tsx
```diff
@@ -0,0 +1,56 @@
...
+        ref={ref}
+        data-testid={dataTestId || getTestId(ComponentDefaultTestId.EDITABLE_TEXT, id)}
+        renderTypography={({ value, className, ...clickableProps }) => (
```
shlomitc, line 41: 1. you get the `value` already from the props, why not simply render it instead of passing it throgh the `EditableTypography`?

src/components/EditableTypography/EditableTypography.tsx
shlomitc: just make sure there are no redundant renders because you call the function each time

src/components/EditableText/EditableText.tsx
talkor: It's not the same value, this value is controlled inside `EditableTypography`

shlomitc (approved):

PR #1746, opened 2023-11-14
docs: Related components rename <Steps>, <LinearProgressBar>, PascalCase

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5506782044


src/storybook/components/related-components/component-description-map.js
```diff
@@ -73,10 +73,10 @@ export const COMBOBOX = "combobox";
 export const AVATAR = "avatar";
 export const AVATAR_GROUP = "avatar-group";
 export const ICONS = "icons";
-export const STEPPER = "stepper";
+export const STEPS = "stepper";
 export const SPINNER = "spinner";
 export const SKELETON = "skeleton";
-export const PROGRESS_BAR = "progress-bar";
+export const LINEAR_PROGRESS_BAR = "progress-bar";
```
talkor, line 79: Why don't we change these values already?
And not related to this PR but why don't we just export these as an object?

talkor (approved):

src/storybook/components/related-components/component-description-map.js
SergeyRoyt: Good catch about the values. Weren't feeling well, while doing it...
Regarding the second question, I actually have no idea... Maybe let's refactor it later

PR #1752, opened 2023-11-15
fix(Button): fix loading behavior when button is type submit in a for…

talkor:
https://monday.monday.com/boards/3532714909/pulses/5514464875


talkor:
Split from #1749

SergeyRoyt:
👍

PR #1758, opened 2023-11-20
Editable typography animation and a11y

talkor:
https://monday.monday.com/boards/3532714909/pulses/5414963372

### DIFF
--- a/src/components/EditableTypography/EditableTypography.tsx
+++ b/src/components/EditableTypography/EditableTypography.tsx
@@ -108,7 +108,7 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> = 
 
     const AnimatedTypography = motion(TypographyComponent);
     const isOverflowing = useIsOverflowing({ ref: typographyRef, ignoreHeightOverflow: true });
-    const shouldAnimate = useMemo(() => isOverflowing && isEditing !== null, [isEditing, isOverflowing]);
+    const shouldAnimate = useMemo(() => isOverflowing && !isEditing, [isEditing, isOverflowing]);
 
     useLayoutEffect(() => {
       if (!typographyRef.current) {


src/components/EditableHeading/EditableHeading.tsx
```diff
@@ -38,11 +38,7 @@ const EditableHeading: VibeComponent<EditableHeadingProps, HTMLElement> & {
       <EditableTypography
         ref={ref}
         data-testid={dataTestId || getTestId(ComponentDefaultTestId.EDITABLE_HEADING, id)}
-        renderTypography={({ value, className, ...clickableProps }) => (
-          <Heading type={type} weight={weight} className={className} {...clickableProps}>
-            {value}
-          </Heading>
-        )}
+        component={Heading}
```
talkor, line 41: Had to change the component's API (internal) because framer doesn't work well with render functions

src/components/EditableText/__stories__/EditableText.stories.mdx
```diff
@@ -69,8 +69,8 @@ Editable Text component is based on the <Link href="/?path=/docs/text-text--over
 
 <Canvas>
   <Story name="Types">
-    <EditableText type={EditableText.types.H1} value="Text1" />
-    <EditableText type={EditableText.types.H2} value="Text2" />
+    <EditableText type={EditableText.types.TEXT1} value="Text1" />
```
talkor, line 72: This was a mistake

src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -15,14 +16,15 @@
...
+    color: var(--primary-text-color);
 
     &:focus,
     &:active {
       outline: none;
     }
   }
 
-  .typograpy {
+  .typography {
```
talkor, line 27: This was a typo (graphy)

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -109,22 +114,27 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
-            <input
+      <LazyMotion features={domAnimation}>
+        <div
+          ref={mergedRef}
+          id={id}
+          aria-label={ariaLabel}
+          data-testid={dataTestId}
+          className={cx(styles.editableTypography, className)}
+        >
+          <HiddenInputPlaceholder
```
talkor, line 130: I had to always render this placeholder and not just in edit mode but also when the typography component is shown, because we use it to calculate overflow, because only if there's overflow the animation is shown, and if it's not rendered we have no way of knowing whether there's overflow
It has `aria-hidden=true` but still wondering

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -109,22 +114,27 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
-      >
-        {isEditing && !readOnly ? (
-          <>
-            <HiddenInputPlaceholder
-              className={cx(styles.input, typographyClassName)}
-              value={inputValue || placeholder}
-              onChange={setInputWidth}
-            />
-            <input
+      <LazyMotion features={domAnimation}>
```
talkor, line 122: Motion is lazy loaded (and it's tree shakable, plus we take only the `m` function and not the `motion` function which is the light version of it, see [this](https://www.framer.com/motion/guide-reduce-bundle-size/#how-to-reduce-the-size-of-the-motion-component) for more info

talkor: If it's not loaded the animation just isn't shown

SergeyRoyt (approved):
Looks very nice! 💪

src/components/EditableTypography/HiddenInputPlaceholder.tsx
```diff
@@ -1,26 +1,33 @@
...
+  ({ className, value, onChange }, ref) => {
+    const valueRef = useRef(null);
+    const mergedRef = useMergeRefs({ refs: [ref, valueRef] });
 
-  useLayoutEffect(() => {
-    const { width } = valueRef.current.getBoundingClientRect();
-    onChange(width);
-  }, [onChange, value]);
+    useLayoutEffect(() => {
+      if (!valueRef.current) {
```
YossiSaadi, line 18: what's the motivation of using `valueRef`/`mergedRef` here instead?

YossiSaadi: What I mean is, can't you use only the `mergedRef` here?

src/components/EditableTypography/EditableTypography.tsx
YossiSaadi: Can you please elaborate how is this helping us understand overflow? I don't get when `onChange` is triggered. aren't you using `motion.input` for editing?

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -40,18 +44,19 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
-      renderTypography
+      component: TypographyComponent
     },
     ref
   ) => {
     const componentRef = useRef(null);
     const mergedRef = useMergeRefs({ refs: [ref, componentRef] });
 
-    const [isEditing, setIsEditing] = useState(false);
+    const [isEditing, setIsEditing] = useState(null);
```
YossiSaadi, line 54: why is `false` doesn't get the job done?

YossiSaadi (commented):
Had a few questions, but it looks great! 👏🏼

src/components/EditableHeading/EditableHeading.tsx
talkor: Yes

package.json
```diff
@@ -109,6 +109,7 @@
     "body-scroll-lock": "^4.0.0-beta.0",
     "browserslist-config-monday": "1.0.6",
     "classnames": "^2.3.2",
+    "framer-motion": "^6.5.1",
```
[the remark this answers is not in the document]
talkor: React 16/18

src/components/EditableTypography/EditableTypography.tsx
talkor: Good catch, useless now, was used in previous attempts, thanks

src/components/EditableTypography/EditableTypography.tsx
talkor: The thing is, we need to adjust the input's size based on the text. There's no way of doing it with just CSS, see [this for more info](https://github.com/mondaycom/monday-ui-react-core/pull/1727#discussion_r1378548622). To get the animation to work properly, I need to figure out if there's an overflow. And to do this, I need to render this `HiddenInputPlaceholder`, because the typography components is unmounts before the overflow calculation is called. I can't figure it out based on the input because it's impossible with that element. Regardless I'm using `motion.input`. I hope that answers the question, open to ideas :)

src/components/EditableTypography/HiddenInputPlaceholder.tsx
talkor: I might be mistaken but there's a bug in the return value of useMergedRefs that causes some typing issues, I checked quickly and it looks like it

package.json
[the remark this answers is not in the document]
YossiSaadi: Once we upgrade Vibe, can we upgrade this? or are we obligated to keep Vibe with a lower version?

src/components/EditableTypography/EditableTypography.tsx
YossiSaadi: Make sure that it doesn't break anything, maybe it was intended
Notice the `useEffect` where you check specifically for nullish, not false

package.json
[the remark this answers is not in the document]
SergeyRoyt: Yeah, we are doing it cause of the React version on the monolith :(

src/components/EditableHeading/__tests__/EditableHeading-interactions.tsx
```diff
@@ -15,10 +15,7 @@ import { resetFocus } from "../../../__tests__/interactions-helper";
 const text = "This heading is an editable heading";
 
 function getComponent(canvas: Canvas) {
-  return getByTestId(canvas, ComponentDefaultTestId.CLICKABLE);
-}
-function getHeading(canvas: Canvas) {
```
talkor, line 20: This was uselss

src/components/EditableText/__tests__/EditableText-interactions.tsx
```diff
@@ -15,10 +15,7 @@ import { resetFocus } from "../../../__tests__/interactions-helper";
 const text = "This text is an editable text";
 
 function getComponent(canvas: Canvas) {
-  return getByTestId(canvas, ComponentDefaultTestId.CLICKABLE);
-}
-function gettext(canvas: Canvas) {
```
talkor, line 20: This was useless

src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -14,20 +16,29 @@
...
     border: 1px solid transparent;
     padding: 0 var(--spacing-xs);
     border-radius: var(--border-radius-small);
 
     &:hover:not(.disabled) {
       border-color: var(--ui-border-color);
     }
+
+    &.hidden {
```
talkor, line 36: This is instead of the HiddenPlaceholder, it had this style

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -52,12 +55,13 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
     const inputRef = useRef(null);
+    const typographyRef = useRef(null);
 
-    function toggleEditMode() {
+    function toggleEditMode(event: React.KeyboardEvent | React.MouseEvent) {
       if (readOnly || isEditing) {
         return;
       }
-
+      event.preventDefault();
```
talkor, line 64: To prevent page scroll when clicking "space" when the element is focused

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -88,14 +92,7 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
 
-    const clickableProps = useClickableProps(
-      {
-        onClick: toggleEditMode,
-        id,
-        role: "button"
-      },
-      mergedRef
-    );
+    const toggleKeyboardEditMode = useKeyboardButtonPressedFunc(toggleEditMode);
```
talkor, line 95: I couldn't use `useClickableProps` because it has a default role and this component has a more complex situation where the role changes, removing the default will be breaking so I just implemented clickable props here

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -109,22 +106,32 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+    const isOverflowing = useIsOverflowing({ ref: typographyRef, ignoreHeightOverflow: true });
+    const shouldAnimate = useMemo(() => isOverflowing && isEditing !== null, [isEditing, isOverflowing]);
+
+    useLayoutEffect(() => {
+      if (!typographyRef.current) {
+        return;
+      }
+      const { width } = typographyRef.current.getBoundingClientRect();
+      setInputWidth(width);
+    }, [inputValue]);
```
talkor: This was called in the HiddenPlaceholder component that is no longer needed, because I calculate the width using the already-rendered typography, it's just hidden when you're in edit mode

talkor:
@YossiSaadi @SergeyRoyt Please take another look, I fixed some accessibility issues, plus I got rid of the HiddenPlaceholder and I now always render the typography even in edit mode (but as a hidden element) to calculate the width
Thanks

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -136,11 +143,20 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
             />
-          </>
-        ) : (
-          renderTypography({ value: inputValue, className: styles.typograpy, ...clickableProps })
-        )}
-      </div>
+          )}
+          <AnimatedTypography
+            ref={typographyRef}
+            aria-hidden={isEditing}
```
talkor, line 149: When it's in edit mode I "hide" it but keeping the width to calculate the width of the input

src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -109,22 +106,32 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
       }
     }, [isEditing]);
 
+    const AnimatedTypography = motion(TypographyComponent);
+    const isOverflowing = useIsOverflowing({ ref: typographyRef, ignoreHeightOverflow: true });
+    const shouldAnimate = useMemo(() => isOverflowing && isEditing !== null, [isEditing, isOverflowing]);
```
YossiSaadi, line 111: Don't forget this null handling (it's never null)

YossiSaadi (approved):
That looks a hell lot cleaner and makes sense that you can take advantage of the rendered text. Nice 👏🏻

src/components/EditableTypography/EditableTypography.tsx
talkor: Oh right thanks! I remembered for a moment and then forgot again :(

PR #1759, opened 2023-11-21
EditableText/EditableHeading fixed and update docs

talkor:
https://monday.monday.com/boards/3532714909/pulses/5548619802

### DIFF
--- a/src/components/EditableHeading/__stories__/EditableHeading.stories.mdx
+++ b/src/components/EditableHeading/__stories__/EditableHeading.stories.mdx
@@ -1,14 +1,14 @@
 import EditableHeading from "../EditableHeading";
 import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
 import { createStoryMetaSettingsDecorator } from "../../../storybook";
-import { createComponentTemplate } from "vibe-storybook-components";
-import { useState } from "react";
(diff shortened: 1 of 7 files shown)


src/components/EditableHeading/__stories__/EditableHeading.stories.mdx
```diff
@@ -1,14 +1,14 @@
...
-import { createComponentTemplate } from "vibe-storybook-components";
-import { useState } from "react";
+import { createComponentTemplate, Tip, Link } from "vibe-storybook-components";
 import { overviewPlaySuite } from "../__tests__/EditableHeading-interactions";
 import {
   TEXT,
   HEADING,
   EDITABLE_TEXT
 } from "../../../storybook/components/related-components/component-description-map";
+import styles from "./EditableHeading.stories.module.scss";
```
shlomitc, line 11: don't we have components in vibe-storybook-components that take care of the layout?
it's just that consumers of the stories can't see what's in these classes unless they inspect it

talkor: I don't think we do, I saw we use mostly classes, @SergeyRoyt no?
I can start making some basic components for that purpose

shlomitc (approved):
let's discuss layout documentation component on a different occasion

PR #1771, opened 2023-11-26
fix: typography vertical overflow tolerance

talkor:
Since we set a line-height which is smaller than the font's normal line-height in the `Heading` component, whenever we check for vertical overflow - it returns true even if it doesn't seem to be overflowing. There's a 2-3px gap.
To avoid it, and instead of setting `line-height: normal` on `Heading`, we:
1. Ignore vertical overflow when it's unneeded, when `maxLines` is 1 (default).
2. Set a tolerance value to "close the gap" as described above. 4px is enough for this case. 

Also we no longer need to pass `ignoreHeightOverflow` prop to the context, I use it for the tolerance now instead.

This is kinda 🤮 but that's the best we can do.

https://monday.monday.com/boards/3532714909/pulses/5413034707

### DIFF
--- a/src/components/Heading/__stories__/Heading.stories.mdx
+++ b/src/components/Heading/__stories__/Heading.stories.mdx
@@ -180,12 +180,9 @@ We support two kinds of ellipsis: single-line ellipsis with a tooltip displayed 
       align={Flex.align.STRETCH}
       style={{ width: "480px" }}
     >
-      <Heading type={Heading.types.H2} ellipsis={false}>
-        Heading without overflow
-      </Heading>
+      <Heading type={Heading.types.H2}>Heading without overflow</Heading>
       <Heading
         type={Heading.types.H2}
-        ellipsis
         /**for testing purposes**/
         data-testid={ONE_LINE_ELLIPSIS_TEST_ID}
         tooltipProps={{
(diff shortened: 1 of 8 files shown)


src/components/Heading/__stories__/Heading.stories.mdx
```diff
@@ -195,10 +192,9 @@ We support two kinds of ellipsis: single-line ellipsis with a tooltip displayed
         Heading with ellipsis and tooltip when hovering
       </Heading>
       <div>
-        <Heading type={Heading.types.H2} ellipsis withoutTooltip maxLines={2}>
-          Heading with two lines overflow heading without tooltip heading with two lines overflow heading without
-          tooltip heading with two lines overflow heading without tooltip heading with two lines overflow heading
-          without tooltip
+        <Heading type={Heading.types.H2} maxLines={2}>
```
talkor, line 195: I don't understand why we set it to `withoutTooltip` before, it has no point of not showing the tooltip as it's the default when there's an overflow

SergeyRoyt:
Seems like a good workaround, but maybe it would be more correct to change the line-height instead? What are we loosing here? It doesn't fit the design requirements then? Maybe let's discuss it with Meirav?

talkor:
> Seems like a good workaround, but maybe it would be more correct to change the line-height instead? What are we loosing here? It doesn't fit the design requirements then? Maybe let's discuss it with Meirav?

Yes, it has a major effect on the design, the difference between the normal line height and what we set is 8px

SergeyRoyt (approved):
If Meirav for sure doesn't approve changing the `line-height`, then I believe it's a good enough solution for now

PR #1773, opened 2023-11-26
feat(Tooltip): add a max width prop

talkor:
https://monday.monday.com/boards/3532714909/pulses/5570948205


src/components/Tooltip/Tooltip.tsx
```diff
@@ -115,6 +115,11 @@ interface TooltipBaseProps extends VibeComponentProps {
    * Overwrites z-index of the tooltip
    */
   zIndex?: number;
+  // TODO: make defualt next major
+  /**
+   * Limit tooltip to 240px
+   */
+  maxWidth?: boolean;
```
talkor, line 122: Couldn't find a better name that will be short enough, open to suggestions

src/components/Tooltip/__stories__/tooltip.stories.js
```diff
@@ -19,6 +19,7 @@ const tooltipTemplate = args => {
         // The modifier's purpose is to prevent the tipseen from being displayed when the user scrolls the story upwards / downwards.
         // Therefore, there is no need to move this prop in your implementations.
         modifiers={modifiers}
+        maxWidth
```
talkor, line 22: Internally showing it in the example so users will take this code as an example

src/components/Tooltip/Tooltip.tsx
shlomitc: setOfficialMaxWidth

src/components/Tooltip/Tooltip.module.scss
```diff
@@ -11,6 +11,10 @@ $tooltip-z-index: 9999999999999999;
   @include font-input();
   max-width: var(--tooltip-max-width, 50vw);
   word-break: break-word;
+
+  &.maxWidth {
```
YossiSaadi, line 15: I would consider maybe a naming change, to make it clearer

src/components/Tooltip/Tooltip.tsx
YossiSaadi: Even `withMaxWidth` or something similar, `limitWidth`

YossiSaadi (approved):
Looks cool, but I would indeed consider a naming change

PR #1775, opened 2023-11-27
infra: disable cron for update monday-ui-style version workflow

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5574910459

### DIFF
--- a/.github/workflows/update-icons.yml
+++ b/.github/workflows/update-icons.yml
@@ -1,7 +1,8 @@
 name: Update monday-ui-style
-on:
-  schedule:
-    - cron: "*/10 * * * *"
+on: workflow_dispatch
+# cron schedule disabled for now
+#  schedule:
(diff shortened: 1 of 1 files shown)


YossiSaadi:
> Just curious, why this change?

@SergeyRoyt @talkor We should implement Renovate for this use cases. 
I agree that the cron is redundant since it creates a lot of open PRs that we only merge the latest of them when really needs to upgrade

YossiSaadi (approved):

talkor:
> > Just curious, why this change?
> 
> @SergeyRoyt @talkor We should implement Renovate for this use cases. I agree that the cron is redundant since it creates a lot of open PRs that we only merge the latest of them when really needs to upgrade

We can also start with dependabot for starters, it's easier to configure I believe

SergeyRoyt:
@shlomitc asked for this
It will be automated once we have monorepo

PR #1776, opened 2023-11-27
fix: <Dropdown> Control - use --primary-color on focus-within instead of default color from react-select

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5570522551

Resolves https://github.com/mondaycom/monday-ui-react-core/issues/1767

### DIFF
--- a/src/components/Dropdown/Dropdown.styles.js
+++ b/src/components/Dropdown/Dropdown.styles.js
@@ -150,6 +150,9 @@ const control =
       minHeight: "30px",
       border: "0 solid transparent",
       borderRadius: getCSSVar("border-radius-small"),
+      ":focus-within": {
+        boxShadow: `0 0 0 1px ${getCSSVar("primary-color")}`
+      },
       ...(!isDisabled && {
         ":hover": {
           borderColor: "transparent",


src/components/Dropdown/Dropdown.styles.js
```diff
@@ -150,6 +150,9 @@ const control =
       minHeight: "30px",
       border: "0 solid transparent",
       borderRadius: getCSSVar("border-radius-small"),
+      ":focus-within": {
+        boxShadow: `0 0 0 1px ${getCSSVar("primary-color")}`
```
YossiSaadi, line 154: I see we are not defining box shadow for control for focused, is the current 'blueish' a react-select default?

YossiSaadi (approved):

src/components/Dropdown/Dropdown.styles.js
YossiSaadi: I'm asking since now the blue will probably be different than the one until now
need to make sure with Meirav that it's cool?

SergeyRoyt: image

YossiSaadi: I'm not sure you had to, was just a suggestion 😅 sorry if it sounded like a blocker by my side 😣

PR #1785, opened 2023-11-28
fix: replace deprecated text-fixed-color tokens with identical color tokens

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5557143080

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "browserslist-config-monday": "1.0.6",
         "classnames": "^2.3.2",
         "lodash-es": "^4.17.21",
-        "monday-ui-style": "0.1.203",
+        "monday-ui-style": "0.1.204",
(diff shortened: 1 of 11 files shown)


YossiSaadi (approved):
I'm thinking about it again. Maybe we shouldn't have removed from monday-ui-style entirely. It's basically breaking change. Wdyt? The chances they were used is low?

SergeyRoyt:
> I'm thinking about it again. Maybe we shouldn't have removed from monday-ui-style entirely. It's basically breaking change. Wdyt? The chances they were used is low?

It's kinda breaking change, yeah. But I removed it everywhere in DaPulse organization and about external usages... it's bassa, but I guess they'll notice and migrate to the new token if they need it. Besides it was public not for that long... So I guess it's not that big deal
We thought that removing it is better than dealing with mess-consequences afterwards

PR #1804, opened 2023-12-07
docs: fix storybook's sidebar overflow

talkor:
https://monday.monday.com/boards/3532714909/pulses/5643763136

### DIFF
--- a/.storybook/manager-head.html
+++ b/.storybook/manager-head.html
@@ -38,6 +38,11 @@
   .os-content button:hover svg {
     color: #5034ff;
   }
+
+  /* https://github.com/storybookjs/storybook/issues/25094 */
+  div[data-radix-scroll-area-viewport] > div {
+    display: block !important;
+  }
 </style>
 
 <script>


.storybook/manager-head.html
```diff
@@ -38,6 +38,11 @@
   .os-content button:hover svg {
     color: #5034ff;
   }
+
+  /* https://github.com/storybookjs/storybook/issues/25094 */
+  div[data-radix-scroll-area-viewport] > div {
+    display: block !important;
+  }
```
YossiSaadi, line 45: Really? that simple? :|

.storybook/manager-head.html
```diff
@@ -38,6 +38,11 @@
   .os-content button:hover svg {
     color: #5034ff;
   }
+
+  /* https://github.com/storybookjs/storybook/issues/25094 */
+  div[data-radix-scroll-area-viewport] > div {
+    display: block !important;
```
talkor, line 44: Unfortunately it's inline style and no selectors, so it's 🤮 
Hoping it's temp until the bug is fixed on their side

YossiSaadi (approved):

.storybook/manager-head.html
talkor: Yep lol

.storybook/manager-head.html
talkor: This is the best we can do

YossiSaadi:
@talkor Post this snippet maybe in the [issue page](https://github.com/storybookjs/storybook/issues/25094)?
So they'll be able to understand what went wrong and fix it quickly

PR #1805, opened 2023-12-07
useMergeRef new hook instead of useMergeRefs with type issues 

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5645813414

Fixes also ref forwarding here https://github.com/mondaycom/monday-ui-react-core/pull/1793, so target this branch


src/hooks/useMergeRef.ts
```diff
@@ -0,0 +1,34 @@
...
+ *   const mergedRef = useMergeRef(ref, internalRef);
+ *   return <div {...props} ref={mergedRef} />;
+ * });
+ */
+
+/**
+ * Returns a single ref callback that merges multiple ref callbacks - replacement for `useMergeRefs` hook
+ * @param refs
+ */
+function useMergeRef<T>(...refs: (RefObject<T> | ForwardedRef<T> | null)[]): RefObject<T> {
```
SergeyRoyt, line 16: Returns `RefObject`, which can be forwarded later on normally and not a weird function as `useMergeRefs` does

src/utils/function-utils.ts
```diff
@@ -48,4 +48,13 @@ export function convertToArray<T>(input: T | Array<T>): Array<T> {
   return isArray(input) ? input : [input];
 }
 
+/**
+ * Checks if React component is wrapped with forwardRef function
+ * @param elementType
+ */
+export const isForwardedRef = (elementType: ElementType): elementType is ForwardRefExoticComponent<unknown> => {
```
orrgottlieb, line 55: dangerous - what happens if React decides to change this Symbol. 

Your users have multiple versions of React, is this true for all React versions

SergeyRoyt: @YossiSaadi you have any ideas how to make it better maybe?

YossiSaadi: @SergeyRoyt Let's consider using [react-is](https://www.npmjs.com/package/react-is).
The numbers talk for themselves, this lib is awesome.
```
import { isForwardRef: reactIsIsForwardRef } from 'react-is';

function isForwardRef(element: React.ReactElement): boolean {
    return reactIsIsForwardRef(element);
}
```
or something similar
I think `React.ReactElement` should do the job from TS side.

YossiSaadi (commented):
Let's consider using react-is, wdyt?

src/hooks/useMergeRef.ts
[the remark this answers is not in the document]
talkor: This haha. The return value's type was wrong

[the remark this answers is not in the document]
YossiSaadi: 💪🏼

src/components/MenuButton/__stories__/MenuButton.stories.js
```diff
@@ -150,21 +151,25 @@ export const Disabled = {
...
+        )}
+      >
+        <Menu id="menu" size={Menu.sizes.MEDIUM}>
+          <MenuItem icon={Sun} onClick={NOOP} iconType={MenuItem.iconType.SVG} title="The sun" />
+          <MenuItem icon={Moon} onClick={NOOP} iconType={MenuItem.iconType.SVG} title="The moon" />
+          <MenuItem icon={Favorite} onClick={NOOP} iconType={MenuItem.iconType.SVG} title="And the stars" />
+        </Menu>
+      </MenuButton>
+    );
+  },
```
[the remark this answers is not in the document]
SergeyRoyt: No usage here - just an example, so they would know how to use if they'd need it
Other usage btw would be - passing `ref` to `MenuButton` directly and wrapping triggerElement with forwardRef function - that would also work

src/utils/function-utils.ts
SergeyRoyt: Removed - using `react-is` instead

package.json
```diff
@@ -115,6 +115,7 @@
     "prop-types": "^15.8.1",
     "react-dates": "21.8.0",
     "react-inlinesvg": "^3.0.1",
+    "react-is": "^16.9.0",
```
YossiSaadi, line 118: Do we need the types also?
https://www.npmjs.com/package/@types/react-is
or there's no errors shown?

src/components/MenuButton/MenuButton.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { forwardRef, useCallback, useLayoutEffect, useMemo, useRef, useState } from "react";
 import cx from "classnames";
 import { camelCase } from "lodash-es";
+import { isForwardRef } from "react-is";
```
YossiSaadi, line 4: I think it's worth doing what you did before
have a util that exports the logic

src/utils/function-utils.ts
YossiSaadi, line 1: ElementType, ForwardRefExoticComponent are still here

src/components/MenuButton/MenuButton.tsx
YossiSaadi: wdyt about creating a new `lib` under `utils` or as a sibling dir of util? and let's migrate slowly all of our imports from libraries like lodash, etc. so we can import from `lib` instead of importing from libraries all over our app.
we can start with your new `isForwardRef` to start pushing it

YossiSaadi: a suggestion:

. lib
|----- react-is.ts (that includes whatever utils we'll need to export)
|----- loadsh.ts (same) (not for this PR)
|----- index.ts (?) (to import from /lib whatever needed, not necessarily from /lib/react-is)

SergeyRoyt: What for? If it's one function which can be imported directly...

package.json
SergeyRoyt: Tried doing that, but there is not @types/react-is with the exact same version `^16.9.0`, so I thought it's ok, if I'll just declare it in `types/packages.d.ts` - check it out

src/components/MenuButton/MenuButton.tsx
YossiSaadi: it's a better practice for cases you want to change a lib and keep the logic, or log stuff every time you want to call some function
it's ok to leave it like that for now, but it's just a better practice

package.json
YossiSaadi: you don't need the same version, you can use the latest that exists for 16.x, it means there were no significant changes that required upgrading versions

SergeyRoyt: Ok, installed `"@types/react-is": "^16.7.5"`👍

src/components/ButtonGroup/ButtonGroup.tsx
```diff
@@ -50,7 +50,7 @@ interface ButtonGroupProps extends VibeComponentProps {
   tooltipMoveBy?: MoveBy;
 }
 
-const ButtonGroup: React.ForwardRefExoticComponent<ButtonGroupProps & React.PropsWithChildren<unknown>> & {
+const ButtonGroup: VibeComponent<ButtonGroupProps, HTMLDivElement> & {
```
YossiSaadi, line 53: Insert `children` to `ButtonGroupProps` or extend the `ButtonGroupProps` interface with `PropsWithChildren`:
```
interface ButtonGroupProps extends VibeComponentProps {
  ...same props as today,
  children?: React.ReactNode
}
```

or better I think:
```
interface ButtonGroupProps extends VibeComponentProps, React.PropsWithChildren {
```

src/components/List/List.tsx
```diff
@@ -69,17 +69,21 @@ const List: VibeComponent<ListProps> & {
...
-    const childrenRefs: React.MutableRefObject<HTMLElement[]> = useRef([]);
 
-    const updateFocusedItem = useCallback((id: string) => {
-      setFocusIndex(getListItemIndexById(childrenRefs, id));
-      componentRef?.current?.setAttribute("aria-activedescendant", id);
-    }, []);
+    const updateFocusedItem = useCallback(
+      (id: string) => {
+        setFocusIndex(getListItemIndexById(childrenRefs, id));
+        mergedRef?.current?.setAttribute("aria-activedescendant", id);
```
YossiSaadi, line 83: it isn't dangerous, right? should be exactly the same

src/components/List/List.tsx
```diff
@@ -69,17 +69,21 @@ const List: VibeComponent<ListProps> & {
...
+      },
+      [mergedRef]
```
YossiSaadi, line 85: componentRef wasn't a dependency, need to make sure no side effect happens (compare two behaviors in the story page?)

src/components/MultiStepIndicator/MultiStepIndicator.tsx
```diff
@@ -135,7 +132,7 @@ const MultiStepIndicator: VibeComponent<MultiStepIndicatorProps> & {
 
     return (
       <ol
-        ref={mergedRef}
+        ref={ref}
```
YossiSaadi, line 135: so no need for local ref here?

src/components/Typography/TypographyHooks.tsx
```diff
@@ -4,11 +4,11 @@ import useIsOverflowing from "../../hooks/useIsOverflowing/useIsOverflowing";
 import { TooltipProps } from "../Tooltip/Tooltip";
 import styles from "./Typography.module.scss";
 
-export function useEllipsisClass(ref: (node: HTMLElement) => void, ellipsis: boolean, maxLines: number) {
+export function useEllipsisClass(ref: MutableRefObject<HTMLElement>, ellipsis: boolean, maxLines: number) {
   let ellipsisClass;
   const overrideRef = (node: HTMLElement) => {
     node?.style.setProperty("--text-clamp-lines", maxLines.toString());
-    ref(node);
+    ref.current = node;
```
YossiSaadi, line 11: Can you please elaborate about the changes here?

src/hooks/useHover/useHover.ts
```diff
@@ -1,9 +1,9 @@
 import { MutableRefObject, useEffect, useRef, useState } from "react";
 
 // Standard hook implementation https://usehooks.com/useHover/
-export default function useHover(): [MutableRefObject<HTMLElement>, boolean] {
+export default function useHover<T extends HTMLElement = HTMLElement>(): [MutableRefObject<T>, boolean] {
```
YossiSaadi, line 4: is there a need in extending HTMLElement if your passing in a type and initializing it to be HTMLElement anyway?
BTW, I LOVE generics so I like it 😆

src/hooks/useIsMouseEnter.ts
```diff
@@ -1,7 +1,8 @@
 import { useCallback, useState, RefObject } from "react";
 import useEventListener from "./useEventListener";
 
-export default function useMergeRefs({ ref }: { ref: RefObject<HTMLElement> }) {
+// TODO can be replaced with useIsMouseOver
```
YossiSaadi, line 4: no difference at all? maybe this hook (without reading it throughly, only by naming) checks for the `enter` event and `over` hook check for if the mouse is over something at any time

types/files.d.ts
YossiSaadi, line 1: Does this change have any affect?

YossiSaadi (commented):
I really like your refactors 😄

src/hooks/useHover/useHover.ts
SergeyRoyt: There is a need actually, cause `T extends HTMLElement` means that whatever class is passed it should extend `HTMLElement`, which allows to use below some `HTMLElement` functions e.g. `addEventListener`
and = `HTLMElement` is just a default value for the the type

src/hooks/useIsMouseEnter.ts
SergeyRoyt: Seems to me like no difference at all, yeah. Didn't replaced it, cause if replacing thought that might be needed to rename corresponding variables as well + also though if it might cause merge conflicts for Yulia

types/files.d.ts
SergeyRoyt: Renamed it from `global.d.ts`, cause at the moment there were also `packages.d.ts`, but currently there is not, so whatevere you prefer - either `global.d.ts` either `files.d.ts`
Content of the file is like this:
```
// SCSS
declare module "*.module.scss" {
  const classes: { [key: string]: string };
  export default classes;
}

// SVG
declare module "*.svg" {
  const content: string;
  export default content;
}
```

YossiSaadi (approved):

PR #1813, opened 2023-12-12
build(Rollup): fail build if circular dependency detected

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5668670716


shlomitc (approved):
👏

PR #1824, opened 2023-12-17
docs(Chips): remove colors which we don't want devs to use from docs

YossiSaadi:
not removing for the meantime from `Chips.colors` as it is a breaking change

https://monday.monday.com/boards/3532714909/pulses/5581617323

### DIFF
--- a/src/components/Chips/__stories__/chips.stories.js
+++ b/src/components/Chips/__stories__/chips.stories.js
@@ -92,30 +92,37 @@ export const Clickable = {
 };
 
 export const ChipsPalette = {
-  render: () => (
-    <Flex
-      style={{
-        width: "100%",
-        height: 300
-      }}
-      align={Flex.align.STRETCH}
-    >
-      {_chunk(Object.keys(Chips.colors), 7).map(colorChunk => {
-        return (
-          <Flex
-            direction={Flex.directions.COLUMN}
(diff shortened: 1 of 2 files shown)


src/components/Chips/__stories__/chips.stories.js
```diff
@@ -92,30 +92,37 @@ export const Clickable = {
...
-            {colorChunk.map(colorName => (
-              <Chips label={colorName} key={colorName} color={Chips.colors[colorName]} readOnly allowTextSelection />
-            ))}
-          </Flex>
-        );
-      })}
-    </Flex>
-  ),
+  render: () => {
+    const excludedColors = [Chips.colors.DARK_INDIGO, Chips.colors.BLACKISH];
```
talkor, line 96: Did we decide to add a deprecate or TODO eventually?

talkor (approved):

src/components/Chips/__stories__/chips.stories.js
YossiSaadi: There's no intent (in the meantime at least) to disable the usage of it.
I'll ask design again to make sure

YossiSaadi: @talkor my bad, added a comment, thanks!

PR #1826, opened 2023-12-17
feat: add a Finish button to Steps

talkor:
- The button appears only for the last step, if `onFinish` prop is provided
- It's opt-in because adding it as default is breaking because the component is controlled, so adding a button that doesn't do anything is bad for UX
- Added TODO to make it the default in next major
https://monday.monday.com/boards/3532714909/pulses/5636406872

### DIFF
--- a/src/components/Steps/Steps.tsx
+++ b/src/components/Steps/Steps.tsx
@@ -27,7 +27,7 @@ export interface StepsProps extends VibeComponentProps {
   backButtonProps?: ButtonProps;
   nextButtonProps?: ButtonProps;
   finishButtonProps?: ButtonProps;
-  onFinish?: (e: React.MouseEvent) => void;
+  onFinish?: (e: React.MouseEvent | React.KeyboardEvent) => void;
 }
 
 const Steps: VibeComponent<StepsProps> & { types?: typeof StepsType } = forwardRef(
--- a/src/components/Steps/Steps.tsx
+++ b/src/components/Steps/Steps.tsx
@@ -26,6 +26,8 @@ export interface StepsProps extends VibeComponentProps {
   areButtonsIconsHidden?: boolean;
   backButtonProps?: ButtonProps;
   nextButtonProps?: ButtonProps;
+  finishButtonProps?: ButtonProps;
+  onFinish?: (e: React.MouseEvent) => void;
 }
 
 const Steps: VibeComponent<StepsProps> & { types?: typeof StepsType } = forwardRef(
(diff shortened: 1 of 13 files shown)


src/components/Steps/StepsCommand.module.scss
```diff
@@ -1,25 +1,11 @@
 .command {
   display: flex;
   align-items: center;
-}
+  gap: 12px;
```
talkor, line 4: By design 🤷🏼‍♂️

src/components/Steps/__stories__/steps.stories.js
```diff
@@ -46,6 +46,7 @@ const NavigableStepsTemplate = args => {
       }}
       {...args}
       onChangeActiveStep={onChangeActiveStep}
+      onFinish={() => {}}
```
talkor, line 49: We want it to appear on the examples because it's not the default now

src/components/Steps/Steps.tsx
```diff
@@ -26,6 +26,8 @@ export interface StepsProps extends VibeComponentProps {
   areButtonsIconsHidden?: boolean;
   backButtonProps?: ButtonProps;
   nextButtonProps?: ButtonProps;
+  finishButtonProps?: ButtonProps;
+  onFinish?: (e: React.MouseEvent) => void;
```
YossiSaadi, line 30: can also be keyboard event, isn't it?

src/components/Tipseen/TipseenWizard.tsx
```diff
@@ -5,15 +5,18 @@ import Button from "../../components/Button/Button";
...
 
+const FINISH_TEXT = "Got it";
+
 interface TipseenWizardProps extends StepsProps {
   title?: string;
   /**
    * Classname for overriding TipseenTitle styles
    */
   titleClassName?: string;
+  onFinish?: (e: React.MouseEvent) => void;
```
YossiSaadi, line 16: same, can't it be a keyboard event in addition?

src/components/Tipseen/TipseenWizard.tsx
```diff
@@ -5,15 +5,18 @@ import Button from "../../components/Button/Button";
 import TipseenBasicContent from "./TipseenBasicContent";
 import styles from "./TipseenWizard.module.scss";
 
+const FINISH_TEXT = "Got it";
```
YossiSaadi, line 8: This is a bit problematic, since if someone doesn't supply a value for it, it won't be translated even if all other strings are translated

YossiSaadi: How does it behave today?

YossiSaadi: Ok I see it also the behavior today. We should think about it, since it means we're not allowing devs to translate it.
Can it be in the `StepsConstants`, same as the `Next`, `Back`

talkor: Yes I talked about it with Meirav, it is problematic but we are letting users override it with any string (using *ButtonProps)

src/components/Steps/Steps.tsx
talkor: Yes, added

src/components/Tipseen/TipseenWizard.tsx
YossiSaadi: ok cool, np

YossiSaadi (approved):
LGTM 👍🏼

PR #1827, opened 2023-12-17
fix: Remove fucos for disabled button

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/5524711549


src/components/Button/Button.tsx
```diff
@@ -251,7 +250,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
         id,
         onFocus,
         onBlur,
-        tabIndex,
+        tabIndex: disabled ? -1 : 0,
```
talkor, line 253: Isn't it a breaking change? We're removing the `tabIndex` actually

talkor (approved):
Great

PR #1833, opened 2023-12-18
fix: trigger scaling for menuButton only while clicking

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/4590288789

### DIFF
--- a/src/components/MenuButton/MenuButton.module.scss
+++ b/src/components/MenuButton/MenuButton.module.scss
@@ -20,8 +20,11 @@
   background-color: var(--primary-background-hover-color);
 }
 
-.active {
+button:active {
   transform: scale(0.95);
+}
+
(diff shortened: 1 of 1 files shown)


src/components/MenuButton/MenuButton.module.scss
```diff
@@ -20,8 +20,11 @@
   background-color: var(--primary-background-hover-color);
 }
 
-.active {
+button:active {
```
talkor, line 23: Wasn't the task to remove the scaling when the button is selected but not active? It looks like it's still scaling down

shaharzil: It is supposed to scale while the user clicks (before he let go of the mouse)

talkor (approved):

PR #1836, opened 2023-12-19
feat(Box): add a scrollable prop

talkor:
https://monday.monday.com/boards/3532714909/pulses/5713244678

### DIFF
--- a/src/components/Box/Box.module.scss
+++ b/src/components/Box/Box.module.scss
@@ -1,13 +1,19 @@
 // Monday ui styles dependencies
 @import "~monday-ui-style/src/common";
+@import "~monday-ui-style/dist/mixins";
 
 // Utilities map
 @import "../../styles/utilities/utilities-settings";
 
-.box{
+.box {
   overflow: hidden;
+
+  &.scrollable {
+    /* TODO: make default in the next major version */
+    overflow: auto;
+    @include scroller;
+  }
 }
 
 // Generating Utility class locally for Box component
 @include utility-class-factory($utilities-modules, $export: "modules");
-
--- a/src/components/Box/Box.tsx
+++ b/src/components/Box/Box.tsx
@@ -71,6 +71,10 @@ interface BoxProps extends VibeComponentProps {
   paddingStart?: ValueOf<PaddingStart>;
   backgroundColor?: ValueOf<BackgroundColor>;
   textColor?: ValueOf<Color>;
+  /**
+   * TODO: make default in next major version
+   */
+  scrollable?: boolean;
 }
 
 const Box: VibeComponent<BoxProps> & {
@@ -121,7 +125,8 @@ const Box: VibeComponent<BoxProps> & {
       paddingBottom,
       paddingStart,
       textColor,
-      backgroundColor
+      backgroundColor,
+      scrollable
     },
     ref
   ) => {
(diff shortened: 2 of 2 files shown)


src/components/Box/Box.module.scss
```diff
@@ -1,13 +1,19 @@
...
+@import "~monday-ui-style/dist/mixins";
 
 // Utilities map
 @import "../../styles/utilities/utilities-settings";
 
-.box{
+.box {
   overflow: hidden;
+
+  &.scrollable {
```
YossiSaadi, line 11: I still think the naming is confusing, it sounds like we're adding scrolling capabilities while the only thing we do is painting the scroller 🥲

src/components/Box/Box.module.scss
```diff
@@ -1,13 +1,19 @@
...
+    /* TODO: make default in the next major version */
+    overflow: auto;
```
YossiSaadi, line 13: Oh we are adding this? Why aren't we letting people to do it on their own?

src/components/Box/Box.tsx
```diff
@@ -121,7 +125,8 @@ const Box: VibeComponent<BoxProps> & {
       paddingBottom,
       paddingStart,
       textColor,
-      backgroundColor
+      backgroundColor,
```
YossiSaadi, line 128: Note to ourselves because I keep forgetting: let's add trailing comma for out prettier/lint 😅

YossiSaadi (approved):

src/components/Box/Box.module.scss
talkor: Yes we do, we have to, see line 9 lol

talkor: I mean, we're indeed adding scrolling capabilities because the component doesn't give such right now

src/components/Box/Box.module.scss
talkor: See line 9 :)

src/components/Box/Box.tsx
talkor: Yes definitely

src/components/Box/Box.module.scss
YossiSaadi: But we're allowing people to add classname and they can add scroll if they need. I would even expect to get overfloe auto out-of-the box as default behavior (?)

talkor: But it's not out-of-the-box, it's only when you use the `scrollable` prop

talkor: Which means you explicitly want to have overflow auto, no? Otherwise the scroller mixin is useless

shlomitc:
Just something to think about, maybe instead of a dedicated scrollable prop, we can consider the “overflow” property and apply it when set

src/components/Box/Box.module.scss
YossiSaadi: Yep, I'm trying to understand if users would expect overflow auto to be the default once height is set thfough class

YossiSaadi: Yeah I see your point, you're right

talkor:
> Just something to think about, maybe instead of a dedicated scrollable prop, we can consider the “overflow” property and apply it when set

I can change it to `overflow` and apply it only when it's auto/scroll

PR #1838, opened 2023-12-20
Split button alignment fix

talkor:
The SplitButton is constructed of two elements: Button, and a Button that is wrapped as a Dialog, which actually make the component wrapper have two elements: Button, and a div (dialog+button). The Button has a line height but the div doesn't. The div inherits line-height (different than 1/normal) which makes the entire components break alignment.
The solution would be to remove `align-items: center` to align it normally without letting external line height to affect. In any way, no need to align it to the center anyway.

https://monday.monday.com/boards/3532714909/pulses/5718731997


shaharzil (approved):

YossiSaadi (approved):
Was not aware of this alignment issue, nice for finding it 💪🏼

PR #1841, opened 2023-12-20
chore(playground-addon): support next components and add an initial code example

shlomitc:
I made some change in the `next` entry that could be risky, so please verify it

image

### DIFF
--- a/src/next.ts
+++ b/src/next.ts
@@ -1,6 +1,5 @@
 export * from "./components/Heading/Heading";
 import Heading from "./components/Heading/Heading";
-
 export * from "./components/EditableHeading/EditableHeading";
 import EditableHeading from "./components/EditableHeading/EditableHeading";
 
--- a/src/next.ts
+++ b/src/next.ts
@@ -1,2 +1,7 @@
-export * as Heading from "./components/Heading/Heading";
-export * as EditableHeading from "./components/EditableHeading/EditableHeading";
+export * from "./components/Heading/Heading";
+import Heading from "./components/Heading/Heading";
+
+export * from "./components/EditableHeading/EditableHeading";
+import EditableHeading from "./components/EditableHeading/EditableHeading";
+
+export { Heading, EditableHeading };
--- a/src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
+++ b/src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
@@ -8,14 +8,20 @@ export default {
 export const Playground = {
   args: {
     code: {
-      jsx: `
-<div style={{display: 'flex', flexDirection: 'column', gap: '12px'}}>
+      jsx: `<div style={{ display: "flex", flexDirection: "column", gap: "12px" }}>
   <div>
     <VibeNext.Heading>Online Playground</VibeNext.Heading>
-    <VibeNext.Heading type="h3" color="secondary">Prototype with actual components.</VibeNext.Heading>
+    <VibeNext.Heading type="h3" color="secondary">
+      Prototype with actual components.
+    </VibeNext.Heading>
   </div>
-  <div>  
-    <Text>Can't see the code editor?<br/>Click on the <VibeIcons.Settings/> button on the left panel and select the <b>"Change Addons Orientation"</b> </Text>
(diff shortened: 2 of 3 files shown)


src/next.ts
```diff
@@ -1,2 +1,7 @@
-export * as Heading from "./components/Heading/Heading";
-export * as EditableHeading from "./components/EditableHeading/EditableHeading";
+export * from "./components/Heading/Heading";
```
YossiSaadi, line 1: I believe it should just be export { default as Heading } from..
Not sure why it exports everything as Heading, are there more stuff being exported from there?

src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
```diff
@@ -5,4 +5,28 @@ export default {
   decorators: [withPlaygroundRenderer]
 };
 
-export const Playground = {};
+export const Playground = {
```
YossiSaadi, line 8: Nice, I liked the initial code 👌🏻. 
Does it still behave well to changes?

src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
```diff
@@ -5,4 +5,28 @@ export default {
...
+  args: {
+    code: {
+      jsx: `<div style={{ display: "flex", flexDirection: "column", gap: "12px" }}>
```
YossiSaadi, line 11: Maybe change it to have className with code in the css section that includes the content

YossiSaadi (commented):
If there are more stuff being exported out of Heading then it would be a risky change, since the import is no longer through import Heading from /next and Heading.something, but straight import {heading} from /next

src/next.ts
shlomitc: types?

src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
shlomitc: It works, but if you move between stories, it will always come back to this snippet first. did it behave differently before?

src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
shlomitc: sure. I wanted to use the `Flex` component but we have some issues there with the static properties

src/next.ts
YossiSaadi: So they are probably unused

src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
YossiSaadi: It should apply this snippet only on Playground's story

src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
YossiSaadi: I'll look into it, not sure why it happens.
I guess it's an issue with react-jsx-parser library. I'll try to check how it renders with react-live's renderer

src/storybook/stand-alone-documentaion/playground/Playground.stories.ts
shlomitc: cool

src/next.ts
shlomitc: why not?

YossiSaadi: I guess no one used them till today. It's only a guess of course.
But if someone did used them then this is a breaking change basically.
But I believe we can make a safe guess no one did (as they would have to use Heading.HeadingProps from what I see?)

YossiSaadi (approved):
Cool! looks awesome 💪🏼

src/next.ts
shlomitc: `dist/esm/next.d.ts` before and after my change looks the same:
```typescript
export { default as Heading } from "../components/Heading/Heading";
export { default as EditableHeading } from "../components/EditableHeading/EditableHeading";
```

So seems like my `export *` in addition to the named export didn't work as I expected as the types exported from this index file

PR #1842, opened 2023-12-21
type: change menu children type

idoyana:
https://monday.monday.com/boards/4317577777/views/98288257/pulses/5698952308


src/components/Menu/Menu/Menu.tsx
```diff
@@ -48,7 +49,7 @@ export interface MenuProps extends VibeComponentProps {
   useDocumentEventListeners?: boolean;
   focusItemIndexOnMount?: number;
   shouldScrollMenu?: boolean;
-  children?: ReactElement | ReactElement[];
+  children?: ReactNode;
```
talkor, line 52: Thanks for contributing! Please use ElementContent, it a type we use for `ReactNode | ReactNode[]`

PR #1854, opened 2023-12-25
docs: <Chips> with custom renderers - story

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5740971200


src/components/Chips/__stories__/chips.stories.js
shaharzil, line 1: Didn't we say that we don't want a story about it?

shaharzil: (Didnt noticed the PR is closed so ignore)

PR #1857, opened 2023-12-27
chore: prettier

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5561513517


YossiSaadi:
That's a pretty heavy PR, lol 😅
Did you run with the Prettier CLI?
I've noticed already a few times that it does not always follow the same rules as running prettier per file (I'm not really sure why, though, never put time into it)

SergeyRoyt:
> That's a pretty heavy PR, lol 😅 Did you run with the Prettier CLI? I've noticed already a few times that it does not always follow the same rules as running prettier per file (I'm not really sure why, though, never put time into it)

Run this command `prettier:fix`
image

We changed prettier rules in the monorepo, so decided to create separate PRs in all the repos to not overload the monorepo process (especially, if we are not gonna merge all the packages at once eventually)

It should follow rules, I've wrote in `prettier.rc` config in all of the files. If you've found any places where it doesn't follow the config, please, show me

tducasse:
note that if you want, we could ignore this commit when git-blaming, so that we don't lose history – and it's even supported in github https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

It's something that can be put in the git config as well for this specific repo, with `git config blame.ignoreRevsFile .git-blame-ignore-revs`

talkor:
> note that if you want, we could ignore this commit when git-blaming, so that we don't lose history – and it's even supported in github https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view
> 
> It's something that can be put in the git config as well for this specific repo, with `git config blame.ignoreRevsFile .git-blame-ignore-revs`

That's awesome @tducasse! Thanks, we'll try that!

talkor:
@SergeyRoyt too many conflicts now in the new structure

PR #1859, opened 2023-12-31
docs: remove unneeded callbacks

talkor:
https://monday.monday.com/boards/3532714909/pulses/5762803720

We have a bug that `backButtonProps` and `nextButtonProps` override the other buttons props, opened a backlog task, but anyway it's not needed because `onChangeActiveStep` can control the state alone


YossiSaadi (approved):
Don't we want it to be reactive?
if not there's content here that can be removed since we'll always only see one step

talkor:
> Don't we want it to be reactive? if not there's content here that can be removed since we'll always only see one step

Is it reactive (line 7)

PR #1861, opened 2024-01-01
feat(Table): add no border variant

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5761975221

### DIFF
--- a/src/components/Table/Table/__stories__/table.stories.js
+++ b/src/components/Table/Table/__stories__/table.stories.js
@@ -265,7 +265,7 @@ export const Sizes = {
   },
   decorators: [
     Story => (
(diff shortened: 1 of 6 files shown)


src/components/Table/Table/__stories__/table.stories.js
```diff
@@ -261,9 +265,135 @@ export const Sizes = {
...
+    );
+  },
+  decorators: [
+    Story => (
+      <Flex direction={Flex.directions.COLUMN} gap={40}>
+        <Story />
+      </Flex>
     )
   ],
   name: "Sizes"
```
shaharzil, line 399: I think the name here needs to be "Borders"

YossiSaadi: Oops
Nice catch, thanks!

shaharzil (approved):

PR #1869, opened 2024-01-04
feat!: <ThemeProvider> add wrapping div (!), rename theme prop to themeConfig

SergeyRoyt:
https://monday.monday.com/boards/3532714909/pulses/5729775904

Targeting this PR https://github.com/mondaycom/monday-ui-react-core/pull/1852


src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -13,15 +13,15 @@ import { withStaticProps } from "../../types";
 
 export interface ThemeProviderProps {
   /**
-   * The theme to apply, consists of name - name of css class that will be added to the children, which should be unique - and the object of colors overrides for each system theme.
+   * The theme config to apply, consists of name - name of css class that will be added to the children, which should be unique - and the object of colors overrides for each system theme.
    */
-  theme?: Theme;
+  themeConfig?: Theme;
```
shlomitc, line 18: I assume that `systemTheme` should be replaced to `theme` in this PR as well, no?

src/components/ThemeProvider/__tests__/themeProvider-tests.jest.js
```diff
@@ -13,7 +13,7 @@ describe("ThemeProvider tests", () => {
       name: THEME_NAME
     };
     const { getByTestId } = render(
-      <ThemeProvider theme={theme} themeClassSpecifier={ADDITIONAL_STRING_SELECTOR}>
+      <ThemeProvider themeConfig={theme} themeClassSpecifier={ADDITIONAL_STRING_SELECTOR}>
```
shlomitc, line 16: worth changing the `theme` variable to `themeConfig` to keep the consistency in the tests as well

src/components/ThemeProvider/ThemeProvider.tsx
```diff
@@ -13,15 +13,15 @@ import { withStaticProps } from "../../types";
 
 export interface ThemeProviderProps {
   /**
-   * The theme to apply, consists of name - name of css class that will be added to the children, which should be unique - and the object of colors overrides for each system theme.
+   * The theme config to apply, consists of name - name of css class that will be added to the children, which should be unique - and the object of colors overrides for each system theme.
```
shlomitc, line 16:
```suggestion
   * The theme config to apply, consists of a "name" - the name of css class that will be added to the children, which should be unique, and the object of colors overrides for each system theme.
```

shlomitc (changes requested):
Just to mention that this is a breaking change and although knowing it's still an alpha component.

1. if we're doing the mix (`theme` => `themeConfig`, `systemTheme` =>  `theme`) worth verifying that the updated `theme` won't cause a major break if the object is passed (considering [this PR](https://github.com/mondaycom/monday-ui-react-core/pull/1852))

src/components/ThemeProvider/ThemeProvider.tsx
shaharzil: @shlomitc In general I agree with you and think we need to go with "themeConfig" and "theme" just like in the SDK API, but I do have a little concern about changing the prop name.
Since we already have users that are using the current theme provider, I think that it will be too confusing for them to understand the change that we did (they won't notice we change the names)
If we will change it from "theme" to "themeConfig" and "systemTheme" it will be pretty clear
But if we will change it from "theme" to "themeConfig" and "theme", they will just assume we added a new prop called "themeConfig" and wont understand why their component is braking.
WDYT?

shlomitc: if we're feeling comfortable with breaking - give it the best names
if we're not feeling comfortable with breaking - give it the closest names

SergeyRoyt: I actually also thought that it'll be confusing if we'll do it like this, so what I did currently and what I'm proposing is this:
- renaming `theme` to `themeConfig`. It'd be corresponding to the SDK prop name + it makes sense cause it's actually configuration of the theme
- new `systemTheme` prop, so it'll be corresponding to `theme` in the SDK. It won't be such confusing as if we'd go with naming it `theme` cause we previously had `theme` and it was smth else. Name make sense cause in our terminology it's system theme precisely

shlomitc: cool. If we want, we can do it in two steps.
1. introduce `themeConfig` and `systemTheme` in the upcoming minor
2. remove `theme` in the minor afterwards

This will allow us to do the gradual opt-in without breaking immediately. I'd also add a `console.error()` if the `theme` prop is still being used by the consumer to let him know he should move to `themeConfig`

shaharzil (approved):

PR #1877, opened 2024-01-08
fix(EditableTypography): change animation to start from -100%

talkor:


talkor:
We decided to drop the animation

PR #1882, opened 2024-01-09
fix(TableCellSkeleton): circle and rectangle loading types are broken (have width 0)

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5819198449

### DIFF
--- a/src/components/Table/TableCellSkeleton/TableCellSkeleton.tsx
+++ b/src/components/Table/TableCellSkeleton/TableCellSkeleton.tsx
@@ -1,4 +1,4 @@
-import React, { FC } from "react";
+import React from "react";
 import cx from "classnames";
 import TableCell from "../TableCell/TableCell";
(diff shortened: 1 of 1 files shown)


src/components/Table/TableCellSkeleton/TableCellSkeleton.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC } from "react";
+import { FC } from "react";
```
talkor, line 1: Isn't it unneeded only starting React 17?

YossiSaadi: @talkor I'm not sure, but the lint yelled at me for it
I think it's from 16
I anyway made a change so it won't yell on it

YossiSaadi:
@talkor btw, it's weird Chromatic didn't catch it

PR #1887, opened 2024-01-10
fix: replace primary-background-hover color in non-hover

talkor:
https://monday.monday.com/boards/3532714909/pulses/5804841496

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "classnames": "^2.3.2",
     "framer-motion": "^6.5.1",
     "lodash-es": "^4.17.21",
-    "monday-ui-style": "0.1.206",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -111,7 +111,7 @@
     "classnames": "^2.3.2",
     "framer-motion": "^6.5.1",
     "lodash-es": "^4.17.21",
-    "monday-ui-style": "0.1.206",
+    "monday-ui-style": "0.1.206-ui-background-color-91321e78d7.0",
```
talkor, line 114: Note to self: replace with actual release

YossiSaadi: you want to merge it with this version?

YossiSaadi (approved):

package.json
YossiSaadi: ok cool

PR #1891, opened 2024-01-11
fix(TableVirtualizedBody): take height of header into consideration

talkor:
The Virtualized list doesn't take the Header into consideration in the height calculation so I've added it. Not sure it's the best solution, would love to hear thoughts.
https://monday.monday.com/boards/3532714909/pulses/5840033094
Before:
Screenshot 2024-01-11 at 19 35 46
After:
Screenshot 2024-01-11 at 19 35 32

### DIFF
--- a/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.module.scss
+++ b/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.module.scss
@@ -1,0 +1,3 @@
+.tableBody {
+  height: calc(100% - var(--table-row-size));
+}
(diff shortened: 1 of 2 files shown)


src/components/Table/TableVirtualizedBody/TableVirtualizedBody.module.scss
```diff
@@ -0,0 +1,3 @@
+.tableBody {
+  height: calc(100% - var(--table-row-size));
```
shlomitc, line 2: Makes sense, just worth adding a comment explaining why

YossiSaadi:
Regarding not sure if it's a good solution - If it works for now than it's cool. If we'll change it that won't be breaking, since the behavior should be the same no matter what's the tech solution will be

src/components/Table/TableVirtualizedBody/TableVirtualizedBody.module.scss
talkor: Added

YossiSaadi (approved):

PR #1900, opened 2024-01-18
chore(chromatic): add a command to run chromatic locally

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5883381568

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,6 +75,7 @@
     "storybook": "storybook dev -p 7008",
     "build-storybook": "npm run build:storybook-exports && storybook build",
     "deploy-storybook": "storybook-to-ghpages",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -75,6 +75,7 @@
     "storybook": "storybook dev -p 7008",
     "build-storybook": "npm run build:storybook-exports && storybook build",
     "deploy-storybook": "storybook-to-ghpages",
+    "chromatic:local": "chromatic -t $CHROMATIC_PROJECT_TOKEN",
```
talkor, line 78: Why just `local`? Wouldn't it work if we use if in the action this way?

YossiSaadi: is it their default cli command?

YossiSaadi: cause using only `chromatic` would override it

talkor: ohhh got it

PR #1904, opened 2024-01-21
fix(MenuItem): split menu items does not open sub-menu when navigating directly between two split menus

YuliaGold:
https://monday.monday.com/boards/2520706822/pulses/5699284655

split menu item doesn't open consistently when hovering between menu items.
isActive is not turning to true even though the mouse is over the correct menu item.
isActive ---> index === activeIndex ---> true
looks like there is missing a render cycle for the menu item in order to update the isActive value so that hover on the split menu item will result in opening the sub menu.

before:
ScreenRecording2024-01-21at11 59 48-ezgif com-video-to-gif-converter

after:
ScreenRecording2024-01-21at11 56 15-ezgif com-video-to-gif-converter

### DIFF
--- a/src/components/Menu/MenuItem/hooks/useMenuItemMouseEvents.ts
+++ b/src/components/Menu/MenuItem/hooks/useMenuItemMouseEvents.ts
@@ -40,16 +40,11 @@ export default function useMenuItemMouseEvents({
     }
   }, [
     setSubMenuIsOpenByIndex,
-    splitMenuItemIconButtonRef,
     index,
-    hasChildren,
-    splitMenuItem,
     isMouseEnterOnIconButton,
     prevIsMouseEnterOnIconButton,
     isActive,
-    resetOpenSubMenuIndex,
-    setActiveItemIndex,
-    isMouseEnter
+    resetOpenSubMenuIndex
   ]);
 
   useLayoutEffect(() => {
@@ -78,11 +73,16 @@ export default function useMenuItemMouseEvents({
 
     if (!isActive && splitMenuItem) {
       setActiveItemIndex(index);
+
+      if (isMouseEnterOnIconButton) {
+        setSubMenuIsOpenByIndex(index, true);
+      }
     }
   }, [
     resetOpenSubMenuIndex,
     prevIsMouseEnter,
     isMouseEnter,
+    isMouseEnterOnIconButton,
     setSubMenuIsOpenByIndex,
     isActive,
     setActiveItemIndex,


src/components/Menu/MenuItem/hooks/useMenuItemMouseEvents.ts
```diff
@@ -40,16 +40,11 @@ export default function useMenuItemMouseEvents({
...
-    splitMenuItemIconButtonRef,
     index,
-    hasChildren,
-    splitMenuItem,
     isMouseEnterOnIconButton,
     prevIsMouseEnterOnIconButton,
     isActive,
-    resetOpenSubMenuIndex,
-    setActiveItemIndex,
-    isMouseEnter
```
YuliaGold, line 52: removed unused dependencies

YossiSaadi:
@YuliaGold I think you are experiencing a different issue.
The behavior in the app does not reproduce on storybook:

YuliaGold:
@YossiSaadi yeah i know that it doesn't reproduce in storybook
when i debug it in monday i can see that even though i'm on the correct menu item, the index doesn't equal to activeIndex

YossiSaadi:
> @YossiSaadi yeah i know that it doesn't reproduce in storybook when i debug it in monday i can see that even though i'm on the correct menu item, the index doesn't equal to activeIndex

@YuliaGold if it works on an isolated environment, it probably means the bug is not with the component
can you share a reproduce?

YuliaGold:
@YossiSaadi 
you can see it in the pr description

YossiSaadi:
> @YossiSaadi you can see it in the pr description

@YuliaGold I mean something I can play with and investigate
It sounds like it's not a Vibe issue, I want to make sure

YuliaGold:
@YossiSaadi ohhh,
i checked it in my local env, and created a link to monday-ui-react-core.
do you want me to show you or just tell you which feature flags to open?

YossiSaadi:
> @YossiSaadi ohhh, i checked it in my local env, and created a link to monday-ui-react-core. do you want me to show you or just tell you which feature flags to open?

if it's already on staging/prod with FF then even better if I don't need the MF.
Slack me with the details.

src/components/Menu/MenuItem/hooks/useMenuItemMouseEvents.ts
talkor, line 1: @YuliaGold I think that just removing line 59 will result in the same effect

YuliaGold: @talkor hey, i tried but it didn't make any affect

YuliaGold: @talkor @YossiSaadi hey do you have any updates?

talkor: Screenshot 2024-01-20 at 22 22 06
@YuliaGold the issue isn't reproduced for me

talkor: @YuliaGold the issue doesn't reproduce for me

https://github.com/mondaycom/monday-ui-react-core/assets/18269880/76352f43-b22e-42a0-813a-c79e7987a219

talkor:
Verified in React 18.2, @YuliaGold please try this [prerelease](https://github.com/mondaycom/monday-ui-react-core/pull/1912#issuecomment-1910814260), should resolve this issue anyway

talkor (approved):

PR #1909, opened 2024-01-21
fix(TableVirtualized): support row height

talkor:
https://monday.monday.com/boards/3532714909/pulses/5866572168


src/components/Table/Table/TableConsts.ts
```diff
@@ -5,4 +5,9 @@ export enum RowSizes {
   LARGE = "large"
 }
 
-export type RowSize = typeof RowSizes | number;
+export enum RowHeights {
+  medium = 40,
+  large = 48
+}
```
YossiSaadi, line 11: I know we came up with this quickly, we'll need to think of some nicer solution other than this 😆

YossiSaadi: Adding to our tech debts 🥲

YossiSaadi (approved):
Awesome solution!!!

src/components/Table/Table/TableConsts.ts
talkor: Yeah agree! Let me think about something real quick

src/components/Table/Table/TableConsts.ts
```diff
@@ -5,4 +5,9 @@ export enum RowSizes {
   LARGE = "large"
 }
 
-export type RowSize = typeof RowSizes | number;
+export const RowHeights = {
```
talkor, line 8: @YossiSaadi WDYT?

YossiSaadi: Great!

src/components/Table/Table/TableConsts.ts
```diff
@@ -5,4 +5,9 @@ export enum RowSizes {
...
+  [RowSizes.MEDIUM]: 40,
+  [RowSizes.LARGE]: 48
+};
+
+export type RowSize = typeof RowSizes;
```
YossiSaadi, line 13: is this really needed? we don't support number anyway and I guess it was here at the first place because of that, otherwise we can just use typeof RowSizes on where RowSize is used

we can add support for number later if needed

talkor: Good catch, removing

talkor: Turns out it's not even used

src/components/Table/Table/TableConsts.ts
```diff
@@ -9,5 +9,3 @@ export const RowHeights = {
   [RowSizes.MEDIUM]: 40,
   [RowSizes.LARGE]: 48
 };
-
-export type RowSize = typeof RowSizes;
```
YossiSaadi: oh it's not even used once? lol 😄

talkor: Yep lol

PR #1911, opened 2024-01-21
fix(MenuButton): fix dialog close behavior

talkor:
https://monday.monday.com/boards/3532714909/pulses/5859293525


src/components/MenuButton/MenuButton.tsx
```diff
@@ -179,8 +179,8 @@ const MenuButton: VibeComponent<MenuButtonProps> & {
 
     const onMenuDidClose = useCallback(
```
talkor, line 180: This event is called when you click on a MenuItem, but not on a SubMenu or SplitMenu Item, which is exactly what we want. Not sure why it reacted only to keyboard event though

YossiSaadi (approved):
amazing bro, great catch
well done 👏🏼

PR #1914, opened 2024-01-23
fix(MenuItem): tooltip should on all item and not only on title

YossiSaadi:
this is a temp hack until the next major, it adds a hidden title element, not visible to screen readers and keyboard navigation, but it catches hover for tooltip width

https://monday.monday.com/boards/3532714909/pulses/5910955456

### DIFF
--- a/src/components/Menu/Menu/Menu.module.scss
+++ b/src/components/Menu/Menu/Menu.module.scss
@@ -3,6 +3,7 @@
 .menu {
   margin: unset;
   padding: unset;
+  position: relative;
 }
 
 .menu:focus {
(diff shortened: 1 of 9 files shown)


src/components/Menu/Menu/Menu.module.scss
```diff
@@ -3,6 +3,7 @@
 .menu {
   margin: unset;
   padding: unset;
+  position: relative;
```
talkor, line 6: It should be on the MenuItem, no?

talkor: Actually I see that it messes it up, maybe because there's some positioning already in it

PR #1922, opened 2024-02-04
fix(MenuButton): tooltip should only appear on trigger element hover, not on dialog hover

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/3539968129

### DIFF
--- a/src/components/MenuButton/MenuButton.tsx
+++ b/src/components/MenuButton/MenuButton.tsx
@@ -321,7 +321,7 @@ const MenuButton: VibeComponent<MenuButtonProps> & {
             ref: isForwardRef(TriggerElement) ? mergedRef : undefined
           };
 
-    const TriggerElementNode = (
+    const triggerElementNode = (
       <TriggerElement
         id={id}
         data-testid={dataTestId || getTestId(ComponentDefaultTestId.MENU_BUTTON, id)}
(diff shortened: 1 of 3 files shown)


src/components/MenuButton/MenuButton.tsx
```diff
@@ -326,49 +379,14 @@ const MenuButton: VibeComponent<MenuButtonProps> & {
...
-          </TriggerElement>
-        </Dialog>
+        {tooltipChildren}
       </Tooltip>
     );
+
+    if (showTooltipOnlyOnTriggerElement) {
+      return DialogNode(TooltipNode(TriggerElementNode));
+    }
+    return TooltipNode(DialogNode(TriggerElementNode));
```
YossiSaadi, line 389: I have to render DialogNode and TooltipNode as functions and not JSX because if they are rendered as JSX (which means they are not functions, but a functional component) Tooltip component (through Dialog's Refable) adds a span wrapper to it (which is the reason for this workaround at first place)

This should be the default at next major, even at cost of adding span by default

src/components/MenuButton/MenuButton.tsx
```diff
@@ -316,7 +321,55 @@ const MenuButton: VibeComponent<MenuButtonProps> & {
...
+        aria-disabled={disabled}
+        {...triggerElementProps}
+      >
+        {componentPosition === MenuButton.componentPositions.START && icon}
+        {text && <span className={styles.innerText}>{text}</span>}
+        {componentPosition === MenuButton.componentPositions.END && icon}
+      </TriggerElement>
+    );
+
+    const DialogNode = (dialogChildren: React.ReactElement) => (
```
talkor, line 348: Shouldn't it be lowercase? (DialogNode) becasue it's not a component

talkor (approved):

PR #1931, opened 2024-02-06
feat(EditableTypography): send onEditModeChange when entering edit mode

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6007451118


talkor (approved):
Toda raba!

PR #1933, opened 2024-02-07
fix(TextField): avoid failing to cleanup on unmount, avoid inputRef is undefined

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6016944708


talkor (approved):
Was GPT wrong eventually?

YossiSaadi:
> Was GPT wrong eventually?

🙃 I'm not taking a risk, did this refactor instead 😅

PR #1935, opened 2024-02-08
feat(Table): new sort behavior

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/5898339907


YossiSaadi:
Not ready yet - design wants extra UX functionalities

packages/core/src/components/Button/__tests__/button.jest.js
```diff
@@ -99,17 +96,19 @@ describe("<Buttoon />", () => {
...
     fireEvent.blur(button);
     expect(onBlur.mock.calls.length).toEqual(1);
   });
 
   it("should call do blur on mouseup", () => {
     const onBlur = jest.fn();
-    cleanup();
-    const { getByText } = renderComponent({ onBlur, blurOnMouseUp: false });
+    const { getByText } = render(
+      <Button onBlur={onBlur} blurOnMouseUp={false}>
```
talkor, line 108: I didn't even notice that we have this prop

packages/core/src/components/Button/__tests__/button.jest.js
```diff
@@ -118,43 +117,56 @@ describe("<Buttoon />", () => {
...
   });
 
   describe("mouse down", () => {
+    const onClick = jest.fn();
     it("should call the click callback when clicked", () => {
-      const { getByText } = buttonComponent;
+      const { getByText } = render(<Button onClick={onClick}>{text}</Button>);
       fireEvent.mouseDown(getByText(text));
-      expect(clickActionStub.mock.calls.length).toEqual(0);
+      expect(onClick.mock.calls.length).toEqual(0);
```
talkor, line 131:
```suggestion
      expect(onClick).not.toBeCalled()
```

packages/core/src/components/Table/TableHeaderCell/TableHeaderCell.tsx
```diff
@@ -37,14 +38,24 @@ const TableHeaderCell: VibeComponent<ITableHeaderCellProps, HTMLDivElement> = fo
...
+        className={cx(styles.tableHeaderCell, { [styles.sortActive]: isSortActive }, className)}
         data-testid={dataTestId || getTestId(ComponentDefaultTestId.TABLE_HEADER_CELL, id)}
         role="columnheader"
-        aria-sort={getAriaSort(sortState)}
+        onMouseOver={() => setIsHovered(true)}
+        onMouseLeave={() => setIsHovered(false)}
+        onFocus={() => setIsHovered(true)}
+        onBlur={() => setIsHovered(false)}
+        aria-sort={onSortClicked ? ariaSort : undefined}
+        tabIndex={onSortClicked ? 0 : undefined}
```
talkor, line 58: null instead?

talkor (approved):
Amazing!

packages/core/src/components/Table/TableHeaderCell/TableHeaderCell.tsx
YossiSaadi: on purpose, since I do not want the prop to be defined, it is the same as we're not passing a prop at all
for `aria-sort` for example, I do not want to show aria-sort in the DOM since a screen-reader can think it is sortable
for `tabIndex`, null is not a correct value, and I want to use the default

PR #1939, opened 2024-02-08
fix(MenuItem): change condition of ariaLabel

talkor:
https://monday.monday.com/boards/3532714909/pulses/6027956948


YossiSaadi (approved):
Great!!!

PR #1952, opened 2024-02-13
ci(prerelease): match to new monorepo

YossiSaadi:


packages/core/src/components/Button/Button.tsx
```diff
@@ -206,7 +206,6 @@ const Button: VibeComponent<ButtonProps, unknown> & {
         getStyle(styles, camelCase("kind-" + kind)),
         getStyle(styles, camelCase("color-" + calculatedColor)),
         {
-          [styles.loading]: loading,
```
talkor, line 209: @YossiSaadi unused?

YossiSaadi: this PR is dead, sorry 🥲 
and yes, it was removed as part of the Table New Sort behavior (#1935)

YossiSaadi:
#1967

PR #1962, opened 2024-02-14
ci(release): add name for matrix step

YossiSaadi:


YossiSaadi:
this change is part of #1966

PR #1967, opened 2024-02-14
ci(prerelease): run only on push (except master), set tag to always be `prerelease`, comment on PR with published versions

YossiSaadi:
This PR fixes:
- the git-tag assigned to the prerelease, which before included the branch name but should be only `prerelease` to not create too many prerelease tags on npmjs list.
- a comment was not posted on a PR
- the comment that was supposed to be posted did not include correct version
- the build step did not include --include-dependencies tag, so it did not built monday-ui-style before building monday-ui-react-core


talkor:
@sahariko can you review this pls? 🙏🏼

YossiSaadi:
A note about all the EOF tricks, it is the recommended way by GitHub to store multilne strings in env and output

.github/workflows/prerelease.yml
```diff
@@ -1,69 +1,71 @@
...
       - name: Build all affected packages
-        run: yarn lerna run build --since=origin/master
-      - name: Generate prerelease prefix
-        run: |
-          normalized_branch_name=$(git rev-parse --abbrev-ref HEAD | tr -dc '[:alnum:]\n\r' | tr '[:upper:]' '[:lower:]')
-          echo "BRANCH=$normalized_branch_name" >> $GITHUB_ENV
+        run: yarn lerna run build --since=origin/master --include-dependencies
+      - name: Run Tests
+        shell: bash
+        run: yarn lerna run test --since=origin/master
```
sahariko, line 29: Why?

.github/workflows/prerelease.yml
```diff
@@ -1,69 +1,71 @@
...
       - run: yarn config set registry https://registry.npmjs.org/
       - name: Setup .npmrc for publish
         run: npm set "//registry.npmjs.org/:_authToken=$NODE_AUTH_TOKEN"
         env:
-          NODE_AUTH_TOKEN: ${{secrets.npm_token}}
-      - run: npm whoami
+          NODE_AUTH_TOKEN: ${{ secrets.npm_token }}
       - name: Publish new versions
-        run: yarn lerna publish from-package --dist-tag $BRANCH -y
+        run: yarn lerna publish from-package --dist-tag prerelease -y
```
sahariko, line 50: This means that if several people try to prerelease at the same time they will override each other. This can be very very unexpected and frustrating, not sure why it's needed.

.github/workflows/prerelease.yml
```diff
@@ -1,69 +1,71 @@
...
       - name: Get current PR id
-        uses: 8BitJonny/gh-get-current-pr@2.2.0
+        uses: 8BitJonny/gh-get-current-pr@3.0.0
         id: PR
-        if: success()
-      - name: Create comment with prerelease version details
-        if: success()
-        uses: peter-evans/create-or-update-comment@v1
+      - name: Prepare PR comment
```
sahariko, line 54: Why all the changes to these steps? It feels like we're not gaining a lot, but it makes them much much more complicated to understand, debug and maintain.

.github/workflows/prerelease.yml
talkor: We wanted to avoid a case where we have lots of random tags along with "latest", "version1" etc. We thought it might be confusing for consumers to know which version to take, it might be chaotic.
Maybe we should instead prefix the tag with `prerelease-` or something? WDYT?
Screenshot 2024-02-18 at 13 35 52

sahariko: Sure, sounds good!
Generally speaking, I don't think people really bother with tags - they just install and get the default `latest`.

.github/workflows/prerelease.yml
YossiSaadi: They used to wait for tests to finish, we can pass on it, but just to be on safe side..

.github/workflows/prerelease.yml
YossiSaadi: To keep same functionality. we used to say which version is published, currently it points to the wrong versioning (only branch name without version and prerelease patch count), also now it says which version was published.

YossiSaadi: Which packages*

.github/workflows/prerelease.yml
sahariko: I think we can forego tests altogether 🤷‍♂️

.github/workflows/prerelease.yml
sahariko: Got ya, although it's important to state that it points to the correct tag, you don't have to have the full version.

YossiSaadi: Yep, but in case we'll have 2 prereleases (though it's rare at Vibe, we only use it at certain cases) or for future reference

YossiSaadi: I wanna make sure that it's understood I'm not against the current implementation, though 👀 just wanted it to have the same as it had till today

.github/workflows/prerelease.yml
YossiSaadi: @talkor mind removing this step plz? 😅🙏🏻

talkor: Done

PR #1973, opened 2024-02-18
docs: tsify box

talkor:
Experimenting with converting stories files to TS. One step ahead being fully TS-compatible


packages/core/src/components/Box/__stories__/Box.stories.tsx
```diff
@@ -154,15 +155,6 @@ export const Shadow = {
 export const Margin = {
   render: () => (
     <div className={styles.boxWrapper}>
-      <div className={styles.boxMarginWrapper}>
-        <Box
-          backgroundColor={Box.backgroundColors.PRIMARY_BACKGROUND_COLOR}
-          border={Box.borders.DEFAULT}
-          margin={Box.margins.NONE}
```
talkor, line 161: No such variant

shaharzil (approved):

Issue #1766, opened 2023-11-24
Button doesn't dispatch onClick event on click on the edges of it 


Andrew-Zakharov:
For some reason when click on the button's edges it doesn't dispatch onClick although click animation is being played 

https://github.com/mondaycom/monday-ui-react-core/assets/108681710/4f7c18ca-2090-4eed-8ddb-2528bd0a9668

Andrew-Zakharov:
I see, thank you for the explanation 👍

Ma1kovich:
@talkor are there any plans on fixing it directly in the package?

RitikBora:
Hey @talkor  
The diffrence between onMouseDown and onClick is marginal. Mouse down event is fired when the button is pressed and onClick when a full click is registered. Basically after the click event

Is it ok if we dispach onClick on MouseDown as well? Will this hamper the workflow in any manner?

talkor:
> Hey @talkor The diffrence between onMouseDown and onClick is marginal. Mouse down event is fired when the button is pressed and onClick when a full click is registered. Basically after the click event
> 
> Is it ok if we dispach onClick on MouseDown as well? Will this hamper the workflow in any manner?

Hey @RitikBora! Yes that's definitely ok, this is also the [workaround suggested](https://github.com/mondaycom/monday-ui-react-core/issues/1766#issuecomment-1827778358)

PR #1978, opened 2024-02-21
fix: Button does not dispatch onClick event on edge click

RitikBora:
Hey @talkor ,

This PR is raised in regards to the issue pointed out in #1766  .

Based on our discussion, there is a very slight diffrence in the onClick and onMouseDown events . Hence, I have fired an onClick event whenever a mouseDownEvent occurs. It takes care of the button edge clicks as well as multiple clicks as everytime a mouseDown event is surely registered

Thanks,
Ritik

### DIFF
--- a/packages/core/src/components/Button/Button.tsx
+++ b/packages/core/src/components/Button/Button.tsx
@@ -192,6 +192,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
 
         if (onMouseDown) {
           onMouseDown(event);
+          onClick(event);
         }
       },
       [onMouseDown, disabled, loading, success]


talkor (commented):
@RitikBora I appreciate it, but I didn't mean it's a change we should make in the component but a workaround to avoid the bug where button clicks on the edges do no trigger `onClick` when the animation is on. We want to allow both `onClick` and `onMouseDown` for the consumer of the component to decide which to use according to their needs.

RitikBora:
Hey @talkor ,

Oh Okay. The change i made wouldn't not hamper the onMouseDown functionality at all. It just fires the onClick event as well.

I'm happy to pullback my request and look into some better solution than a walkaround :) . Let me know if you are ok with that..

Cheers.

talkor:
> Hey @talkor ,
> 
> Oh Okay. The change i made wouldn't not hamper the onMouseDown functionality at all. It just fires the onClick event as well.
> 
> I'm happy to pullback my request and look into some better solution than a walkaround :) . Let me know if you are ok with that..
> 
> Cheers.

Since it's a breaking change (users who called both `onClick` and `onMouseDown` will have `onClick` called twice for example) we wouldn't want to push this change at the moment, especially if it's meant to fix a different minor bug. I suggest to focus on trying to resolve the original bug without affecting the API for now. If you have an idea we'd be more than happy. Thanks!

RitikBora:
Hey @talkor ,

Understood,

Thanks!!

PR #1982, opened 2024-02-25
docs: revamp icon catalog

talkor:
https://monday.monday.com/boards/3532714909/pulses/6115102363


packages/core/src/components/Icon/__stories__/Icon.stories.module.scss
```diff
@@ -0,0 +1,17 @@
+.wrapper {
+  width: 100%;
+
+  .grid {
```
shaharzil, line 4: Loved the usage of grid here!

shaharzil (approved):

PR #1983, opened 2024-02-25
ci: always build before building storybook

talkor:
We need to always build it as it's a dependency of storybook as well, no need to build core though

### DIFF
--- a/.github/workflows/publish-storybook.yml
+++ b/.github/workflows/publish-storybook.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Build & deploy Storybooks to GitHub Pages
         run: |
           yarn
-          yarn lerna run build --since=origin/master --include-dependencies
+          yarn lerna run build --scope monday-ui-style
           yarn lerna run --scope monday-ui-react-core build-storybook
           yarn lerna run --scope monday-ui-style build-storybook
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-storybook.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Build & deploy Storybooks to GitHub Pages
         run: |
           yarn
-          yarn lerna run build --since=origin/master --include-dependencies
+          yarn lerna run build --scope monday-ui-style
```
YossiSaadi, line 23: Don't you always need to build core? And as a result to build style when it's with include deps flag?

Btw, just a thought, I wonder how Chromatic hosts (with the CNAME thing that we wanted to do) Storybooks without building, etc. 🤔 We need to further check how it works and if it worth it

YossiSaadi (approved):

.github/workflows/publish-storybook.yml
talkor: Hmm yeah I think you're right, but we'll have to remove the --since flag anyway because the publishing can be done even when there are no changes in either packages. I'll fix

shaharzil (approved):

PR #1989, opened 2024-03-03
feat(Flex): new align's value - baseline

sosdan73:
Fixes #1969


talkor (approved):
Thank you @sosdan73!! That's really great

Issue #1988, opened 2024-03-01
[Bug]: Tipseen nests a div inside of a p tag


ahlivehr:
### Vibe version

2.93.0

### Describe the bug

Getting the following warning in the Chrome console when using a Tipseen. 
Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.

When inspecting the DOM in the [Storybook basic example,](https://style.monday.com/?path=/docs/popover-tipseen--docs) you can see it does indeed nest a div inside a p tag (see screenshot). Though in Storybook, I don't see the same warning in the Chrome console.

### Expected behavior

Tipseen doesn't nest a div inside of a p tag.

### Steps to reproduce

1. Add a Tipseen to a react project (I'm using vite) and run the project
```
<Tipseen
  content={<TipseenContent title="Title">Popover message will appear here loremipsum dolor samet…</TipseenContent>}
  modifiers={[
    {
      name: 'preventOverflow',
      options: {
        mainAxis: false
      }
    },
    {
      name: 'flip',
      options: {
        fallbackPlacements: []
      }
    }
  ]}
  position="right"
>
  <div className="monday-storybook-tipseen_container" />
</Tipseen>
```
3. Open the Chrome console and you can see the warning: `Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>.
`

### System Info

```shell
System:
    OS: macOS 14.3.1
    CPU: (14) arm64 Apple M3 Max
    Memory: 141.75 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.6.1 - /opt/homebrew/bin/node
    Yarn: 1.22.21 - /usr/local/bin/yarn
    npm: 10.2.4 - /opt/homebrew/bin/npm
  Browsers:
    Chrome: 122.0.6261.94
    Safari: 17.3.1
```

### Additional context, Screenshots

![image](https://github.com/mondaycom/monday-ui-react-core/assets/5676562/49774bf7-e6cf-468d-9561-27e0c45f1036)

HermanBide:
has this been fixed ?

ahlivehr:
@HermanBide I don't think it has. I just filed this bug yesterday.

HermanBide:
can i contribute to fixing this bug ?

shaharzil:
Hi @ahlivehr and thank you for bringing this bug to our attention :)

[...]

PR #1991, opened 2024-03-04
fix: fixed bug #1988 Tipseen component 

HermanBide:
#### Basic

- [ ] PR has description.
- [ ] Component is written in TypeScript.

#### Style

- [ ] Styles are added to `NewComponent.modules.scss` file inside the `NewComponent` folder.
- [ ] Styles are defined using [monday-ui-style](https://github.com/mondaycom/monday-ui-react-core/tree/monorepo-style/packages/monday-ui-style) tokens
- [ ] Displays correctly in all the themes

#### Storybook

- [ ] Stories were added to `/src/components/NewComponent/__stories__/NewComponent.mdx` file.
- [ ] Stories implemented using [vibe-storybook-components](https://github.com/mondaycom/vibe-storybook-components) components and tokens.

#### Tests

#### Accessibility

- [ ] Component is accessible.
- [ ] Component is compatible with [Vibe Accessibility Guidelines](https://style.monday.com/?path=/docs/foundations-accessibility--docs)

////////////////////////////////////////////////////////////////////////////////////////////////
I made changes to the code by removing the div element inside Tipseen element  since a 'p' element could not nest a 'div' and instead placed the Tipseen  inside the div in component "tipseen.stories.js" which stopped the error.


talkor (commented):
Hi @HermanBide! Thank you for your contribution and taking the time to resolve this! Please notice that the project is now a monorepo, so to run it locally you only need to `yarn install` and then `yarn storybook` in the root of the project. We will update the contributions readme really soon. No need to change any dependency or package.json.
Also please notice that the changes you made are in the storybook story page and not in the actual component, which means that the bug won't be fixed for the library users.

HermanBide:
@talkor thank you for getting back to me i can fix that error if you can let me know in what the actual component folder/file i should make the changes in.

talkor:
> @talkor thank you for getting back to me i can fix that error if you can let me know in what the actual component folder/file i should make the changes in.

Sure @HermanBide ! First I suggest walking through the [contribution guide](https://github.com/mondaycom/monday-ui-react-core/blob/master/packages/core/CONTRIBUTING.md). Here are some pointers: Is the error happens only because of the story? You removed the trigger `<div>` of the Tipseen so nothing can trigger it. If it's a bug in the component and not in the story, there are no actual changes in the Tipseen.tsx file other than a comment. Also let's avoid commented-out code. Thanks!

PR #1994, opened 2024-03-05
feat(Button): when aria-hidden is true, button should behave as disabled (solves a11y issue)

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6194425851

reference:
https://dequeuniversity.com/rules/axe/4.8/aria-hidden-focus

### DIFF
--- a/packages/core/src/components/Button/Button.tsx
+++ b/packages/core/src/components/Button/Button.tsx
@@ -155,6 +155,8 @@ const Button: VibeComponent<ButtonProps, unknown> & {
     const { loading } = useButtonLoading({ isLoading });
     const overrideDataTestId = backwardCompatibilityForProperties([dataTestId, backwardCompatabilityDataTestId]);
 
+    const disabledButton = disabled || ariaHidden;
+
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Button/Button.tsx
```diff
@@ -220,7 +222,7 @@ const Button: VibeComponent<ButtonProps, unknown> & {
           [styles.leftFlat]: leftFlat,
           [styles.preventClickAnimation]: preventClickAnimation,
           [styles.noSidePadding]: noSidePadding,
-          [styles.disabled]: disabled,
+          [styles.disabled]: disabledButton,
```
talkor, line 225: Is it ok that it's disabled style if `aria-hidden` is provided? Not a side effect?

YossiSaadi: hmm I get what you mean. yep. let's leave it with the `disabled` only, unless something weird comes up

talkor (approved):

PR #1997, opened 2024-03-05
chore: Add storybook component squashable

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6196411117


talkor (approved):

talkor:
@shaharzil we need to also add the package to the `publish-storybook` workflow

PR #2000, opened 2024-03-07
ci: add reusable workflow for upload and download build artifacts in actions

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6212333840

TL;DR:
- Call the reusable workflow of `build-and-upload` as first job in your workflow
- Call the reusable action of `download-builds` instead the previous `yarn lerna run build ...` step

@shaharzil Add your `storybook-blocks` to the `.github/workflows/build-and-upload.yml` file under `path`.
An example usage to call the build and upload is as following:
```
jobs:
  build:
    name: Build
    uses: ./.github/workflows/build-and-upload.yml
    secrets:
      npm_token: ${{ secrets.npm_token }}
  ...rest of jobs (ci, build-storybook, etc.)
```

After this you'll have an artifact that includes all the `packages/*/dist` for the current workflow run id

And you need to state that the rest of the jobs cannot run in parallel, and they `needs` the build step to succeed. an example:
```
  ci:
    needs: build
    runs-on: ubuntu-latest
    ...rest of the ci job
```

```
  build-storybook:
    needs: [build, ci]
    runs-on: ubuntu-latest
    ...rest of the ci job
```

And instead of building inside jobs, you can now call the workflows.
Old:
```
steps:
  - uses: actions/checkout@v4
    with:
      fetch-depth: 0
  ...rest of steps
  - name: Run Build
    shell: bash
    run: yarn lerna run build --since=origin/master --include-dependencies
  ...rest of steps
```
New:
```
steps:
  - uses: actions/checkout@v4
    with:
      fetch-depth: 0
  ...rest of steps
  - uses: ./.github/actions/download-builds
  ...rest of steps
```


.github/workflows/build-and-upload.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Run Setup
+        uses: ./.github/actions/setup
+        with:
+          npm_token: ${{ secrets.npm_token }}
+      - name: Build
+        shell: bash
+        run: yarn lerna run build --since=origin/master --include-dependencies
```
shaharzil, line 22: this is relevant only for branches that aren't master, maybe we can emphasis it somehow so no one will use it for master?

.github/actions/download-builds/action.yml
shaharzil, line 1: you wrote in the PR's description: "Call the reusable action of download-builds instead the previous yarn lerna run build ... step"
Are we cool with that happening also for the lint steps?

.github/actions/download-builds/action.yml
```diff
@@ -0,0 +1,11 @@
+name: Download and Extract Build Artifacts
+description: Download build artifacts for all the packages and extract them to the correct location
+
+runs:
+  using: "composite"
```
shaharzil, line 5: Why this action is composite and the second one isn't?

shaharzil (commented):
Very very cool!

.github/workflows/build-and-upload.yml
YossiSaadi: yep :)
#2006 does exactly that and I will update all of our workflows once it is merged, including this new build-and-upload workflow

.github/actions/download-builds/action.yml
YossiSaadi: yep, it is happening in like 2-3s so we're cool with it

.github/actions/download-builds/action.yml
YossiSaadi: the build and upload can be a job of its own (reusable workflow)
reusable action is a `step` that encapsulates multiple `steps`
reusable workflow is a `job` that encapsulates multiple `jobs`

the reason I chose going with a reusable workflow for the `build-and-upload` is that in case we need the build, I'm saving the consumer from doing `checkout` and `setup`. the consumer can just do:
```
  build:
    name: Build
    uses: ./.github/workflows/build-and-upload.yml
    secrets:
      npm_token: ${{ secrets.npm_token }}
```

if the `build-and-upload` was a reusable action (`composite`), it would have looked like this:
```
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
      - name: Run Setup
        uses: ./.github/actions/setup
        with:
          npm_token: ${{secrets.npm_token}}
      - name: Run Build
        shell: bash
        run: yarn lerna run build --since=origin/master --include-dependencies
```

in probably 100% of the cases where consumers would need the build, they would just want it to happen, they don't need to interfere what happens with it

shaharzil (approved):

PR #2002, opened 2024-03-08
fix: add href support to actions in the Toast component

jashar11:
When passing in actions to a Toast, the href property is not acted upon if the type of the action is ToastActionType.BUTTON. With this fix, the url supplied to href is opened in a new tab when the button is clicked on.

### DIFF
--- a/packages/core/src/components/Button/__tests__/__snapshots__/button-snapshot-tests.jest.js.snap
+++ b/packages/core/src/components/Button/__tests__/__snapshots__/button-snapshot-tests.jest.js.snap
@@ -78,6 +78,23 @@ exports[`Button renders correctly renders Button color- fixed-light 1`] = `
 </button>
 `;
 
+exports[`Button renders correctly renders Button color- inverted 1`] = `
+<button
+  aria-busy={false}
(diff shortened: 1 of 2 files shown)


talkor:
@jashar11 thanks for contributing! When using a link in the Toast it should be provided with the `LINK` type and not a button.
For example:
```
const actions = useMemo(() => [{
      type: Toast.actionTypes.LINK,
      text: "Link to action",
      href: "https://google.com"
    }], []);
    return <Toast open actions={actions}>
        General message toast
      </Toast>;
  },
```
See live example [here](https://style.monday.com/?path=/story/feedback-toast--toast-with-link).
Since it's not a fix I'm closing this PR, if there's any else fix needed please create a new issue. Thanks!

jashar11:
I think this should still be categorised as a fix because if I simply pass 
```
{
      type: Toast.actionTypes.BUTTON,
      content: "Button text",
      href: "https://google.com"
    }
```
as the actions, I would expect the button click to take me to the external link. If not, then I don't see the point of having a button. At the very least, shouldn't there be an `onClick` property if a button is provided?

PR #2007, opened 2024-03-11
ci: align workflows standard to use the new reusable workflows and actions

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6234512700


YossiSaadi:
Notice that currently (also on `master`) Chromatic publishes even though the monorepo has no changes.
It happens only in the `push` event.

shaharzil (approved):

PR #2010, opened 2024-03-12
feat(AvatarGroupCounter): pass noAnimation prop

talkor:
https://monday.monday.com/boards/3532714909/pulses/6234996352


YossiSaadi (commented):
I believe you should also change extend the counter props in AvatarGroupCounterProps

YossiSaadi (approved):
Cool 💪🏼

PR #2020, opened 2024-03-18
fix(Combobox): clear selected option when query changes

talkor:
https://monday.monday.com/boards/3532714909/pulses/6272860665


YossiSaadi (approved):
oy 🤦🏼‍♂️

PR #2028, opened 2024-03-19
fix(EditableTypography): fix specificity issues when changing types a…

### LINKED ISSUE #1998: [Bug]: Editable heading type is not taken
### Vibe version

2.94.0

### Describe the bug

Hey, I try to use the editable heading and I tried to use the type and weights and change it to TEXT1 or other than normal it works only in edit state.
Attached a video showing the behaviour

### Expected behavior

Attached video - https://www.loom.com/share/087d5865a518495197886d91dd8a9e3b?sid=97295544-207b-4834-8bc0-e05d68f903c8

### Steps to reproduce

Just add an editable text and try to change it to type "TEXT2" and try to change the weight

### Reproduction example link

https://www.loom.com/share/087d5865a518495197886d91dd8a9e3b?sid=97295544-207b-4834-8bc0-e05d68f903c8

### System Info

_No response_

### Additional context, Screenshots

_No response_

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6244033346

fixes #1998

### DIFF
--- a/packages/core/src/components/EditableHeading/EditableHeading.tsx
+++ b/packages/core/src/components/EditableHeading/EditableHeading.tsx
@@ -40,6 +40,8 @@ const EditableHeading: VibeComponent<EditableHeadingProps, HTMLElement> & {
         data-testid={dataTestId || getTestId(ComponentDefaultTestId.EDITABLE_HEADING, id)}
         component={Heading}
         typographyClassName={getStyle(styles, camelCase(type + "-" + weight))}
+        type={type}
+        weight={weight}
         {...editableTypographyProps}
       />
     );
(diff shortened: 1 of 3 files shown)


YossiSaadi (approved):
Great! do we plan on removing the `typographyClassName` usage on both `EditableX` components? or is there a reason they're still there?

shaharzil:
> Great! do we plan on removing the `typographyClassName` usage on both `EditableX` components? or is there a reason they're still there?
> 
> edit: I wasn't clear I think, I know we agreed on leaving both for the solution, but do we plan on removing the class later?

It is problematic for 2 reasons:
1. It can be a breaking change
2. the input needs it, otherwise it won't have the correct style (can be solved of course with some specificity with the selector)

So yes, we want to remove it but not now :) (I can open a task for it)

YossiSaadi:
> > Great! do we plan on removing the `typographyClassName` usage on both `EditableX` components? or is there a reason they're still there?
> > edit: I wasn't clear I think, I know we agreed on leaving both for the solution, but do we plan on removing the class later?
> 
> It is problematic for 2 reasons:
> 
> 1. It can be a breaking change
> 2. the input needs it, otherwise it won't have the correct style (can be solved of course with some specificity with the selector)
> 
> So yes, we want to remove it but not now :) (I can open a task for it)

hmm got ya
we can create a task with a very low priority, yep

PR #2029, opened 2024-03-21
feat(IconButton): loading state

SergeyRoyt:
https://monday.monday.com/boards/3713040192/pulses/6317999430

Screenshot 2024-03-25 at 13 49 55


packages/core/src/components/IconButton/IconButton.module.scss
```diff
@@ -6,4 +6,16 @@
 
 .referenceWrapper {
   display: inline-flex;
-}
\ No newline at end of file
+}
+
+.loader.loader {
```
talkor, line 11: Why are we providing only a partial set of options? Let's give the ability to show the loader for all sizes, is it by design?

packages/core/src/components/IconButton/IconButton.module.scss
```diff
@@ -6,4 +6,16 @@
...
+  width: 16px;
+  height: 16px;
```
talkor, line 13: Do we have variables?

packages/core/src/components/IconButton/IconButton.module.scss
```diff
@@ -6,4 +6,16 @@
...
+  &.xxs {
+    width: 10px;
+    height: 10px;
+    svg {
+      display: block;
+    }
```
talkor, line 19: Why is it necessary? Is it inline by default?

packages/core/src/components/IconButton/IconButton.tsx
```diff
@@ -86,6 +87,8 @@ export interface IconButtonProps extends VibeComponentProps {
   insetFocus?: boolean;
   /** Specifies the tab order of an element */
   tabIndex?: number;
+  /** loading boolean which switches the text to a loader */
```
talkor, line 90:
```suggestion
  /** Show a loader instead of the icon  */
```

talkor (changes requested):

packages/core/src/components/IconButton/IconButton.module.scss
SergeyRoyt: It's showing for all of the sizes
We just need an override for the xxs size - 10 instead of 16

packages/core/src/components/IconButton/IconButton.module.scss
SergeyRoyt: Which variables would you use here?

packages/core/src/components/IconButton/IconButton.module.scss
SergeyRoyt: image

packages/core/src/components/IconButton/IconButton.module.scss
SergeyRoyt: image
It's inline by default, yeah
Without display block it isn't displaying properly

packages/core/src/components/IconButton/IconButton.tsx
talkor: @SergeyRoyt have you pushed it?

talkor (approved):
See two comments :)

PR #2050, opened 2024-04-07
fix(DatePicker): added min width to buttons inside year picker

shaharzil:
This bug happened since some of the years' widths are shorter than others (mostly years that start with 1), so the fix is to set a min width to the years' button

https://monday.monday.com/boards/3532714909/pulses/6234283969

### DIFF
--- a/packages/core/src/components/DatePicker/YearPicker/YearPicker.module.scss
+++ b/packages/core/src/components/DatePicker/YearPicker/YearPicker.module.scss
@@ -38,6 +38,10 @@
     position: absolute;
     top: 55px;
     pointer-events: all;
+
+    Button {
+      min-width: 66px;
+    }
   }
 
   :global(.CalendarYear--blocked) {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/DatePicker/YearPicker/YearPicker.module.scss
```diff
@@ -38,6 +38,10 @@
     position: absolute;
     top: 55px;
     pointer-events: all;
+
+    Button {
+      min-width: 66px;
```
talkor, line 43: Isn't there a class? Is it the `button` tag selector?

shaharzil: There isn't an existing class on the button (Except for the usual button classes). I considered adding a class but it didn't feel critical just for that. I can add if you feel it is important :)

talkor: it might cause some css leakage if the specificity is the same in other elements/component, we have a similar issue recently. It's rare but I think it's better to avoid targeting tag selectors all along

shaharzil: It is inside the scope of another class, and I made sure that there is no leak. the case we had before was different but Ill change it to be on the safe side :)

talkor (approved):

PR #2053, opened 2024-04-09
chore(BaseInput): reusable, internal Input component

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6437205266


packages/core/src/components/BaseInput/BaseInput.tsx
```diff
@@ -0,0 +1,35 @@
...
+      </div>
+    );
+  }
+);
+
+export default BaseInput;
+
+const App = () => {
+  return <BaseInput inputSize={"small"}></BaseInput>;
+};
```
talkor, line 35: What's this?

YossiSaadi: oops :)

packages/core/src/components/BaseInput/BaseInput.tsx
talkor, line 1: I'd call it just Input

packages/core/src/components/BaseInput/BaseInput.types.ts
talkor, line 1: Yes yes yes!!

packages/core/src/components/BaseInput/BaseInput.module.scss
```diff
@@ -0,0 +1,118 @@
+@import "../../styles/Typography";
+
+.wrapper {
+  width: 100%;
+  position: relative;
+  display: flex;
+  align-items: center;
+  gap: var(--spacing-small);
+  padding-block: var(--spacing-xs);
+  padding-inline: var(--spacing-medium) var(--spacing-xs);
```
talkor, line 10: Let's make it a standard?

packages/core/src/components/BaseInput/BaseInput.module.scss
```diff
@@ -0,0 +1,118 @@
...
+    height: 48px;
+    padding-block: var(--spacing-small);
+  }
+
+  &.underline {
+    border-left: none;
+    border-right: none;
+    border-top: none;
+    padding: var(--spacing-small) var(--spacing-large) var(--spacing-small) var(--spacing-xs);
+    border-radius: 0 !important;
```
talkor, line 41: Necesarry?

YossiSaadi: The PR was still in draft state, copied some of the logic from `TextField`, removed unnecessary stuff now

packages/core/src/components/BaseInput/BaseInput.module.scss
YossiSaadi: wdym?

packages/core/src/components/BaseInput/BaseInput.tsx
YossiSaadi: Hmm let's discuss it
I'm leaving it open for having an Input component which would be exported from Vibe

YossiSaadi:
added ability to hide stories here #2062 
make sure to merge this only after #2062 is merged

packages/core/src/components/BaseInput/BaseInput.module.scss
talkor: Logical properties

YossiSaadi: hmm I agree
Chrome 87+ support
we don't have users with less than 90, but anyway when we upgrade browsers it should be supported

packages/core/src/components/BaseInput/BaseInput.module.scss
```diff
@@ -0,0 +1,112 @@
...
+    &:hover {
+      border-color: var(--positive-color);
+    }
+
+    &:has(.input:active, .input:focus) {
+      border-color: var(--positive-color);
+    }
+  }
+
+  &:has(.input[aria-invalid="true"]) {
```
talkor, line 79: Didn't we say we can't use `has` for now?

packages/core/src/components/BaseInput/BaseInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+  (
+    {
+      size = "medium",
+      leftRender,
+      rightRender,
+      withReadOnlyStyle,
+      success,
+      error,
+      wrapperRole,
+      inputRole,
```
talkor, line 17: Can they get two different roles?

packages/core/src/components/BaseInput/__tests__/BaseInput.jest.tsx
```diff
@@ -0,0 +1,109 @@
...
+describe("BaseInput", () => {
+  it("should render correctly", () => {
+    const { getByLabelText } = renderBaseInput();
+    expect(getByLabelText("base-input")).toBeInTheDocument();
+  });
+
+  describe("with declared props", () => {
+    it("should apply the size class", () => {
+      const { getByLabelText } = renderBaseInput({ size: "large" });
+      expect(getByLabelText("base-input").parentNode).toHaveClass("large");
```
talkor, line 21: I think you can use `container` or something similar to get the outer element

packages/core/src/components/BaseInput/__tests__/BaseInput.jest.tsx
```diff
@@ -0,0 +1,109 @@
...
+      userEvent.type(input, "Hello, World!");
+      expect(input).toHaveValue("Hello, World!");
+    });
+
+    it("should call onChange on every input", () => {
+      const onChange = jest.fn();
+      const { getByLabelText } = renderBaseInput({ onChange });
+      const input = getByLabelText("base-input");
+      userEvent.type(input, "Hello, World!");
+      expect(onChange).toHaveBeenCalledTimes("Hello, World!".length);
```
talkor, line 82: Extract to a variable?

packages/core/src/components/BaseInput/__tests__/BaseInput.jest.tsx
```diff
@@ -0,0 +1,109 @@
...
+    });
+
+    it("should handle focus and blur events", () => {
```
talkor, line 85: Nice!

talkor (approved):
Very cool!

packages/core/src/components/BaseInput/BaseInput.module.scss
YossiSaadi: added support for both browsers that do and do not support has

packages/core/src/components/BaseInput/BaseInput.tsx
YossiSaadi: yep, for example in combobox, the wrapper should be "search" and the input "combobox" (it sounds weird, I know)

packages/core/src/components/BaseInput/__tests__/BaseInput.jest.tsx
YossiSaadi: container gives where the BaseInput is rendered, so I need to query the there the `firstChild` so it is pretty much the same..
I'll leave it the way it is I think

PR #2057, opened 2024-04-10
feat: add autoFocus to RadioButton

chensara:
- [ ] I have read the [Contribution Guide](../packages/core/CONTRIBUTING.md) for this project.

### DIFF
--- a/packages/core/src/components/RadioButton/RadioButton.tsx
+++ b/packages/core/src/components/RadioButton/RadioButton.tsx
@@ -1,5 +1,5 @@
 import cx from "classnames";
-import React, { forwardRef, useCallback, useMemo, useRef } from "react";
+import React, { forwardRef, useCallback, useEffect, useMemo, useRef } from "react";
 import useMergeRef from "../../hooks/useMergeRef";
 import Clickable from "../Clickable/Clickable";
 import Text from "../Text/Text";
@@ -27,6 +27,8 @@ export interface RadioButtonProps extends VibeComponentProps {
   value?: string;
   /** A string specifying a name for the input control. This name is submitted along with the control's value when the form data is submitted. */
   name?: string;
+  /** is autoFocus */
+  autoFocus?: boolean;
   /** is disabled */
   disabled?: boolean;
   /** why the input is disabled */
@@ -64,6 +66,7 @@ const RadioButton: VibeComponent<RadioButtonProps, HTMLElement> & object = forwa
        */
       radioButtonClassName,
       disabled = false,
+      autoFocus = false,
       disabledReason,
       defaultChecked = false,
       children,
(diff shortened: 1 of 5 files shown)


packages/core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -114,6 +127,7 @@ const RadioButton: VibeComponent<RadioButtonProps, HTMLElement> & object = forwa
               type="radio"
               value={value}
               name={name}
+              autoFocus={autoFocus}
```
YossiSaadi, line 130: I'm not sure if both methods are needed, the `useEffect` and the `autoFocus`.
I know that `TextField` implements it using the `useEffect` with the animation frame, which I guess is for a reason.
@talkor wdyt?

chensara, line 130: Actually here I'm not sure that I need this line, pls advice

talkor: Both should work but IMO it's better to just use the `autoFocus` prop which is an HTML attribute which makes the useEffect redundant in this case

YossiSaadi: @chensara let's do the following
use the native `autoFocus` instead of the `useEffect`.

- let's release a prerelease
- or just use `yarn pack` on `monday-ui-react-core`'s `packages/core` and use it in your local version
- you can also use `yarn link`, which is better IMO, with your MF. 
Make sure that you run the command using the same node version that is used on your MF, in Vibe we're using node v18.12.1
Also make sure to run `yarn link monday-ui-react-core` afterwards on your MF of course
Notice in this case that you need to run `yarn build` for every change to take effect on your MF

if it works for you with the native approach, amazing. if not, try the same with the `useEffect`

chensara: So I first implemented using only `autoFocus` attr but it doesn't work in stroybook so I added the useEffect and it worked. Should I delete `autoFocus` attr?

YossiSaadi: when I've checked with the `autoFocus` it seems like working

packages/core/src/components/RadioButton/RadioButton.tsx
```diff
@@ -64,6 +66,7 @@ const RadioButton: VibeComponent<RadioButtonProps, HTMLElement> & object = forwa
        */
       radioButtonClassName,
       disabled = false,
+      autoFocus = false,
```
YossiSaadi, line 69: maybe leave it `undefined` in case it isn't supplied
so current snapshots won't add `autoFocus=false` to their snapshots, as it can lead to - when upgrading a version in monolith - to multiple snapshots being broken
we're ok with it as `undefined`
so just remove the default `false`

packages/core/src/components/RadioButton/__tests__/radioButton-tests.jest.tsx
```diff
@@ -1,76 +1,99 @@
...
+    beforeEach(() => {
+      const radiosName = "radios";
 
-  beforeEach(() => {
-    onChangeMock1 = jest.fn();
-    onChangeMock2 = jest.fn();
-    onChangeMock3 = jest.fn();
+      const option1Value = "1";
+      const option2Value = "2";
+      const option3Value = "3";
```
YossiSaadi, line 22: I agree with removing the mocks from the `beforeEach` section, I'm not sure about inserting the `*Value` inside, and also about `radiosName`, sounds like they can lay outside of this block (even in the top of the file, outside of any block)

packages/core/src/components/RadioButton/RadioButton.tsx
chensara: Removed :)

packages/core/src/components/RadioButton/__tests__/radioButton-tests.jest.tsx
chensara: Done

YossiSaadi (approved):
Looks awesome! congrats on your first contribution to Vibe! 🎉

PR #2062, opened 2024-04-11
docs: allow having Internal section with stories only visible for dev mode

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6437258196

image

### DIFF
--- a/packages/core/.storybook/manager.jsx
+++ b/packages/core/.storybook/manager.jsx
@@ -24,6 +24,13 @@ addons.setConfig({
       }
 
       return <SidebarItem status={parameters.status}>{name.replace(storyStatus, "").trim()}</SidebarItem>;
+    },
+    filters: {
+      patterns: filterInternalStoryDocsPage
     }
   }
 });
+
+function filterInternalStoryDocsPage(item) {
+  return process.env.STORYBOOK_ENV === "development" || !item.tags?.includes?.("internal");
+}
--- a/packages/core/.storybook/manager.jsx
+++ b/packages/core/.storybook/manager.jsx
@@ -33,7 +33,6 @@ addons.setConfig({
 });
 
 function filterStory(item) {
-  console.log(process.env.NODE_ENV === "development", item.title, ...item.tags, { item });
   const isDev = isChromatic() || process.env.NODE_ENV === "development";
   const isInternal = !item.tags?.includes?.("internal") && !item.title?.startsWith?.("Internal");
   return isDev || isInternal;
--- a/packages/core/.storybook/manager.jsx
+++ b/packages/core/.storybook/manager.jsx
@@ -1,8 +1,9 @@
-import { addons } from "@storybook/manager-api";
 import React from "react";
+import { addons } from "@storybook/manager-api";
 import { SidebarItem } from "vibe-storybook-components";
 import "vibe-storybook-components/index.css";
 import theme from "./theme";
+import isChromatic from "chromatic/isChromatic";
 
 window.STORYBOOK_GA_ID = "UA-308574295";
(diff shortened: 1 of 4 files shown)


packages/core/.storybook/manager.jsx
```diff
@@ -24,6 +24,13 @@ addons.setConfig({
...
       return <SidebarItem status={parameters.status}>{name.replace(storyStatus, "").trim()}</SidebarItem>;
+    },
+    filters: {
+      patterns: filterInternalStoryDocsPage
     }
   }
 });
+
+function filterInternalStoryDocsPage(item) {
+  return process.env.STORYBOOK_ENV === "development" || !item.tags?.includes?.("internal");
```
talkor, line 35: Can't we filter the entire section without the tags?

talkor, line 35: Storybook uses env variables, I think you can get them somehow here

YossiSaadi: I've tried to also use their support, and their docs, couldn't find anything related

YossiSaadi: basically, we can say that if the name includes Internal.
I could add it as an extra check, and leave the current, just for cases where if we would like to have an internal story inside an existing category. wdyt?

YossiSaadi: would love to know if you find something, I'll change it 🙏🏼

talkor: You can use `process.env.NODE_ENV`

talkor: It should work also with internals stories, no? Anyway I think it's fine for now with tags. It's a new feature in SB7 no? We can use them for the beta/deprecated tags as well

YossiSaadi: Are they setting it?

YossiSaadi: - yep it is
- you mean adding styled tags?
- and if I remove the check for `tags` and only exclude the `Internal` category, then it won't exclude internal stories that are inside regular (exported) categories

talkor: Yes tried it locally also, and they use `production` for build

talkor: No I mean the sidebar tags, we use the name to decide whether a story is a beta or deprecated, I think using tags is better for that purpose. Let's stay with tags, forget about what I said 😅

YossiSaadi: what do they supply in `dev`?
I'm not sure if we're exporting as production build

YossiSaadi: ohhh I got you
TechDebt task 🙃

YossiSaadi: https://monday.monday.com/boards/3532714909/pulses/6453084060

packages/core/.storybook/manager.jsx
```diff
@@ -24,6 +25,16 @@ addons.setConfig({
...
       return <SidebarItem status={parameters.status}>{name.replace(storyStatus, "").trim()}</SidebarItem>;
+    },
+    filters: {
+      patterns: filterStory
     }
   }
 });
+
+function filterStory(item) {
+  console.log(process.env.NODE_ENV === "development", item.title, ...item.tags, { item });
```
talkor, line 36: lol

talkor (approved):
Awesome, just remove the console

packages/core/.storybook/manager.jsx
YossiSaadi: 🙃

YossiSaadi: Thanks!

PR #2065, opened 2024-04-15
fix(Dialog): add onContextMenu to dialog's show trigger

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6453915410


YossiSaadi (approved):
💪🏼

PR #2067, opened 2024-04-15
feat: migrate usages of LegacySearch to new Search

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6450989453

### DIFF
--- a/packages/core/src/components/Combobox/Combobox.module.scss
+++ b/packages/core/src/components/Combobox/Combobox.module.scss
@@ -39,10 +39,6 @@
   z-index: 12;
 }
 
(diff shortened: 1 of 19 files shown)


packages/core/src/components/Combobox/Combobox.module.scss
```diff
@@ -39,10 +39,6 @@
   z-index: 12;
 }
 
-.comboboxSearch {
-  flex: 0 0 auto;
-}
-
```
YossiSaadi, line 45: @talkor I need to tell you why it was here, amazing 🤦🏼

PR #2069, opened 2024-04-16
docs(Dropdown): async options story

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4977987525


packages/core/src/components/Dropdown/__stories__/dropdown.stories.js
```diff
@@ -272,6 +272,37 @@ export const MultiChoiceWithDifferentStates = {
   play: multiInteractionTests
 };
 
+export const AsyncOptions = {
+  render: () => {
+    const fetchUserOptions = async () => {
+      try {
+        const response = await fetch("https://jsonplaceholder.typicode.com/users");
```
talkor, line 279: Chromatic don't recommend dynamic apis for stories as they change, can we make it static?

talkor (commented):
That's great!! We just need to make sure it won't fail chromatic

talkor (approved):

PR #2072, opened 2024-04-16
feat(Text): Add new 12px font size (Text3)

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6418938469

### DIFF
--- a/packages/core/src/components/Text/__stories__/Text.stories.js
+++ b/packages/core/src/components/Text/__stories__/Text.stories.js
@@ -35,29 +35,57 @@ export const Overview = {
 export const SizesAndWeights = {
   render: () => (
     <Flex
-      gap={Flex.gaps.SMALL}
+      gap={Flex.gaps.LARGE}
       direction={Flex.directions.COLUMN}
       justify={Flex.justify.START}
       align={Flex.align.START}
     >
-      <Text type={Text.types.TEXT1} weight={Text.weights.BOLD}>
-        This is text1 bold
-      </Text>
-      <Text type={Text.types.TEXT1} weight={Text.weights.MEDIUM}>
-        This is text1 medium
-      </Text>
-      <Text type={Text.types.TEXT1} weight={Text.weights.NORMAL}>
-        This is text1 normal
-      </Text>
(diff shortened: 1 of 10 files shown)


packages/core/src/components/Text/__stories__/Text.stories.js
```diff
@@ -35,29 +35,57 @@ export const Overview = {
...
+          This is text1 bold
+        </Text>
+        <Text type={Text.types.TEXT1} weight={Text.weights.MEDIUM}>
+          This is text1 medium
+        </Text>
+        <Text type={Text.types.TEXT1} weight={Text.weights.NORMAL}>
+          This is text1 normal
+        </Text>
+      </Flex>
+      {/* todo - add spaces and dos and donts */}
```
talkor, line 59: Comment

packages/core/src/components/Text/__stories__/Text.stories.js
```diff
@@ -35,29 +35,57 @@ export const Overview = {
...
+          This is text2 bold
+        </Text>
+        <Text type={Text.types.TEXT2} weight={Text.weights.MEDIUM}>
+          This is text2 medium
+        </Text>
+        <Text type={Text.types.TEXT2} weight={Text.weights.NORMAL}>
+          This is text2 normal
+        </Text>
+      </Flex>
+      <Flex
```
talkor, line 76: We said that we're gonna support bold but not document it, right?

talkor (approved):
Nice!

packages/core/src/components/Text/__stories__/Text.stories.js
YossiSaadi: Don't we want to disable it typescript-wise?
and to not allow the mixin variants for it?

shaharzil: No, we don't want to block them so it won't look like a bug. we will just guide them to use It right (in Figma)

YossiSaadi: got ya 👍🏼

PR #2077, opened 2024-04-21
feat(withLiveEdit): apply decorators from within self CSF module of a story

YossiSaadi:
this should apply decorators which are declared in the same object of the story, not only global decorators, so we can make our stories cleaner

https://monday.monday.com/boards/3532714909/pulses/6494412165


talkor (approved):
Awesome!!

PR #2078, opened 2024-04-21
feat(withLiveEdit): parse render attribute with ast instead of with regex for variety of cases

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6494355245


talkor (approved):
This is so much better than the regex, does it cover all cases or most of them?

YossiSaadi:
> This is so much better than the regex, does it cover all cases or most of them?

100% of cases 💪🏼

PR #2079, opened 2024-04-21
feat(createStoryMetaSettings): allow generating internal stories and autodocs page with metasettings helper

YossiSaadi:
## **Note**: this PR is on hold, because of a bug from Storybook's side
This PR aims to manage the tags "backstage" for internal stories
It also enables creating autodocs page, with not much info (would mainly be used for internal components)

image

image

### DIFF
--- a/packages/core/.storybook/main.ts
+++ b/packages/core/.storybook/main.ts
@@ -34,9 +34,6 @@ const config: StorybookConfig = {
     name: "@storybook/react-webpack5",
     options: {}
   },
-  docs: {
-    autodocs: false
-  },
   async webpackFinal(config, { configType }) {
     if (configType === "DEVELOPMENT") {
       if (config.resolve) {
--- a/packages/core/src/components/BaseInput/__stories__/BaseInput.stories.tsx
+++ b/packages/core/src/components/BaseInput/__stories__/BaseInput.stories.tsx
@@ -3,7 +3,9 @@ import { createComponentTemplate } from "vibe-storybook-components";
 import BaseInput from "../BaseInput";
 
 const metaSettings = createStoryMetaSettingsDecorator({
-  component: BaseInput
+  component: BaseInput,
+  isInternal: true,
+  shouldCreateAutodocsPage: true
 });
 
 export default {
@@ -11,7 +13,7 @@ export default {
   component: BaseInput,
   argTypes: metaSettings.argTypes,
   decorators: metaSettings.decorators,
-  tags: ["internal"]
+  tags: metaSettings.tags
 };
 
 const baseInputTemplate = createComponentTemplate(BaseInput);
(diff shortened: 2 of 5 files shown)


packages/core/.storybook/main.ts
```diff
@@ -34,9 +34,6 @@ const config: StorybookConfig = {
     name: "@storybook/react-webpack5",
     options: {}
   },
-  docs: {
-    autodocs: false
```
YossiSaadi, line 38: default is not `true`, default is `tags`, meaning only stories which define `tags: ["autodocs"]` would get autodocs page. It's good for internal components to get a "free" docs page

packages/core/src/components/BaseInput/__stories__/BaseInput.stories.tsx
```diff
@@ -3,15 +3,17 @@ import { createComponentTemplate } from "vibe-storybook-components";
 import BaseInput from "../BaseInput";
 
 const metaSettings = createStoryMetaSettingsDecorator({
-  component: BaseInput
+  component: BaseInput,
+  isInternal: true,
+  shouldCreateAutodocsPage: true
```
YossiSaadi, line 8: just to make the magic "backstage" instead of declaring `tags: ["internal", "autodocs"]`.

packages/core/.storybook/main.ts
talkor: You mean `autodocs: 'tag'` no?

packages/core/src/components/BaseInput/__stories__/BaseInput.stories.tsx
talkor: I wish we could just add `internal: true` and that's it, and get the autodocs page, we need to "register" it in 3 different properties

YossiSaadi: @talkor done
autodocs is now opt-out optional if internal is passed

talkor (approved):
Awesome

YossiSaadi:
> Awesome

@talkor Can't determine why it fails.
When I print the metasettings.tags it returns a proper array

YossiSaadi:
> > > Awesome
> > 
> > 
> > @talkor Can't determine why it fails. When I print the metasettings.tags it returns a proper array
> 
> What fails?

```
WARN 🚨 Unable to index ./src/components/BaseInput/__stories__/BaseInput.stories.tsx:
WARN   Error: CSF: Expected tags array
```

will leave it for now

YossiSaadi:
@talkor https://github.com/storybookjs/storybook/issues/26940
It's a bug on Storybook's side

talkor:
> @talkor [storybookjs/storybook#26940](https://github.com/storybookjs/storybook/issues/26940) It's a bug on Storybook's side

oh no :/

YossiSaadi:
Storybook's not planning on fixing it, closing

PR #2087, opened 2024-04-24
fix(Dialog): prevent default for contextmenu event

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6514184411

### DIFF
--- a/packages/core/src/components/Dialog/Dialog.tsx
+++ b/packages/core/src/components/Dialog/Dialog.tsx
@@ -457,6 +457,13 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
   }
 
   onContextMenu(e: React.MouseEvent) {
+    const isShown = this.isShown();
+    if (
+      (this.isShowTrigger(HideShowEvent.CONTEXT_MENU) && !isShown) ||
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -457,6 +457,13 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
   }
 
   onContextMenu(e: React.MouseEvent) {
+    const isShown = this.isShown();
+    if (
+      (this.isShowTrigger(HideShowEvent.CONTEXT_MENU) && !isShown) ||
+      (this.isHideTrigger(HideShowEvent.CONTEXT_MENU) && isShown)
```
talkor, line 463: Can it mess up existing usages? Because it was already before, no?

talkor (approved):

packages/core/src/components/Dialog/Dialog.tsx
shaharzil: only the hide was in use, I talked with meirav and she want it to chenge to the existing users

PR #2091, opened 2024-04-30
fix: hide RenderAction when have inputValue

jessbemonday:
Looking to make a small change to the logic of the new search component - we don't want the renderAction component to show once we have a search term entered.

Currently there's not a way to do this that doesn't result in both the renderAction and the clear button both showing briefly (see below), due the input value change inside the search component not being debounced, so the clear button renders before the onChange method is called and we're able to check for the presence of an input value outside the search component.

Current behaviour when handling hiding renderAction outside of search component:

### DIFF
--- a/packages/core/src/components/Search/Search.tsx
+++ b/packages/core/src/components/Search/Search.tsx
@@ -88,7 +88,7 @@ const Search = forwardRef(
           />
         )}
         {inputValue && !disabled && ClearIcon}
-        {RenderAction}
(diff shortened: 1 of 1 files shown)


jessbemonday:
Decided against this approach! See slack here: https://monday.slack.com/archives/C02S6F817JQ/p1714484831342629?thread_ts=1714481098.302499&cid=C02S6F817JQ

PR #2092, opened 2024-04-30
feat(Dialog): add event args to dialog show methods

yair65:
- added exposing event handling for onDialogDidShow


packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -113,7 +113,7 @@ export interface DialogProps extends VibeComponentProps {
   /**
    * callback to be called when the dialog is shown
    */
-  onDialogDidShow?: () => void;
+  onDialogDidShow?: (event?: DialogEvent, eventName?: HideShowEvent | string) => void;
```
YossiSaadi, line 116: btw I'm not sure they should be optional in here

PR #2095, opened 2024-05-01
feat(colors-json): add correct warning colors syntax

YossiSaadi:
since it is a breaking change to remove the `color-warning*` tokens, we're adding in addition the correct `warning-color*` and not replacing

https://monday.monday.com/boards/3532714909/pulses/6557427601


talkor (approved):
💪🏼

PR #2098, opened 2024-05-01
feat(Search): adds hideRenderActionOnInput option

jessbemonday:
We have a use case for being able to hide the renderAction component passed in to the Search component once a search term is entered, which is not easily achieved by logic outside of the component (see below, clear icon showing at same time as renderAction).

Adding a `hideRenderActionOnInput` option, that defaults to false, which can be enabled so that the RenderAction is not displayed once a search term is entered.

Current behaviour when handling hiding renderAction outside of search component:

### DIFF
--- a/packages/core/src/components/Search/Search.tsx
+++ b/packages/core/src/components/Search/Search.tsx
@@ -19,6 +19,7 @@ const Search = forwardRef(
       clearIconName = CloseSmallIcon,
       clearIconLabel = "Clear",
       renderAction: RenderAction,
+      hideRenderActionOnInput = false,
       value,
       placeholder,
       size = "medium",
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Search/Search.tsx
```diff
@@ -19,6 +19,7 @@ const Search = forwardRef(
       clearIconName = CloseSmallIcon,
       clearIconLabel = "Clear",
       renderAction: RenderAction,
+      hideRenderActionOnInput = false,
```
YossiSaadi, line 22: we're fine with it default to `undefined`

jessbemonday: updated, thanks @YossiSaadi

YossiSaadi (approved):
Awesome! congrats on your first contribution to Vibe! 🎉

YossiSaadi (approved):
🚀

PR #2099, opened 2024-05-02
fix(MenuItem): disabled item should not show its submenu

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6481701216


talkor (approved):
👑

Issue #2045, opened 2024-04-03
[Bug]: DatePicker | no null protection for onPickDate callback


talkor:
Using the DatePicker component without the `onPickDate` prop provided results in a console error.
Suggested solution: add null protection for the `onPickDate` prop.

<img width="480" alt="Screenshot 2024-04-03 at 13 53 44" src="https://github.com/mondaycom/monday-ui-react-core/assets/18269880/52dafbc0-2485-4390-afa8-2fa8ec98dc86">

monday ticket: https://monday.monday.com/boards/3532714909/pulses/6381326498

YossiSaadi:
resolved at #2102

PR #2102, opened 2024-05-03
fix: add null protection for onPickDate callback (#2045)

dvir1225:
add a null check to avoid a console error when using the DatePicker component without passing the onPickDate prop, to solve this issue:
https://github.com/mondaycom/vibe/issues/2045

### DIFF
--- a/packages/core/src/components/DatePicker/DatePicker.tsx
+++ b/packages/core/src/components/DatePicker/DatePicker.tsx
@@ -116,6 +116,7 @@ const DatePicker: VibeComponent<DatePickerProps, HTMLElement> = forwardRef<HTMLD
 
     const onDateRangeChange = useCallback(
       (date: RangeDate) => {
+        if (!onPickDate) return;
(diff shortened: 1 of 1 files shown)


talkor (approved):
Thank you @dvir1225!

YossiSaadi:
solves #2045

YossiSaadi:
resolves #2045

packages/core/src/components/DatePicker/DatePicker.tsx
```diff
@@ -116,6 +116,7 @@ const DatePicker: VibeComponent<DatePickerProps, HTMLElement> = forwardRef<HTMLD
 
     const onDateRangeChange = useCallback(
       (date: RangeDate) => {
+        if (!onPickDate) return;
```
ShayKlig, line 119: Not exactly sure what is the library convention, but i believe in this way it is more clear.
Another option is run the onPickDate?.(....)
```suggestion
        if (!onPickDate) { 
            return;
        }
```

PR #2105, opened 2024-05-05
feat: add a11y props for `search` and `combobox` components

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6453545563


packages/core/src/components/Search/Search.tsx
```diff
@@ -116,8 +118,10 @@ const Search = forwardRef(
         inputRole={searchResultsContainerId ? "combobox" : undefined}
         aria-label={inputAriaLabel}
         aria-busy={loading}
-        aria-owns={searchResultsContainerId}
+        aria-controls={ariaExpanded ? searchResultsContainerId : undefined}
         aria-activedescendant={currentAriaResultId}
+        aria-haspopup={ariaHasPopup}
```
YossiSaadi, line 123: from what I know, it basically says that the Search input can trigger an open of a popup
I'm not sure we have such case
in Combobox, the Search is part of the dialog, and does not trigger an open of a popup

packages/core/src/components/Search/Search.tsx
```diff
@@ -116,8 +118,10 @@ const Search = forwardRef(
...
+        aria-expanded={ariaExpanded}
```
YossiSaadi, line 124: this should be set for elements that have child elements that are either shown or hidden, same in this case - Search does not have child elements

packages/core/src/components/Search/Search.tsx
```diff
@@ -116,8 +118,10 @@ const Search = forwardRef(
         inputRole={searchResultsContainerId ? "combobox" : undefined}
         aria-label={inputAriaLabel}
         aria-busy={loading}
-        aria-owns={searchResultsContainerId}
+        aria-controls={ariaExpanded ? searchResultsContainerId : undefined}
```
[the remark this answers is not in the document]
YossiSaadi: in addition, after another read on the differences (wasn't aware they are related), it sounds like `controls` mean that it can be controlled through other element, e.g. if I choose something from some other component, it would change the Search input content

with `owns`, we said that it is contained in another element, even though they're not in the same DOM hierarchy.

packages/core/src/components/Combobox/components/ComboboxItems/ComboboxItems.tsx
```diff
@@ -128,17 +130,19 @@ export const ComboboxItems: React.FC<ComboboxItemsProps> = forwardRef(
           className={cx(styles.optionsContainer, className)}
           items={options}
           itemRenderer={createItemElementRenderer}
-          role="treegrid"
+          role="listbox"
```
YossiSaadi, line 133: sounds correct 💪🏼

packages/core/src/components/Combobox/components/ComboboxItems/ComboboxItems.tsx
```diff
@@ -128,17 +130,19 @@ export const ComboboxItems: React.FC<ComboboxItemsProps> = forwardRef(
...
           style={style}
+          id={id}
         />
       );
     } else {
       itemsElements = (
         <div
           className={cx(styles.scrollableContainer, styles.optionsContainer, className)}
-          role="treegrid"
+          role="listbox"
```
YossiSaadi, line 144: I wonder if we have a case where combobox items are collapsible
if so, we should allow passing custom role, or have a boolean that should indicate whether it is a list or a tree

I think the change here is correct, I just wonder about the breaking change

packages/core/src/components/Search/Search.tsx
```diff
@@ -116,8 +118,10 @@ const Search = forwardRef(
         inputRole={searchResultsContainerId ? "combobox" : undefined}
```
YossiSaadi, line 118: I think while trying to keep the functionality of the original Search I made a mistake here
the input should not have a searchbox role, the wrapper should not have a role at all, it doesn't make sense 🤦🏼

YossiSaadi (commented):
Just ignore all my comments, and focus on this summary, or just use them as reference.

I think our result should look like this:
```typescript
<dialog and its props>
	<div role="combobox" aria-expanded="false/true (if the dialog is opened)" aria-haspopup="listbox" aria-owns="options-container" aria-label="My awesome combobox that should do this and that"> // this should be the dialog content container
	  <div id="search-wrapper"> // we have a wrapper for the search input, it shouldn't get any a11y prop I believe
// from mdn: The aria-autocomplete attribute indicates whether inputting text could trigger display of one or more predictions of the user's intended value for a combobox and specifies how predictions will be presented if they are made.
	    <input type="text" role="searchbox" aria-controls="combobox-items-container" aria-autocomplete="list">
	  </div>
	  <div role="listbox" id="combobox-items-container">
	    <div role="option" aria-selected="false">Option 1</div>
	    <div role="option" aria-selected="true">Option 2</div>
	  </div>
	</div>
</dialog>
```

Wanna discuss it?

YossiSaadi (approved):
after discussing everything, I understood aria-controls changed all the hierarchy so everything makes sense
Amazing!!

PR #2110, opened 2024-05-08
feat: added reason to toast close callback

moro256:
- add close reason to the toast component's `onClose()` callback

### DIFF
--- a/packages/core/src/components/Toast/Toast.tsx
+++ b/packages/core/src/components/Toast/Toast.tsx
@@ -11,7 +11,7 @@ import Flex from "../Flex/Flex";
 import CloseSmall from "../Icon/Icons/components/CloseSmall";
 import ToastLink from "./ToastLink/ToastLink";
 import ToastButton from "./ToastButton/ToastButton";
-import { ToastAction, ToastActionType, ToastType } from "./ToastConstants";
+import { ToastAction, ToastActionType, ToastCloseReason, ToastType } from "./ToastConstants";
 import { getIcon } from "./ToastHelpers";
 import { NOOP } from "../../utils/function-utils";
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Toast/Toast.tsx
```diff
@@ -102,7 +105,7 @@ const Toast: FC<ToastProps> & { types?: typeof ToastType; actionTypes?: typeof T
 
       clearTimeout(timerAutoHide.current);
       timerAutoHide.current = setTimeout(() => {
-        handleClose();
+        handleClose(ToastCloseReason.TIMEOUT);
```
YossiSaadi, line 108: let's convert `ToastCloseReason` to a type, and instead of accessing it through `ToastCloseReason`, just send "timeout" there.
so `ToastCloseReason` should look like:
```
type ToastCloseReason = "timeout" | "close-click"
```

packages/core/src/components/Toast/Toast.tsx
```diff
@@ -153,7 +156,7 @@ const Toast: FC<ToastProps> & { types?: typeof ToastType; actionTypes?: typeof T
         {closeable && (
           <IconButton
             className={cx(styles.closeButton)}
-            onClick={handleClose}
+            onClick={() => handleClose(ToastCloseReason.CLOSE_CLICKED)}
```
YossiSaadi, line 159: same

YossiSaadi (commented):
we want to take a different direction, instead of using enums, to rely more on typescript
can you please change the reason to type?

YossiSaadi:
@moro256 also, please add few tests for it, to expect different outcomes of reason on onClose

YossiSaadi:
Hey @moro256 , checking in on this one, is it still needed?

PR #2111, opened 2024-05-08
refactor(MenuItem): separate basic logic from MenuItem to BaseMenuItem

YossiSaadi:
https://monday.monday.com/boards/3532715121/pulses/6600529809


packages/core/src/components/Menu/MenuItem/components/BaseMenuItem/BaseMenuItem.tsx
```diff
@@ -0,0 +1,164 @@
...
+              disabled={disabled}
+            />
+            <MenuItemSubMenu
+              anchorRef={anchorRef}
+              open={shouldShowSubMenu}
+              onClose={closeSubMenu}
+              autoFocusOnMount={!useDocumentEventListeners}
+            >
+              {subMenu}
+            </MenuItemSubMenu>
```
talkor, line 156: Why are these in the Base and not in the MenuItem?

packages/core/src/components/Menu/MenuItem/components/BaseMenuItem/BaseMenuItem.types.ts
```diff
@@ -0,0 +1,31 @@
+import { MenuItemProps } from "../../MenuItem";
+import { MenuChild } from "../../../Menu/MenuConstants";
+import React from "react";
+import { VibeComponentProps } from "../../../../../types";
+
+export type LossMenuItemProps = Omit<
```
talkor, line 6: What's this for?

packages/core/src/components/Menu/MenuItem/components/BaseMenuItem/BaseMenuItem.tsx
YossiSaadi: cause I want to be able to add sub menu for CustomMenuItem and for each MenuItem variant
MenuItem is an opinionated implemenation now, but the basic should have the capabilities for sub menu

packages/core/src/components/Menu/MenuItem/components/BaseMenuItem/BaseMenuItem.types.ts
YossiSaadi: to not cause breaking change, I'm taking the MenuItem props, but just what I need, and to override those who are conflicting

packages/core/src/components/Menu/MenuItem/components/BaseMenuItem/BaseMenuItem.tsx
talkor: Is this intended to replace the option to pass a component to the menu item using the title prop?

YossiSaadi: CustomMenuItem is the one intended to replace, basically yes.
but it puts more weight on the developer to "know what he's doing".
BaseMenuItem is internal

talkor: Sounds legit to me, but I'm wondering why should devs use the old MenuItem currently and not just the new CustomMenuItem? It gives them more flexibility in the first place and behaves the same

YossiSaadi: they shouldn't basically
only for tooltip reasons (if they want)

PR #2112, opened 2024-05-08
feat(CustomMenuItem): export custom menu item component to be able to render anything inside

YossiSaadi:
https://monday.monday.com/boards/3532715121/pulses/6602051386


packages/core/src/components/Menu/MenuItem/components/MenuItemSubMenuIcon/MenuItemSubMenuIcon.module.scss
```diff
@@ -1,5 +1,6 @@
 .subMenuIconWrapper {
-  margin-left: var(--spacing-xs);
+  margin-left: auto;
+  padding-left: var(--spacing-xs);
```
YossiSaadi, line 3: to be able to push the icon always to the right, regardless to the size of the content
as for today, MenuItem just adds flex-grow 1, which is ok, and then sub menu icon is pushed to the right, but we need something that **always** work, regardless of what the child is
we cannot expect people to implement flex-grow 1 for themselves, the sub menu icon should be pushed to the right by default

talkor (approved):

YossiSaadi:
Decided to not achieve this capability for now, we want menu to just be menu and to not allow weird menu items

PR #2121, opened 2024-05-15
feat(label): add small variant

rivka-ungar:
https://monday.monday.com/boards/3532714909/pulses/6623032676


packages/core/src/components/Label/Label.tsx
```diff
@@ -12,6 +12,14 @@ import useClickableProps from "../../hooks/useClickableProps/useClickableProps";
 import useMergeRef from "../../hooks/useMergeRef";
 import styles from "./Label.module.scss";
 import LabelCelebrationAnimation from "./LabelCelebrationAnimation";
+import { TextType } from "../Text/TextConstants";
+
+type Sizes = "small" | "medium";
```
talkor, line 17: Let's move it to a new file called `Label.types.ts` (for refrences: `Search.types.ts`, `TextArea.types.ts` etc) as it's the new convention we're advancing

packages/core/src/components/Label/Label.tsx
```diff
@@ -12,6 +12,14 @@ import useClickableProps from "../../hooks/useClickableProps/useClickableProps";
...
+
+const mapSizesToTextSize: Record<Sizes, TextType> = {
```
talkor, line 19: Excellent

packages/core/src/components/Label/__stories__/label.stories.tsx
```diff
@@ -56,6 +56,27 @@ export const Kinds = {
...
 
+export const Sizes = {
+  render: () => (
+    <>
+      <div className="monday-storybook-label_group monday-storybook-label_states-gap">
+        <Label text="New" />
+      </div>
+      <div className="monday-storybook-label_group monday-storybook-label_states-gap">
+        <Label text="New" size="small" />
+      </div>
```
talkor, line 67: To avoid the divs appear in the code snippet in storybook you can use a decorator, you can see some examples in `TextArea.stories.tsx`. It will make it less cluttered

talkor (commented):
Great! See two minor comments, regarding types file and storybook decorators

packages/core/src/components/Label/Label.module.scss
```diff
@@ -147,6 +147,15 @@
...
+
+  .smallText {
+    line-height: 14px;
+  }
+}
+
+.small {
+  padding: 0 var(--spacing-xs);
+  border-radius: 2px;
 }
```
talkor, line 159: Let's nest it inside label, it will increase specificity

packages/core/src/components/Label/Label.tsx
```diff
@@ -7,29 +7,12 @@ import { backwardCompatibilityForProperties } from "../../helpers/backwardCompat
...
 import Leg from "./Leg";
 import { LabelColor, LabelKind } from "./LabelConstants";
-import { VibeComponent, VibeComponentProps, withStaticProps } from "../../types";
+import { VibeComponent, withStaticProps } from "../../types";
 import useClickableProps from "../../hooks/useClickableProps/useClickableProps";
 import useMergeRef from "../../hooks/useMergeRef";
 import styles from "./Label.module.scss";
 import LabelCelebrationAnimation from "./LabelCelebrationAnimation";
-
-export interface LabelProps extends VibeComponentProps {
```
talkor, line 16: Let's still export it also in this file to avoid breaking usages where users take the type for dist

packages/core/src/components/Label/Label.types.ts
talkor, line 1: Awesome!!

packages/core/src/components/Label/__stories__/label.stories.tsx
```diff
@@ -18,6 +21,21 @@ export default {
   decorators: metaSettings.decorators
 };
 
+const withGrid: Decorator = Story => (
+  <div
```
talkor, line 25: That's great! We want to extract these to storybook layout components and use them all over

packages/core/src/components/Label/__stories__/label.stories.tsx
```diff
@@ -56,28 +73,37 @@ export const Kinds = {
...
+  parameters: {
+    chromatic: {
+      pauseAnimationAtEnd: true
+    }
+  }
+};
+
+export const Colors = {
+  render: () => (
+    <>
```
talkor, line 95: This is so much leaner now

talkor (commented):
Great! Let's just export the LabelProps interface

packages/core/src/components/Label/Label.tsx
rivka-ungar: Done

packages/core/src/components/Label/Label.module.scss
rivka-ungar: Done

talkor (approved):
🎉

PR #2124, opened 2024-05-16
fix(Dialog.tsx): getContainer causing react error 200

yair65:
When document.querySelector(containerSelector) returned a `node` that was not a `Element` the react-dom createPortal threw a [react error 200](https://react.dev/errors/200?invariant=200).
Here we fix the issue by validating the returned `contanerElement` as an Element.

- [x ] I have read the [Contribution Guide](../packages/core/CONTRIBUTING.md) for this project.

### DIFF
--- a/packages/core/src/components/Dialog/Dialog.tsx
+++ b/packages/core/src/components/Dialog/Dialog.tsx
@@ -171,7 +171,7 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
   static positions = DialogPosition;
   static animationTypes = AnimationType;
   static defaultProps = {
-    // position: "top",
+    position: "top",
     modifiers: [] as Modifier<any>[],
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -171,7 +171,7 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
   static positions = DialogPosition;
   static animationTypes = AnimationType;
   static defaultProps = {
-    position: "top",
+    // position: "top",
```
YossiSaadi, line 174: is it on purpose? what's the reason if yes?

MosheZemah (approved):

packages/core/src/components/Dialog/Dialog.tsx
yair65: mistake, fixed

PR #2127, opened 2024-05-19
chore(Responsive list): migrate to typescript

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/113186840/pulses/4247428543


packages/core/src/components/ResponsiveList/ResponsiveList.tsx
```diff
@@ -1,37 +1,35 @@
...
-const ResponsiveList = forwardRef(
+const ResponsiveList: VibeComponent<ResponsiveListProps> = forwardRef<HTMLDivElement, ResponsiveListProps>(
   (
     {
-      id,
-      className,
-      rootClassName,
+      id = "",
+      className = "",
+      rootClassName = "",
```
talkor, line 16: These and the rest empty strings can just be undefined

packages/core/src/components/ResponsiveList/ResponsiveList.tsx
```diff
@@ -103,44 +101,4 @@ const ResponsiveList = forwardRef(
...
-
-ResponsiveList.menuButtonSizes = MenuButton.sizes;
-ResponsiveList.propTypes = {
-  id: PropTypes.string,
-  className: PropTypes.string,
-  menuButtonClassName: PropTypes.string,
-  menuWrapperClassName: PropTypes.string,
-  /**
-   These attributes will be passed to the MenuButton
-   */
```
talkor, line 115: Let's keep those JSDoc comments as they help documenting the props

packages/core/src/components/ResponsiveList/ResponsiveList.tsx
```diff
@@ -74,7 +72,7 @@ const ResponsiveList = forwardRef(
             {directChildren}
             {!!menuChildren.length && (
               <MenuButton
-                componentClassName={cx(styles.listMenuButton, menuButtonClassName)}
+                className={cx(styles.listMenuButton, menuButtonClassName)}
```
talkor, line 75: componentClassName was deprecated?

packages/core/src/components/ResponsiveList/ResponsiveList.tsx
```diff
@@ -58,7 +56,7 @@ const ResponsiveList = forwardRef(
     }, [children, index]);
 
     const hiddenChildren = useMemo(() => {
-      const childrenArray = React.Children.toArray(children);
+      const childrenArray = React.Children.toArray(children) as React.ReactElement[];
```
talkor, line 59: Can it be a single element?

talkor (approved):
Excellent 💪🏼 See some minor comments

packages/core/src/components/ResponsiveList/ResponsiveList.tsx
rivka-ungar: Seems so
https://github.com/mondaycom/vibe/blob/2ca65624abb07c88fde5e758a47c7e6c4e62b60f/packages/core/src/components/MenuButton/MenuButton.tsx#L30

packages/core/src/components/ResponsiveList/ResponsiveList.tsx
rivka-ungar: Even if it's a single element it is converted into an array of 1 item

PR #2138, opened 2024-05-23
fix(SplitButton storybook): - fix stories that use wrong prop

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/113186840/pulses/6495343163

### DIFF
--- a/packages/core/src/components/SplitButton/__stories__/splitButton.stories.js
+++ b/packages/core/src/components/SplitButton/__stories__/splitButton.stories.js
@@ -188,7 +188,7 @@ export const SplitButtonAsThePrimaryAction = {
   render: () => (
     <SplitButton
       children="Use template"
-      secondaryDialogPosition={DialogPosition.BOTTOM_START}
+      secondaryDialogPosition={SplitButton.secondaryDialogPositions.BOTTOM_START}
       secondaryDialogContent={
         <SplitButtonMenu id="split-menu">
           <MenuItem icon={Download} title="Import template" />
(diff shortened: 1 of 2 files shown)


packages/core/src/components/SplitButton/__stories__/splitButton.stories.js
```diff
@@ -187,7 +188,7 @@ export const SplitButtonAsThePrimaryAction = {
   render: () => (
     <SplitButton
       children="Use template"
-      secondaryPositions={SplitButton.secondaryPositions.BOTTOM_END}
+      secondaryDialogPosition={DialogPosition.BOTTOM_START}
```
YossiSaadi, line 191: You can use the secondaryDialogPosition from the SplitButton's static values. If I'm not wrong, we do not export the DialogPosition, so we want to show stories that users can just copy and use

packages/core/src/components/SplitButton/__stories__/splitButton.stories.js
```diff
@@ -211,7 +212,7 @@ export const SecondarySplitButton = {
       <SplitButton
         children="Export to CSV"
         kind={SplitButton.kinds.SECONDARY}
-        secondaryPositions={SplitButton.secondaryPositions.BOTTOM_MIDDLE}
+        secondaryDialogPosition={DialogPosition.BOTTOM_START}
```
YossiSaadi, line 215: Same

YossiSaadi (commented):
Awesome! Had a suggestion to make stories usable for users

packages/core/src/components/SplitButton/__stories__/splitButton.stories.js
rivka-ungar: The issue is that the secondaryDialogPositions from the static values is of type SplitButtonSecondaryContentPosition and the secondaryDialogPosition is of type DialogPosition

rivka-ungar: We resolved the issue by adding SplitButtonSecondaryContentPosition type to secondaryDialogPosition, and will remove the DialogPosition type in the next major release

PR #2141, opened 2024-05-26
fix(tooltip storybook) Tooltips are open outside of stories

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/113186840/pulses/6426278071


talkor (approved):
💯

PR #2147, opened 2024-05-29
chore: remove next components [BREAKING]

talkor:
https://monday.monday.com/boards/3532714909/pulses/6717338787


packages/core/src/components/EditableInput/EditableInput.module.scss
talkor, line 1: Was used internally for the old EditableHeading, it was an unstyled input, but for some reason was exported so I added it to the changelog anyway

packages/core/.storybook/preview.tsx
```diff
@@ -1,6 +1,5 @@
 import React from "react";
 import * as VibeComponents from "../src/components";
-import * as VibeComponentsNext from "../src/next";
```
shaharzil, line 3: Let's consider not removing it completely but leaving it as is and just removing the components imports inside

talkor: Brought it back

packages/core/package.json
```diff
@@ -29,9 +29,9 @@
     },
     "./storybookComponents": "./dist/storybook/index.js",
     "./next": {
-      "types": "./dist/esm/next/next.d.ts",
-      "import": "./dist/esm/src/next/next.js",
-      "default": "./dist/esm/src/next/next.js"
+      "types": "./dist/esm/components/next/next.d.ts",
```
talkor, line 32: There were two files which one of them was redundant, so now I pointed it to the other file (makes more sense)

packages/core/rollup.config.js
```diff
@@ -67,8 +67,7 @@ export default {
     index: path.join(SRC_PATH, "index.js"),
     icons: path.join(SRC_PATH, "components/Icon/Icons/index.ts"),
     interactionsTests: path.join(SRC_PATH, "tests/interactions-utils.ts"),
-    testIds: path.join(SRC_PATH, "tests/test-ids-utils.ts"),
-    next: path.join(SRC_PATH, "next/next.ts")
```
shaharzil, line 71: dont we also need to add this line back?

packages/core/src/storybook/decorators/withLiveEdit/components/LiveContent/LiveContent.tsx
```diff
@@ -6,9 +6,8 @@ import { LiveContentProps } from "./LiveContent.types";
 import styles from "./LiveContent.module.scss";
 import * as VibeComponents from "../../../../../components";
 import * as VibeIcons from "../../../../../components/Icon/Icons";
-import * as VibeComponentsNext from "../../../../../next";
```
shaharzil, line 9: dont we want to add next back?

packages/core/.storybook/preview.tsx
talkor: Moved to line 4

packages/core/rollup.config.js
talkor: Good catch, fixing

packages/core/src/storybook/decorators/withLiveEdit/components/LiveContent/LiveContent.tsx
talkor: good catch, fixing

shaharzil (approved):

PR #2153, opened 2024-06-02
feat: export component types

talkor:
https://monday.monday.com/boards/3532714909/pulses/6086045892


packages/core/src/__tests__/exports.test.ts
```diff
@@ -0,0 +1,12 @@
+import * as components from "../components/index";
+import * as nextComponents from "../next";
+
+describe("exports", () => {
```
talkor, line 4: Ran it on master and then again on the branch to verify that the tests has a correct baseline and all components are exported as expected

shaharzil (approved):

packages/core/src/components/BaseInput/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as BaseInput } from "./BaseInput";
```
YossiSaadi, line 1: This shouldn't be exported for consumers of the lib, does it mean now it does export it?

packages/core/src/next.ts
```diff
@@ -1,7 +1,3 @@
-export * from "./components/Heading/Heading";
-import Heading from "./components/Heading/Heading";
-export * from "./components/EditableHeading/EditableHeading";
-import EditableHeading from "./components/EditableHeading/EditableHeading";
-import Search from "./components/Search/Search";
-
-export { Heading, EditableHeading, Search };
+export * from "./components/Heading";
```
YossiSaadi, line 1: I think you've once said that this is not the fils which decides which components are exported, I think it is something that was added only for the playground

packages/core/src/components/index.ts
```diff
@@ -1,108 +1,67 @@
...
-export { default as ListItemAvatar } from "./ListItemAvatar/ListItemAvatar";
-export { default as ListTitle } from "./ListTitle/ListTitle";
-export { default as List } from "./List/List";
-export { default as VirtualizedList } from "./VirtualizedList/VirtualizedList";
-export { default as VirtualizedGrid } from "./VirtualizedGrid/VirtualizedGrid";
-export { default as TextWithHighlight } from "./TextWithHighlight/TextWithHighlight";
-export { default as Accordion } from "./Accordion/Accordion/Accordion";
-export { default as AccordionItem } from "./Accordion/AccordionItem/AccordionItem";
-export { default as Clickable } from "./Clickable/Clickable";
+export * from "./Accordion";
```
YossiSaadi, line 1: It looks awesome! Need to double check no unwanted exports occur now

YossiSaadi (commented):
I'm late to the party, but wanted to say it is awesome, especially how you implemented it

talkor:
> I'm late to the party, but wanted to say it is awesome, especially how you implemented it

Thanks! We do need to point to the generated declaration file created by rollup rather than creating it our own though, to make it work

packages/core/src/components/BaseInput/index.ts
talkor: It's only exporting the content of the folder, I haven't included it in the components' index file, so it's relevant only internally

PR #2157, opened 2024-06-02
chore(Tooltip): make showOnDialogEnter default to true

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6717263394?userId=59970142


packages/core/src/components/Tipseen/__stories__/Tipseen.mdx
```diff
@@ -96,6 +96,7 @@ In this case, the Tipseen position will be the right corner of the screen and wi
               modifiers={modifiers}
               position={Tipseen.positions.RIGHT}
               color={Tipseen.colors.INVERTED}
+              showDelay={100}
```
talkor, line 99: Didn't we decide to make it the default?

rivka-ungar: That was the hideDelay. The showDelay I added to the tipseens just for the storybook so they don't disappear after hover.

talkor: I thought the showDelay was added for that, is there any way to avoid it? it would be best to show a minimal code as possible for each story

rivka-ungar: Made it default

talkor (approved):

PR #2159, opened 2024-06-02
chore(Tooltip): remove paddingSize

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6717264040


talkor (approved):
Finally haha
Reminding to add this change to the changelog as well :)

packages/core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -4,7 +4,7 @@ import cx from "classnames";
 import React, { CSSProperties, isValidElement, PureComponent, ReactElement } from "react";
 import { Modifier } from "react-popper";
 import Dialog from "../Dialog/Dialog";
-import { AnimationType, BASE_SIZES_WITH_NONE, HideShowEvent, JustifyType } from "../../constants";
```
shaharzil, line 7: Maybe we also need to delete the declaration of "BASE_SIZES_WITH_NONE" from the constants?

rivka-ungar: You're right, deleted it

PR #2162, opened 2024-06-05
fix(Table): sync between body scroll and header scroll in virtualized state

YossiSaadi:
this is a temp workaround until the virtualized table component would get a big revamp. this workaround sets List width to be the full width and not the parent container's width

https://monday.monday.com/boards/3532714909/pulses/6513670146

### DIFF
--- a/packages/core/src/components/Table/Table/Table.tsx
+++ b/packages/core/src/components/Table/Table/Table.tsx
@@ -8,6 +8,7 @@ import { getTestId } from "../../../tests/test-ids-utils";
 import { ComponentDefaultTestId } from "../../../tests/constants";
 import { RowHeights, RowSizes } from "./TableConsts";
 import styles from "./Table.module.scss";
+import { TableProvider } from "../context/TableContext/TableContext";
 
 export type TableLoadingStateType = "long-text" | "medium-text" | "circle" | "rectangle";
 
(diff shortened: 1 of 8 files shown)


packages/core/src/components/Table/Table/Table.tsx
```diff
@@ -39,16 +40,6 @@ export interface ITableProps extends VibeComponentProps {
   withoutBorder?: boolean;
 }
 
-export interface ITableContext {
```
talkor, line 42: Haven't we considered it as breaking if it's taken from dist?

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
```diff
@@ -1,38 +1,67 @@
...
 import TableBody from "../TableBody/TableBody";
 import styles from "./TableVirtualizedBody.module.scss";
-import { ScrollDirection } from "react-window";
-import { TableContext } from "../Table/Table";
+import { FixedSizeList as List, ListChildComponentProps, ScrollDirection } from "react-window";
+import { useTable } from "../context/TableContext/TableContext";
 import { RowHeights } from "../Table/TableConsts";
+import AutoSizer from "react-virtualized-auto-sizer";
+
+type Rows = Array<Record<string, unknown> & { id: string }>;
```
talkor, line 10: Both should be exported no?

packages/core/src/components/Table/context/TableContext/TableContext.types.ts
```diff
@@ -0,0 +1,17 @@
+import { ITableProps } from "../../Table/Table";
+import React from "react";
+
+export interface ITableContext {
```
talkor, line 4: Are these supported to be exported as well? If so, they need to be registered in Table/index.ts

talkor (approved):
Awesome solution! 👑

packages/core/src/components/Table/Table/Table.tsx
YossiSaadi: I'm not sure tbh
I do remember we've talked about it, but I don't think we need to cover those kind of cases
we do not export the `TableContext` component anyway outside from Vibe

PR #2165, opened 2024-06-05
feat(Modal): unmount when hidden

talkor:
https://monday.monday.com/boards/3532714909/pulses/6718830547

### DIFF
--- a/packages/core/src/components/Modal/Modal.tsx
+++ b/packages/core/src/components/Modal/Modal.tsx
@@ -76,6 +76,7 @@ export interface ModalProps {
    * z-index attribute of the container
    */
   zIndex?: number;
+  unmountOnClose?: boolean;
 }
 
 const Modal: FC<ModalProps> & { width?: typeof ModalWidth } = ({
@@ -95,6 +96,7 @@ const Modal: FC<ModalProps> & { width?: typeof ModalWidth } = ({
   closeButtonAriaLabel = "Close",
   contentSpacing = false,
   zIndex = 10000,
+  unmountOnClose,
   "data-testid": dataTestId
 }) => {
   const childrenArray: ReactElement[] = useMemo(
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Modal/Modal.tsx
```diff
@@ -76,6 +76,10 @@ export interface ModalProps {
    * z-index attribute of the container
    */
   zIndex?: number;
+  /**
+   * If true, the modal will unmount when it's not shown
+   */
+  unmountOnClose?: boolean;
```
talkor, line 82: Opened a ticket to remove it next major

packages/core/src/components/Modal/useShowHideModal.ts
```diff
@@ -37,22 +38,28 @@ export default function useShowHideModal({
     keys: KEYS
   });
 
-  // show/hide and animate the modal
   useEffect(() => {
-    if (!instance) {
```
shaharzil, line 42: why did you removed this?

packages/core/src/components/Modal/useShowHideModal.ts
```diff
@@ -37,22 +38,28 @@ export default function useShowHideModal({
...
+      const animate = instance?.$el.childNodes[1].animate;
+      if (animate) {
+        instance?.$el.childNodes[1].animate?.(...getAnimationProps());
+      }
     } else {
+      const animate = instance?.$el.childNodes[1].animate;
       if (animate) {
         const animation = instance?.$el.childNodes[1]?.animate(...getAnimationProps(true));
-        animation?.finished.then(() => instance?.hide());
+        animation.finished.then(() => {
```
shaharzil, line 53: why did you removed the "?" after "animation"?

packages/core/src/components/Modal/useShowHideModal.ts
talkor: The check is unnecessary here, it's already checked inside the conditions. If it exists here, when the modal is not being rendered, it will not render at all

packages/core/src/components/Modal/useShowHideModal.ts
talkor: Missed it, fixing

shaharzil (approved):

PR #2167, opened 2024-06-09
feat(toggle): add small variant

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6646098988


packages/core/src/components/Toggle/__stories__/Toggle.stories.tsx
```diff
@@ -50,6 +50,22 @@ export const States: Story = {
   }
 };
 
+export const Size: Story = {
+  render: () => (
+    <MultipleStoryElementsWrapper className="monday-storybook-toggle_row">
+      <Toggle />
```
YossiSaadi, line 56: maybe add the medium one here just for documentation reasons

packages/core/src/components/Toggle/Toggle.tsx
YossiSaadi, line 1: I see that task also had a "remove off label" subtask, should it be part of a new PR?

packages/core/src/components/Toggle/MockToggle.tsx
```diff
@@ -22,15 +24,17 @@ export const MockToggle: FC<MockToggleProps> = ({
...
 }) => (
   <>
     {areLabelsHidden ? null : <ToggleText disabled={disabled}>{offOverrideText}</ToggleText>}
     <div
       className={cx(styles.toggle, className, {
         [cx(styles.selected, selectedClassName)]: checked,
         [styles.notSelected]: !checked,
-        [styles.disabled]: disabled
+        [styles.disabled]: disabled,
+        [styles.small]: size === "small"
```
YossiSaadi, line 37: I believe a better practice here can be to set a className with the size (whether it is a small/medium, e.g. `getStyle(styles, size)`), and afterwards change the css by only a css var.

e.g. instead of declaring width and height twice, declare the width and height to be a --toggle-width for example, with .small class to change the value of the css var

this can also help understand the calculation being done on the css

YossiSaadi: this is not a blocker by me, just for your consideration

YossiSaadi (approved):
Approved, I had a suggestion on something I believe can make the css clearer, not a blocker, for your consideration

packages/core/src/components/Toggle/Toggle.tsx
YossiSaadi: btw, it says that on the do's and don'ts, lol 😄
image

packages/core/src/components/Toggle/MockToggle.tsx
rivka-ungar: @YossiSaadi Good idea, should I declare --toggle-width, --toggle-height, in a global file or in this one?

packages/core/src/components/Toggle/Toggle.tsx
rivka-ungar: This prop already exists offOverrideText

packages/core/src/components/Toggle/MockToggle.tsx
YossiSaadi: it should be in the same module of MockToggle so different toggles can have different values for the vars
so for example
```css
.mockToggle {
  width: var(--toggle-width);
  &.small {
    --toggle-width: X1px;
    --toggle-height: Y1px;
  }
  &.medium {
    --toggle-width: X2px;
    --toggle-height: Y2px;
  }
}
```

packages/core/src/components/Toggle/Toggle.tsx
YossiSaadi: oh cool 💪🏼

packages/core/src/components/Toggle/MockToggle.tsx
rivka-ungar: Done

YossiSaadi (approved):
💯 !

YossiSaadi:
btw, notice to also accept the changes related to Toggle in Chromatic 🙏🏼

PR #2168, opened 2024-06-09
fix(useAfterFirstRender): hook is currently behaving the opposite than its purpose

YossiSaadi:
This PR also includes enabling hooks on playground and live edit

https://monday.monday.com/boards/3532714909/pulses/6799101876


packages/core/plop/hooks/index.js
```diff
@@ -37,7 +37,7 @@ module.exports = plop => {
       },
       {
         type: "append",
-        path: "src/hooks/index.js",
```
talkor, line 40: 🥇

packages/core/src/components/Counter/Counter.tsx
```diff
@@ -97,7 +97,7 @@ const Counter: React.FC<CounterProps> & {
 
   // Effects
   useEffect(() => {
-    if (!isAfterFirstRender.current) {
+    if (isAfterFirstRender.current) {
```
talkor, line 100: LOL

packages/core/src/hooks/useAfterFirstRender/__tests__/useAfterFirstRender.test.tsx
```diff
@@ -0,0 +1,35 @@
+import { renderHook } from "@testing-library/react-hooks";
```
talkor, line 1: Amazing

packages/core/src/hooks/useAfterFirstRender/__tests__/useAfterFirstRender.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+import useAfterFirstRender from "../";
+
+describe("useAfterFirstRender", () => {
+  it("should initially set isAfterFirstRender to false", () => {
+    const { result } = renderHook(() => useAfterFirstRender());
+    expect(result.current.current).toBeFalsy();
```
talkor, line 7: Why is it `current.current`?

packages/core/src/hooks/useAfterFirstRender/index.ts
```diff
@@ -1,11 +1,14 @@
...
 
 export default function useAfterFirstRender(): RefObject<boolean> {
-  const isAfterFirstRender = useRef(true);
+  const isAfterFirstRender = useRef(false);
+
   useEffect(() => {
-    window.requestAnimationFrame(() => {
-      isAfterFirstRender.current = false;
-    });
+    const timer = setTimeout(() => {
```
talkor, line 7: Why replace the RAF?

packages/core/src/hooks/useAfterFirstRender/__tests__/useAfterFirstRender.test.tsx
YossiSaadi: since renderHook's result always return a ref, so I had to get to that ref's value, and my hook returns a ref so the ref's value is another ref

packages/core/src/hooks/useAfterFirstRender/index.ts
YossiSaadi: I saw it is considered a better practice for this kind of cases
I'll give GPT the stage:

> Choose `setTimeout` with a delay of `0` over `requestAnimationFrame` - because it guarantees that the state update occurs after all rendering tasks are complete - regardless of the browser's frame rate or tab visibility, ensuring consistent execution timing across all scenarios

talkor: Ok but will it ensure that it will return true after React's first render?

YossiSaadi: I believe so. because when using `setTimeout` it would run the state update to run after all current tasks (including React's initial render?). 
I believe it ensures that the update occurs after React has completed its first render cycle by placing it at the end of the stack.
I also believe that this is the reason they initially went with a ref and not a state, to not force a re-render right after it.

I'm a believer

talkor (approved):

PR #2169, opened 2024-06-09
chore: remove deprecated props [BREAKING]

talkor:
https://monday.monday.com/boards/3532714909/pulses/6717338767


packages/core/docs/vibe-3-changelog.md
```diff
@@ -1,9 +1,148 @@
 # Vibe 3 Changelog
 
+## Components
+
+### AttentionBox
+
+- `componentClassName` -> `className` [codemod]
```
talkor, line 7: I thought it could be useful both for use to track of which can be changed via codemods and also let the users know that this change is covered by a codemod as well

packages/core/src/components/Button/Button.tsx
```diff
@@ -233,7 +226,6 @@ const Button: VibeComponent<ButtonProps, unknown> & {
       className,
       size,
       kind,
-      loading,
```
talkor, line 236: It's unused in the effect

packages/core/docs/vibe-3-changelog.md
```diff
@@ -1,15 +1,162 @@
...
+- `isDisabled` -> `disabled` [codemod]
+
+## Hooks
 
-- Requires codemod for when people used secondaryDialogPosition={SplitButton.secondaryBlaBla}.
+### useClickableProps
 
-- Once changing DialogPosition to be a const, instead of enum, remove DialogPosition double declaration from SplitButton's declaration
+- `dataTestId` -> `data-testid` [codemod]
```
YossiSaadi, line 162: was just thinking of something we would probably not be able to catch with codemod (?)
what if a user spread props to a component from an object?
hope we won't have much of that

packages/core/docs/vibe-3-changelog.md
YossiSaadi: Love all the structure change of this changelog
Maybe the headline should be more specific stating components **props**?

packages/core/docs/vibe-3-changelog.md
talkor: IMO the codemod should cover all the straightforward cases, for more complex cases you can't really handle it. The props values can come from very distant values/objects. TS will cover it anyway, the codemod is only meant to save the need for the users to make repetitive changes

PR #2172, opened 2024-06-17
feat(Table): add sticky column capabilities

YossiSaadi:
Works for both non and virtualized tables
https://monday.monday.com/boards/3532714909/pulses/6717612609

https://github.com/mondaycom/vibe/assets/30905362/081eefb3-028f-48c1-8a0c-ef88b006487a

### DIFF
--- a/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
+++ b/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
@@ -308,7 +308,7 @@ export const scrollTableColumns = [
 ];
 
 export const virtualizedScrollTableData: TableVirtualizedRow[] = [...new Array(5000)].map((_, index) => ({
-  id: `${index}`,
+  id: index.toString(),
   num: index,
   text: `This is line number ${index}`
 }));
--- a/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
+++ b/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
@@ -2,6 +2,8 @@ import React from "react";
 import Avatar from "../../../Avatar/Avatar";
 import { Calendar, Doc, Status } from "../../../Icon/Icons";
 import { LabelColor } from "../../../Label/LabelConstants";
+import { TableVirtualizedRow } from "../../TableVirtualizedBody/TableVirtualizedBody";
+import { ITableColumn } from "../Table";
 
 export const doAndDontIconsRuleColumns = [
   {
@@ -305,13 +307,13 @@ export const scrollTableColumns = [
   }
 ];
 
-export const virtualizedScrollTableData = [...new Array(5000)].map((_, index) => ({
-  id: index,
+export const virtualizedScrollTableData: TableVirtualizedRow[] = [...new Array(5000)].map((_, index) => ({
+  id: `${index}`,
   num: index,
   text: `This is line number ${index}`
 }));
 
-export const virtualizedScrollTableColumns = [
+export const virtualizedScrollTableColumns: ITableColumn[] = [
   {
     id: "num",
     title: "#",
--- a/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
+++ b/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
@@ -5,6 +5,8 @@
   top: 0;
   z-index: 2;
   background-color: inherit;
+  min-width: 100%;
+  width: fit-content;
 
   > * {
     background-color: inherit;
--- a/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
+++ b/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
@@ -3,7 +3,7 @@
   grid-template-columns: var(--table-grid-template-columns);
   position: sticky;
   top: 0;
-  z-index: 1;
+  z-index: 2;
   background-color: inherit;
 
   > * {
--- a/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
+++ b/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
@@ -11,4 +11,14 @@
   > * {
     background-color: inherit;
   }
+
+  &.virtualized {
+    overflow: auto;
+    width: auto;
+    scrollbar-width: none;
+
+    &::-webkit-scrollbar {
+      display: none;
+    }
+  }
 }
--- a/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
+++ b/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
@@ -8,7 +8,7 @@
   min-width: 100%;
   width: fit-content;
 
-  > * {
+  > [role="columnheader"] {
     background-color: inherit;
   }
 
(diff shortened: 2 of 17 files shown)


packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
```diff
@@ -305,13 +307,13 @@ export const scrollTableColumns = [
   }
 ];
 
-export const virtualizedScrollTableData = [...new Array(5000)].map((_, index) => ({
-  id: index,
+export const virtualizedScrollTableData: TableVirtualizedRow[] = [...new Array(5000)].map((_, index) => ({
+  id: `${index}`,
```
talkor, line 311: toString?

packages/core/src/components/Table/TableHeader/TableHeader.module.scss
```diff
@@ -3,10 +3,22 @@
   grid-template-columns: var(--table-grid-template-columns);
   position: sticky;
   top: 0;
-  z-index: 1;
+  z-index: 2;
   background-color: inherit;
+  min-width: 100%;
```
talkor, line 8: What if the table is smaller than full width? Would it stretch?

packages/core/src/components/Table/TableHeader/TableHeader.module.scss
```diff
@@ -3,10 +3,22 @@
...
   > * {
     background-color: inherit;
   }
+
+  &.virtualized {
+    overflow: auto;
+    width: auto;
+    scrollbar-width: none;
+
+    &::-webkit-scrollbar {
```
talkor, line 20: We don't want?

packages/core/src/components/Table/TableHeaderCell/TableHeaderCell.tsx
```diff
@@ -1,3 +1,4 @@
+// noinspection ES6UnusedImports
```
talkor, line 1: Why?

packages/core/src/components/Table/TableRow/TableRow.module.scss
```diff
@@ -4,11 +4,17 @@
...
   grid-template-columns: var(--table-grid-template-columns);
+  min-width: 100%;
+  width: fit-content;
 
   &[aria-selected="true"] > * {
     background-color: var(--primary-selected-color);
   }
 
+  > * {
+    background-color: var(--primary-background-color);
```
talkor, line 15: Isn't it dangerous? What about components used in rows?

packages/core/src/components/Table/index.ts
```diff
@@ -8,7 +8,6 @@ export { default as TableBody, ITableBodyProps as TableBodyProps } from "./Table
 export {
   default as TableVirtualizedBody,
   ITableVirtualizedBodyProps as TableVirtualizedBodyProps,
-  TableVirtualizedRows,
```
talkor, line 11: Is it breaking?

packages/core/src/components/Table/context/TableContext/TableContext.tsx
```diff
@@ -1,17 +1,63 @@
...
 export const TableProvider = ({ value, children }: ITableProviderProps) => {
-  const [rowWidth, setRowWidth] = useState<number>(0);
-  const contextValue = useMemo(
+  const [isVirtualized, setIsVirtualized] = useState<boolean>(false);
+  const markTableAsVirtualized = useCallback(() => {
+    setIsVirtualized(true);
+  }, []);
+
+  const [scrollLeft, setScrollLeft] = useState<number>(0);
+  const headRef = useRef<HTMLDivElement>(null);
```
talkor, line 13: What's this? What's the usage of the head here?

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
```diff
@@ -19,66 +19,72 @@ export interface ITableVirtualizedBodyProps extends VibeComponentProps {
...
-  const itemRenderer = useCallback<ComponentType<ListChildComponentProps<TableVirtualizedRow>>>(
-    ({ index, style }) => {
-      const currentItem = items[index];
-      const element = rowRenderer(currentItem);
-      return React.cloneElement(element, { style, key: index });
-    },
-    [items, rowRenderer]
-  );
+    useEffect(() => {
+      markTableAsVirtualized();
```
talkor, line 30: Not sure I got this, can you elaborate?

talkor (commented):
Amazing! Very neat implementation! Had some questions but looks great

packages/core/src/components/Table/TableHeaderCell/TableHeaderCell.tsx
YossiSaadi: have no idea where it came from 😨

packages/core/src/components/Table/TableHeader/TableHeader.module.scss
YossiSaadi: on virtualized we're now synchronizing scrolls (at least temporarily?), so we don't want the header scrollbar to be shown, the user should feel as if the header moves as one unit with the body

packages/core/src/components/Table/TableHeader/TableHeader.module.scss
YossiSaadi: it would stretch to fit-content, but at minimum would fit to 100%

YossiSaadi: if the table is in a smaller container it would behave the same, just relative to the smaller container

packages/core/src/components/Table/index.ts
YossiSaadi: yes and no
we only just pushed that type recently, and I'm not sure we're exporting the types correctly right now, aren't we?

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
YossiSaadi: wanted to use context's power to prevent user from passing another prop to table
so that's the way I went with, to let other components within table they should behave in a virtualized variant

packages/core/src/components/Table/context/TableContext/TableContext.tsx
YossiSaadi: to scroll between the header and body horizontal scrolls, as both are always visually appear but behind the container's limits

packages/core/src/components/Table/TableHeader/TableHeader.module.scss
talkor: Oh it's not the header? the table will fit content?

YossiSaadi: both the header row and body row

packages/core/src/components/Table/TableRow/TableRow.module.scss
YossiSaadi: It does not affect deeper nested elements beyond the first level, because of the `>`

YossiSaadi: and the first level should always be the table cell element

YossiSaadi: note for myself: target table cell only anyway, for performance concerns

talkor (approved):

PR #2173, opened 2024-06-19
chore: downgrade react-select version to match that of react-windowed-select

YossiSaadi:
we're hoping this would affect the bundle size

things to consider:
- losing patches from 3.1.1 - https://github.com/JedWatson/react-select/blob/master/packages/react-select/CHANGELOG.md#311
- losing patches and featurs from 3.2.0 - https://github.com/JedWatson/react-select/blob/master/packages/react-select/CHANGELOG.md#320

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -114,11 +114,11 @@
     "react-inlinesvg": "^3.0.1",
     "react-is": "^16.9.0",
     "react-popper": "^2.3.0",
-    "react-select": "^3.2.0",
+    "react-select": "3.1.0",
     "react-transition-group": "^4.4.5",
     "react-virtualized-auto-sizer": "^1.0.7",
     "react-window": "^1.8.7",
-    "react-windowed-select": "^2.0.4",
+    "react-windowed-select": "2.0.5",
     "style-inject": "^0.3.0",
     "vibe-storybook-components": "0.19.0"
   },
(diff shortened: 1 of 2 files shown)


YossiSaadi:
it seem like were dependent on fixes from latest 3.x version
We need to set resolutions instead, to force react-windowed-select to use react-select 3.2.0, though it might also create trouble

packages/core/package.json
```diff
@@ -114,11 +114,11 @@
     "react-inlinesvg": "^3.0.1",
     "react-is": "^16.9.0",
     "react-popper": "^2.3.0",
-    "react-select": "^3.2.0",
+    "react-select": "3.1.0",
```
talkor, line 117: Any missing features here? As downgrading can mean removing new features

packages/core/package.json
```diff
@@ -114,11 +114,11 @@
...
     "react-transition-group": "^4.4.5",
     "react-virtualized-auto-sizer": "^1.0.7",
     "react-window": "^1.8.7",
-    "react-windowed-select": "^2.0.4",
+    "react-windowed-select": "2.0.5",
```
talkor, line 121: Same

YossiSaadi:
#2184 upgraded to 4.3.1 to have both on same version

PR #2174, opened 2024-06-19
 fix(dropdown): Dropdown multi size "small" style and behavior is broken

rivka-ungar:
https://monday.monday.com/boards/3532714909/pulses/6277634529

### DIFF
--- a/packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
+++ b/packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
@@ -16,11 +16,11 @@
 }
 
 .multiselect-chip-single-line-large {
-  --chips-margin: 9px 2px !important;
+  --chips-margin: 9px 2px;
 }
 
 .multiselect-chip-single-line-small {
-  --chips-margin: 1px 2px !important;
+  --chips-margin: 1px 2px;
 }
 
 .multiselect-chip-disabled {
--- a/packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
+++ b/packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
@@ -11,8 +11,16 @@
   --chips-margin: 2px;
 }
 
-.multiselect-chip-single-line {
+.multiselect-chip-single-line-medium {
   --chips-margin: 5px 2px;
+}
+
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
```diff
@@ -11,10 +11,18 @@
   --chips-margin: 2px;
 }
 
-.multiselect-chip-single-line {
+.multiselect-chip-single-line-medium {
   --chips-margin: 5px 2px;
 }
 
+.multiselect-chip-single-line-large {
+  --chips-margin: 9px 2px !important;
```
talkor, line 19: Is the important necessary? Let's try to avoid it since it's catch all and can cause specificity issues

packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
```diff
@@ -11,10 +11,18 @@
...
+}
+
+.multiselect-chip-single-line-small {
+  --chips-margin: 1px 2px !important;
```
talkor, line 23: Same about important

packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
```diff
@@ -11,10 +11,18 @@
   --chips-margin: 2px;
 }
 
-.multiselect-chip-single-line {
+.multiselect-chip-single-line-medium {
```
talkor, line 14: What about other sizes? I see that it supports also XS, XXS etc
Screenshot 2024-06-25 at 14 45 33

packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
rivka-ungar: No, I forgot to remove it, added it just while I was testing, thanks!

talkor (approved):

packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.module.scss
rivka-ungar: XS & XXS don't work, we should remove them in the next major

PR #2176, opened 2024-06-20
chore(codemods): init repo + codemod infra + codemod: change next import to regular import

YossiSaadi:
- initiating codemods package in the monorepo
- initiating codemod writing infra (right now mainly for import handling)
- add a codemod
	- change next import to regular import

### DIFF
--- a/packages/codemods/.eslintrc.json
+++ b/packages/codemods/.eslintrc.json
@@ -1,0 +1,27 @@
+{
+  "root": true,
+  "parser": "@typescript-eslint/parser",
+  "extends": [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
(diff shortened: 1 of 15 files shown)


shlomitc (commented):
As a general comment - I suggest providing a single codemod for vibe 2 =>3 migration instead of multiple ones.
It should be easier to run instead of maintaining and applying multiple ones

YossiSaadi:
> As a general comment - I suggest providing a single codemod for vibe 2 =>3 migration instead of multiple ones.
> It should be easier to run instead of maintaining and applying multiple ones

Yep, we're gonna have a single one at the end, but it's considered a best practice to separate everything to different ones, and supply the option to run all

Right now I'm preparing a single codemod for each "type" (prop change, prop removal, import handling, etc.) to have everyone in the team create similar codemods easily based on those

PR #2180, opened 2024-06-23
feat(TextField): add a controlled variant

talkor:
https://monday.monday.com/boards/3532714909/pulses/6887647837

### DIFF
--- a/packages/core/src/components/TextField/TextField.tsx
+++ b/packages/core/src/components/TextField/TextField.tsx
@@ -1,5 +1,5 @@
 import cx from "classnames";
-import React, { forwardRef, useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, { ChangeEvent, forwardRef, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import useDebounceEvent from "../../hooks/useDebounceEvent";
 import Icon from "../Icon/Icon";
 import { backwardCompatibilityForProperties } from "../../helpers/backwardCompatibilityForProperties";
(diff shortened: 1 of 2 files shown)


packages/core/src/components/TextField/TextField.tsx
```diff
@@ -171,13 +176,25 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
       [onChange, isRequiredAndEmpty]
     );
 
-    const { inputValue, onEventChanged, clearValue } = useDebounceEvent({
+    const {
+      inputValue: uncontrolledInput,
```
YossiSaadi, line 180: why?

YossiSaadi (commented):
I can't understand why if something is controlled it shouldn't get the logic of debounce, etc.

packages/core/src/components/TextField/TextField.tsx
talkor: The thing is that the debounce has a state, this is the root cause of the issue. I can add debounce but I think that if the user wants to control the input they should have the full flexibility

packages/core/src/components/TextField/TextField.tsx
```diff
@@ -171,13 +176,25 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
...
       trim
     });
 
+    const inputValue = useMemo(() => {
+      return controlled ? value : uncontrolledInput;
+    }, [controlled, value, uncontrolledInput]);
+
+    function handleChange(event: ChangeEvent<Partial<HTMLInputElement>>) {
+      controlled ? onChangeCallback(event.target.value) : onEventChanged(event);
+    }
```
YossiSaadi, line 196:
```suggestion
const handleChange = useCallback<ChangeEventHandler>(event => {
      // ...
    }, []);
```

YossiSaadi (approved):
We need, when instructing people to use this new `controlled` to let them know they lose some of the built in functionality as debounce

PR #2181, opened 2024-06-23
chore(icon): Remove clickable prop and set icons as decorative by default

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6887644148

### DIFF
--- a/packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
+++ b/packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
@@ -69,12 +69,11 @@ exports[`Dropdown renders correctly snapshot driver should open menu on click if
             class="dropdown-indicator css-1pxbrck-indicatorContainer"
           >
             <svg
-              aria-hidden="false"
-              class="icon clickable"
+              aria-hidden="true"
+              class="icon"
               data-testid="icon"
               fill="currentColor"
               height="20px"
-              role="button"
               tabindex="-1"
               viewBox="0 0 20 20"
               width="20px"
@@ -255,12 +254,11 @@ exports[`Dropdown renders correctly snapshot driver should open menu on focus if
             class="dropdown-indicator css-slqvkp-indicatorContainer"
           >
             <svg
-              aria-hidden="false"
-              class="icon clickable"
+              aria-hidden="true"
+              class="icon"
               data-testid="icon"
               fill="currentColor"
               height="20px"
-              role="button"
               tabindex="-1"
               viewBox="0 0 20 20"
               width="20px"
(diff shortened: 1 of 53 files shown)


packages/core/src/components/Icon/__stories__/Icon.stories.tsx
```diff
@@ -31,13 +31,13 @@ export const Overview = {
 };
 
 export const VibeIcon = {
-  render: () => <Icon iconType={Icon.type.SVG} icon={Bolt} iconLabel="my bolt svg icon" iconSize={16} />,
```
talkor, line 34: aria-label shouldn't be added for non-interactive elements, right? I wonder why has it been added here in the first place

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -251,14 +248,7 @@ const MenuItem: VibeComponent<MenuItemProps | MenuItemTitleComponentProps> & {
         </div>
       ) : (
         <div className={styles.subMenuIconWrapper}>
-          <Icon
-            clickable={false}
-            icon={DropdownChevronRight}
-            iconLabel={iconLabel}
```
talkor, line 257: Though it's not clickable, it has a functionality, it's the arrow for the submenu, I think it should remain since it will imply that it opens the submenu. So I think that the aria-label should remain nevertheless

packages/core/src/components/TextField/__tests__/TextField.test.js
```diff
@@ -137,7 +137,7 @@ describe("TextField Tests", () => {
       );
     });
 
-    const icon = queryByLabelText(iconNames.primary);
+    const icon = queryByTestId("icon");
```
talkor, line 140: Is has a const somewhere

packages/core/src/components/MultiStepIndicator/components/StepIndicator/StepIndicator.tsx
```diff
@@ -41,10 +41,8 @@ const StepCircleDisplay: React.FC<StepCircleDisplayProps> = ({
     <Icon
       icon={fulfilledStepIcon}
       className={classNames(styles.numberContainerTextCheckIcon)}
-      iconLabel={StepStatus.FULFILLED}
```
talkor, line 44: This should remain

talkor (changes requested):
We need to keep the aria-label 🙆🏼‍♂️

packages/core/src/components/TextField/TextField.tsx
```diff
@@ -118,7 +116,6 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
       inputAriaLabel,
       searchResultsContainerId = "",
       activeDescendant = "",
-      iconsNames = EMPTY_OBJECT,
```
talkor, line 121: Is this a prop? What is this for?

talkor: If it is and is redundant it should be also documented, and make sure it doesn't have UX implications

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -69,12 +69,11 @@ exports[`Dropdown renders correctly snapshot driver should open menu on click if
             class="dropdown-indicator css-1pxbrck-indicatorContainer"
           >
             <svg
-              aria-hidden="false"
-              class="icon clickable"
+              aria-hidden="true"
```
talkor, line 72: Are we sure about this?

PR #2184, opened 2024-06-24
feat(react-select): use react-select version 4.3.1

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6864028773

### Why
We now have `react-windowed-select`, which, for some reason, does not use `react-select` as peer dep, but as a dep.
So this causes two versions of `react-select` at our bundle, as we cannot resolve both our dependency and `react-windowed-select`'s version to same version.

**This should be reviewed with caution**

The only breaking behavior on `react-select` 4 in compare to 3 is the `isMulti` behavior.
It used to return `null` when no options were selected. now it returns always array (empty array in case no options were selected).
So I had, to not cause breaking changes, mimic the current behavior with the new `react-select`'s api for `isMulti`'s `onChange`.

### DIFF
--- a/packages/core/src/components/Dropdown/Dropdown.tsx
+++ b/packages/core/src/components/Dropdown/Dropdown.tsx
@@ -260,7 +260,7 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
           : selectedOptions;
 
         if (customOnChange) {
-          customOnChange(newSelectedOptions, e);
+          customOnChange(newSelectedOptions.length > 0 ? newSelectedOptions : null, e);
         }
         setSelected(newSelectedOptions);
       };
@@ -289,7 +289,8 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
     );
     const onChange = (option: DropdownOption | DropdownOption[], meta: ActionMeta<DropdownOption>) => {
       if (customOnChange) {
-        customOnChange(option, meta);
+        const newValue = multi ? (option.length > 0 ? option : null) : option;
+        customOnChange(newValue, meta);
       }
 
       switch (meta.action) {
--- a/packages/core/src/components/Dropdown/Dropdown.tsx
+++ b/packages/core/src/components/Dropdown/Dropdown.tsx
@@ -260,7 +260,7 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
           : selectedOptions;
 
         if (customOnChange) {
-          customOnChange(newSelectedOptions.length > 0 ? newSelectedOptions : null, e);
+          customOnChange(newSelectedOptions, e);
         }
         setSelected(newSelectedOptions);
       };
--- a/packages/core/src/components/Dropdown/__tests__/Dropdown.test.js
+++ b/packages/core/src/components/Dropdown/__tests__/Dropdown.test.js
@@ -150,7 +150,8 @@ describe("Dropdown", () => {
 
       expect(onChange).toBeCalledTimes(2);
       expect(onChange).toHaveBeenLastCalledWith(null, {
-        action: "clear"
+        action: "clear",
+        removedValues: [mockOptions[0]]
       });
       expect(onClear).toBeCalled();
       expect(component.chips.values).toEqual([]);
--- a/packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
+++ b/packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
@@ -8,7 +8,7 @@ exports[`Dropdown renders correctly snapshot driver should open menu on click if
   >
     <span
       aria-live="polite"
-      class="css-1laao21-a11yText"
+      class="css-1f43avz-a11yText-A11yText"
     >
       <span
         id="aria-selection-event"
@@ -194,7 +194,7 @@ exports[`Dropdown renders correctly snapshot driver should open menu on focus if
   >
     <span
       aria-live="polite"
-      class="css-1laao21-a11yText"
+      class="css-1f43avz-a11yText-A11yText"
     >
       <span
         id="aria-selection-event"
@@ -762,7 +762,7 @@ exports[`Dropdown renders correctly snapshot driver should use async if set 1`] 
   >
     <span
       aria-live="polite"
-      class="css-1laao21-a11yText"
+      class="css-1f43avz-a11yText-A11yText"
     >
       <span
         id="aria-selection-event"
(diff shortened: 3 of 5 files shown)


packages/core/src/components/Dropdown/__tests__/Dropdown.test.js
```diff
@@ -150,7 +150,8 @@ describe("Dropdown", () => {
 
       expect(onChange).toBeCalledTimes(2);
       expect(onChange).toHaveBeenLastCalledWith(null, {
-        action: "clear"
+        action: "clear",
+        removedValues: [mockOptions[0]]
```
YossiSaadi, line 154: `onChange` now returns this in addition for when removing values in `isMulti` state

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -289,7 +289,8 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
     );
     const onChange = (option: DropdownOption | DropdownOption[], meta: ActionMeta<DropdownOption>) => {
       if (customOnChange) {
-        customOnChange(option, meta);
+        const newValue = multi ? (option.length > 0 ? option : null) : option;
```
YossiSaadi, line 292: if multi and array is empty return null, to mimic current behavior, and ignore the behavior of the new react-select's api for isMulti

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -260,7 +260,7 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
           : selectedOptions;
 
         if (customOnChange) {
-          customOnChange(newSelectedOptions, e);
+          customOnChange(newSelectedOptions.length > 0 ? newSelectedOptions : null, e);
```
YossiSaadi, line 263: I'm not 100% this is needed here
wdyt?

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -7,18 +7,18 @@ exports[`Dropdown renders correctly snapshot driver should open menu on click if
     id="dropdown-menu-id"
   >
     <span
+      aria-atomic="false"
       aria-live="polite"
-      class="css-1laao21-a11yText"
+      aria-relevant="additions text"
+      class="css-1f43avz-a11yText-A11yText"
     >
```
YossiSaadi, line 14: all this new aria-live behavior is intended by react-select 4.0.0, does not affect us, only adds a11y

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -193,18 +193,18 @@ exports[`Dropdown renders correctly snapshot driver should open menu on focus if
...
-      >
-         
-      </span>
+        id="aria-selection"
+      />
       <span
         id="aria-context"
       >
-           0 results available.   Selected:  is focused ,type to refine list, press Down to open the menu, 
+            Selected:  is focused ,type to refine list, press Down to open the menu,
```
YossiSaadi, line 207: It seem like it was a bug at react select. no reason that 0 results available would be read to user here, as the input does not filter the results at all. it should only appear if the input filter does not match any option. in here all the options are available

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -761,18 +797,18 @@ exports[`Dropdown renders correctly snapshot driver should use async if set 1`]
...
```
YossiSaadi, line 811: Same. seem like it was a bug. please verify me

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -820,13 +856,13 @@ exports[`Dropdown renders correctly snapshot driver should use async if set 1`]
...
-              class="css-imwzes-Component"
+              class="css-1yvy2vo-LoadingDot"
             />
             <span
-              class="css-1ue98e-Component"
+              class="css-zoievk-LoadingDot"
             />
             <span
-              class="css-fwsb3e-Component"
+              class="css-x748d8-LoadingDot"
```
YossiSaadi, line 865: module name changed in react-select

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -957,7 +993,7 @@ exports[`Dropdown renders correctly snapshot driver should use virtualization if
     >
       <div
         class=""
-        style="position: relative; height: 300px; overflow: auto; will-change: transform; direction: ltr; overflow-y: auto; box-sizing: border-box;"
+        style="position: relative; height: 300px; width: 100%; overflow: auto; will-change: transform; direction: ltr; overflow-y: auto; box-sizing: border-box;"
```
YossiSaadi, line 996: not sure why they've added it, but it doesn't affect the ui it seems

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -1329,7 +1377,7 @@ exports[`Dropdown renders correctly when readOnly 1`] = `
         <input
           aria-autocomplete="list"
           aria-label="Readonly   Selected: "
-          className="css-62g3xt-dummyInput"
+          className="css-wmatm6-dummyInput-DummyInput"
```
YossiSaadi, line 1380: module name changed (?)

YossiSaadi:
@shaharzil @rivka-ungar @talkor need your cooperation on that one to verify me
the tests were not comprehensive enough to put 100% trust on it

packages/core/src/components/Dropdown/Dropdown.tsx
talkor: Can you elaborate on the change?

YossiSaadi: I'm not sure newSelectedOptions could have been null at any scenario.
And if it cannot be null, it means our API is not consistent today, and can sometimes return null and sometimes empty array (?).

YossiSaadi: So this change might be redundant? Or am I misunderstanding something?

talkor: If it's not mandatory for the migration I wouldn't make changes here

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
talkor: It's not breaking, right?

YossiSaadi: it is only for the screen reader, and it is a bug fix, so don't think so

PR #2185, opened 2024-06-25
chore(tooltip): remove justify prop

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6890236950

### DIFF
--- a/packages/core/docs/vibe-3-changelog.md
+++ b/packages/core/docs/vibe-3-changelog.md
@@ -12,6 +12,7 @@
 ## Tipseen
 - Change showDelay default to 100
 - Remove justify prop
+- Remove static prop justifyTypes
 
 ## Dialog
 - Remove JustifyType 
--- a/packages/core/docs/vibe-3-changelog.md
+++ b/packages/core/docs/vibe-3-changelog.md
@@ -7,12 +7,9 @@
 - Change hideDelay default to be 100
 - Change position prop to accept string instead of static prop
 - Change position prop to only accept "top, right, bottom, left"
-<<<<<<< Breaking-Tooltip---remove-justify-prop-6890236950
 - Remove justify prop
-=======
 - Remove arrowPosition prop
 - Remove TooltipArrowPosition enum
->>>>>>> tooltip
 
 ## Tipseen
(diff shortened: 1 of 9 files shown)


packages/core/docs/vibe-3-changelog.md
```diff
@@ -7,6 +7,7 @@
 - Change hideDelay default to be 100
 - Change position prop to accept string instead of static prop
 - Change position prop to only accept "top, right, bottom, left"
+- Remove justify prop
 
 ## Tipseen
```
talkor, line 12: Need to document it also for the Tipseen

packages/core/src/components/Tooltip/__tests__/Tooltip.snapshot.test.tsx
```diff
@@ -52,7 +52,7 @@ describe("Tooltip renders correctly", () => {
   it("with justify", () => {
```
talkor, line 52: Needed?

packages/core/src/constants/dialog.ts
```diff
@@ -1,9 +1,3 @@
-export enum JustifyType {
```
talkor, line 1: Let's document this too

packages/core/src/components/Tooltip/__tests__/Tooltip.snapshot.test.tsx
rivka-ungar: Nope

packages/core/src/components/Tipseen/Tipseen.tsx
```diff
@@ -210,7 +206,6 @@ const Tipseen: VibeComponent<TipseenProps> & {
 export default withStaticProps(Tipseen, {
   closeButtonThemes: TipseenCloseButtonTheme,
   animationTypes: AnimationType,
-  justifyTypes: JustifyType,
```
talkor, line 213: I think we should document this one as well, we might not add this to the final migration guide since it can be covered with a codemod but I think we better keep it documented for us

talkor (approved):
Cool! Just add the `justifyTypes` to changelog please

PR #2186, opened 2024-06-25
chore(tooltip): remove arrowPosition

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6890238631

### DIFF
--- a/packages/core/src/components/Tooltip/TooltipConstants.ts
+++ b/packages/core/src/components/Tooltip/TooltipConstants.ts
@@ -1,7 +1,1 @@
-export enum TooltipArrowPosition {
-  BEGIN = "begin",
-  CENTER = "center",
-  END = "end"
-}
-
 export type TooltipTheme = "dark" | "primary";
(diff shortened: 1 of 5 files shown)


packages/core/src/components/Tooltip/TooltipConstants.ts
```diff
@@ -1,7 +1 @@
-export enum TooltipArrowPosition {
```
talkor, line 1: Need to document this removal as it's exported

packages/core/src/components/Tooltip/__tests__/Tooltip.snapshot.test.tsx
```diff
@@ -19,7 +19,7 @@ describe("Tooltip renders correctly", () => {
   it("with end arrowPosition", () => {
```
talkor, line 19: Do we need this test at all?

rivka-ungar: Nope, removed

talkor (approved):

PR #2189, opened 2024-06-25
feat(TextArea): be able to disable resize

talkor:
https://monday.monday.com/boards/3532714909/pulses/6899262204

### DIFF
--- a/packages/core/src/components/TextArea/TextArea.module.scss
+++ b/packages/core/src/components/TextArea/TextArea.module.scss
@@ -50,6 +50,10 @@
     &.large {
       @include vibe-text(text1, normal);
     }
+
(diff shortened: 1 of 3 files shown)


packages/core/src/components/TextArea/TextArea.module.scss
```diff
@@ -50,6 +50,10 @@
     &.large {
       @include vibe-text(text1, normal);
     }
+
+    &:not(.resize) {
+      resize: none;
```
YossiSaadi, line 55: Textarea is a very weird element. Why this kind of behavior is on css? 🤔

YossiSaadi (approved):

PR #2193, opened 2024-06-27
fix(ToastLink): hover color should inherit from link color

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6891175104


YossiSaadi (approved):
👌🏻

PR #2194, opened 2024-06-27
chore(Box): make scrollable default true

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6887644042


talkor (commented):
@rivka-ungar please add to changelog 🙏🏼

talkor (approved):

PR #2196, opened 2024-06-30
remove old label animation

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6798811711


packages/core/docs/vibe-3-changelog.md
```diff
@@ -70,6 +70,7 @@
 ### Label
 
 - `wrapperClassName` -> `className` [codemod]
+- Removed old animation
```
talkor, line 73: I'd rephrase "old" as "slide" or "rotate", to make it more clear
Plus I think we should specify the `isAnimationDisabled` prop that was removed

packages/core/src/components/Label/Label.tsx
talkor, line 1: I bet there's some redundant CSS also 😅

packages/core/src/styles/_keyframes.scss
```diff
@@ -8,7 +8,6 @@
 @import "keyframes/pop/pop-in-elastic";
 @import "keyframes/pop/pop-in-elastic-bold";
 @import "keyframes/pop/pop-out-from-trigger";
-@import "keyframes/spin/spin-in-emphasized";
```
talkor, line 11: We need to think about this, since it's a breaking change in `style`

talkor (approved):

PR #2203, opened 2024-07-02
feat(AccordionItem): add option for onClick instead of onClickCapture

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6947522708


packages/core/src/components/ExpandCollapse/ExpandCollapse.tsx
```diff
@@ -62,7 +63,8 @@ const ExpandCollapse: FC<ExpandCollapseProps> = forwardRef(
       headerClassName,
       contentClassName,
       componentClassName,
-      "data-testid": dataTestId
+      "data-testid": dataTestId,
+      captureOnClick
```
YossiSaadi, line 67: true by default

PR #2204, opened 2024-07-03
chore(chromatic): update chromatic cli to latest version

YossiSaadi:
At the end, the publish is done through the GitHub Actions action of Chromatic, so we are, currently, using the latest - it is just to keep up to date also the local dep for running Chromatic locally
Found it out during Chromatic's outage when I've tried to run build through local and got a warning from Chromatic regarding this

https://monday.monday.com/boards/3532714909/pulses/6957006224


talkor (approved):
Awesome!

PR #2214, opened 2024-07-07
chore(AttentionBox): codemod for migration of AttentionBox from core - v2 to v3

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6958656022

This PR also includes:
- consts: Vibe's imports (core and core/next)
- utils:
	- wrapper for each transform to supply it with `j` and `root`, and always make sure `root.toSource()` is called at the end (inspired by [Vue-codemod](https://github.com/vuejs/vue-codemod/blob/main/src/wrapAstTransformation.ts))
	- extract actual name from import (e.g. supply it with `AttentionBox` string and find both `import {AttentionBox}` and `import {AttentionBox as SomethingElse}`) to later find with it all the component element usages
	- find all jsx appearances of a specific tag (e.g. find all <AttentionBox jsx usages in a file)
	- get all import declerations of vibe core and vibe core/next
	- update prop name for a jsx element usage

### DIFF
--- a/packages/codemod/src/utils/import-utils.ts
+++ b/packages/codemod/src/utils/import-utils.ts
@@ -1,0 +1,19 @@
+import { Collection, ImportDeclaration } from "jscodeshift";
+import { CORE_IMPORT_PATH, CORE_NEXT_IMPORT_PATH } from "../consts";
+import { VibeCorePath } from "../../types";
+
+function getImports(root: Collection, path: VibeCorePath): Collection<ImportDeclaration> {
+  return root.find(ImportDeclaration, {
+    source: {
+      value: path
+    }
+  });
+}
+
+export function getCoreImportsForFile(root: Collection): Collection<ImportDeclaration> {
+  return getImports(root, CORE_IMPORT_PATH);
+}
+
+export function getCoreNextImportsForFile(root: Collection): Collection<ImportDeclaration> {
+  return getImports(root, CORE_NEXT_IMPORT_PATH);
+}
--- a/packages/codemod/src/utils/import-utils.ts
+++ b/packages/codemod/src/utils/import-utils.ts
@@ -2,6 +2,9 @@ import { ASTPath, Collection, ImportDeclaration, ImportSpecifier, JSCodeshift } 
 import { CORE_IMPORT_PATH, CORE_NEXT_IMPORT_PATH } from "../consts";
 import { VibeCorePath } from "../../types";
 
+/**
+ * Retrieves all import declarations that matches a path for a root (file).
+ */
 function getImports(root: Collection, path: VibeCorePath): Collection<ImportDeclaration> {
   return root.find(ImportDeclaration, {
     source: {
@@ -10,14 +13,23 @@ function getImports(root: Collection, path: VibeCorePath): Collection<ImportDecl
   });
 }
 
+/**
+ * Retrieves all "monday-ui-react-core" import declarations for a root.
+ */
 export function getCoreImportsForFile(root: Collection): Collection<ImportDeclaration> {
   return getImports(root, CORE_IMPORT_PATH);
 }
 
+/**
+ * Retrieves all "monday-ui-react-core/next" import declarations for a root.
+ */
 export function getCoreNextImportsForFile(root: Collection): Collection<ImportDeclaration> {
   return getImports(root, CORE_NEXT_IMPORT_PATH);
 }
 
+/**
+ * Retrieves all import specifiers (named imports) from a given import declaration.
+ */
 function extractSpecifiersFromImport(j: JSCodeshift, path: ASTPath<ImportDeclaration>): ImportSpecifier[] {
   const specifiers: ImportSpecifier[] = [];
   j(path)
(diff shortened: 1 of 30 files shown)


packages/codemod/transformations/core/v2-to-v3/attention-box-component-migration.ts
```diff
@@ -0,0 +1,26 @@
...
+  wrap
+} from "../../../src/utils";
+import { TransformationContext } from "../../../types";
+
+/**
+ * 1. Update the 'componentClassName' prop to 'className'
+ */
+function transform({ j, root }: TransformationContext) {
+  const imports = getCoreImportsForFile(root);
+  const actualComponentName = getComponentNameOrAliasFromImports(j, imports, "AttentionBox");
```
shaharzil, line 15: Im not sure what i think about the name "actualComponentName" (since it is not the actual, it is the alias) how about just "componentName" or "importedComponentName"?

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+  const attributes = elementPath.node?.attributes;
+  if (!attributes) return;
+
+  attributes.forEach(attr => {
+    if (
+      attr.type === "JSXAttribute" &&
+      attr.name.type === "JSXIdentifier" &&
+      propsNamesMappingOldToNew[attr.name.name]
+    ) {
+      attr.name.name = propsNamesMappingOldToNew[attr.name.name];
```
shaharzil, line 16: What will happen if for some reason the component already using the new prop in addition to old one? for example:
< AttentionBox componentClassName="old-class" className="other classes" />

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+export function updatePropName(
+  elementPath: ASTPath<JSXOpeningElement>,
+  propsNamesMappingOldToNew: Record<string, string>
+): void {
+  const attributes = elementPath.node?.attributes;
+  if (!attributes) return;
+
+  attributes.forEach(attr => {
+    if (
+      attr.type === "JSXAttribute" &&
```
shaharzil, line 12: Consider adding question marks after attr and attr?.name

packages/codemod/src/utils/import-utils.ts
shaharzil, line 1: Consider adding some comment for documentation about some of the functions or terms we use here

packages/codemod/src/utils/import-utils.ts
```diff
@@ -0,0 +1,46 @@
...
+  paths: Collection<ImportDeclaration>,
+  componentName: string
+): string | null {
+  let actualName = null;
+  paths.forEach(path => {
+    const componentImportSpecifier = j(extractSpecifiersFromImport(j, path))
+      .filter(spec => spec.node.imported.name === componentName)
+      .at(0);
+    if (componentImportSpecifier.length) {
+      actualName = componentImportSpecifier.get().node.local.name;
```
shaharzil, line 42: same comment like what i wrote about this naming in the attention box file :)

packages/codemod/src/utils/prop-utils.ts
YossiSaadi: `attr.type` would always exist if we're inside the `forEach` loop
`attr.name` throws TS errors if the `attr.type === "JSXAttribute"` isn't checked.
so it means that if `attr.type === "JSXAttribute"` is `true`, then `attr.name` would always exist
same for `attr.name.type` and `attr.name.name`.

As those would always exist, I think it is making the code more readable having no optional chaining so the reader would know those would always exist

image
image

packages/codemod/src/utils/import-utils.ts
YossiSaadi: any chance you can be more specific regarding this one? on stuff that were unclear?
basically that's the reason I've used types wherever possible, the jscodeshift types are pretty self explained once accessed through the IDE

I also want to make a short meeting with the team regarding the codemods infra and how to write codemods, so I want to explain the basics as I explained to you on our meeting, with the ASTExplorer (to understand what are import declarations, Collection vs ASTPath, etc. there's not a good central place to explain about all those as those are stuff that are shared between all the utils and all the codemods)

packages/codemod/src/utils/prop-utils.ts
shaharzil: Not sure how this code will resolve the use case I wrote..

YossiSaadi: you asked what will happen, not how to resolve 😅 
I don't think we should do anything in this case
TS would yell at the consumer that `componentClassName`, and if the consumer does not have TS configured it is its responsibility to check the docs when migrating 🤷🏼‍♂️

If we decide to remove the `componentClassName` no matter what is its content in case `className` also exists, then the user would most probably get hurt (unexpected and unaware changes)

YossiSaadi: sorry, fixed the snippet, was probably confusing

see the fixed one

it is also already committed so you can just go over the commit

YossiSaadi: image

```typescript
elements.forEach(path => {
    const hasClassName = isPropExists(j, path, "className");
    const hasComponentClassName = isPropExists(j, path, "componentClassName");
    if (hasClassName && hasComponentClassName) {
      logPropMigrationWarning(filePath, componentName, "componentClassName", "className");
    } else {
      updatePropName(path, { componentClassName: "className" });
    }
  });
```

shaharzil (approved):

PR #2215, opened 2024-07-07
chore(Avatar): codemod for migration of Avatar from core - v2 to v3

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/6958806157


packages/codemod/src/utils/prop-utils.ts
```diff
@@ -1,31 +1,129 @@
...
+  // can be very complex, we'll have to compare strings
+  return value ? j(value).toSource() : undefined;
+}
+
+/**
+ * Checks if a Collection of props have the same value
+ */
+export function propsValueMatch(j: JSCodeshift, propsCollection: Collection<JSXAttribute>): boolean {
+  const values = propsCollection.nodes().map(prop => getPropValue(j, prop));
+  return values.every(value => value === values[0]);
```
shaharzil, line 98: Maybe I'm missing something, but why do we need to use "every"? This function supposed to compare between only 2 values, am I right? Why not a simple value[0] === value[1]

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -1,31 +1,129 @@
...
+): void {
+  Object.entries(propsNamesMappingOldToNew).forEach(([deprecatedPropName, newPropName]) => {
+    const props = findProps(j, elementPath, deprecatedPropName, newPropName);
+
+    if (props.length !== 2) {
+      updatePropName(j, elementPath, { [deprecatedPropName]: newPropName });
+      return;
+    }
+
+    if (propsValueMatch(j, props)) {
```
shaharzil, line 123: Not needed for now, but let's remember for the future that this will work only on strings and booleans, but won't catch arrays and objects

packages/codemod/transformations/core/v2-to-v3/__tests__/attention-box-component-migration.test.ts
```diff
@@ -56,7 +56,47 @@ describe("AttentionBox component migration", () => {
...
+    prependImport(`<AttentionBox className="class-one" />`),
+    "should remove 'componentClassName' when when both 'componentClassName' and 'className' props exist with same literal values"
+  );
+
+  defineInlineTest(
+    transform,
+    {},
+    prependImport(`<AttentionBox componentClassName={"class-one"} className="class-one" />`),
+    prependImport(`<AttentionBox className="class-one" />`),
+    "should remove 'componentClassName' when when both 'componentClassName' and 'className' props exist with same literal values while one is inside an expression"
```
shaharzil, line 99: Very good!!!!

packages/codemod/transformations/core/v2-to-v3/__tests__/avatar-component-migration.test.ts
shaharzil, line 1: Maybe let's add another test:
    prependImport(`<Avatar square isDisabled />`),
    prependImport(`<Avatar square disabled />`),

To make sure it wont stop on the first one that doesn't need changes

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -1,31 +1,129 @@
...
-      propsNamesMappingOldToNew[attr.name.name]
-    ) {
-      attr.name.name = propsNamesMappingOldToNew[attr.name.name];
-    }
-  });
+  j(elementPath)
+    .find(JSXOpeningElement)
+    .find(JSXIdentifier)
+    .forEach(identifierPath => {
+      if (propsNamesMappingOldToNew[identifierPath.node.name]) {
```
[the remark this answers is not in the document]
YossiSaadi: since we're inside "JSXIdentifier", we'll always have node.name, otherwise we won't get into the "forEach"
the previous way was not a jscodeshift-way, but instead manipulating the node, this way is more readable and efficient

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -1,31 +1,129 @@
...
-  if (!attributes) return false;
-  return j(elementPath).find(JSXIdentifier, { name: propName }).size() > 0;
+export function isPropExists(j: JSCodeshift, elementPath: ASTPath<JSXElement>, propName: string): boolean {
+  return j(elementPath).find(JSXOpeningElement).find(JSXIdentifier, { name: propName }).size() > 0;
+}
+
+/**
+ * Finds props on a JSX element
+ */
+export function findProps(j: JSCodeshift, elementPath: ASTPath<JSXElement>, ...propNames: string[]) {
```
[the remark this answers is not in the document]
YossiSaadi: not spreading, but rather accepting infinite number of props by propNames

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -1,31 +1,129 @@
...
+      }
+    });
 }
 
 /**
  * Checks for whether a prop is used in a JSX element
  */
-export function isPropExists(j: JSCodeshift, elementPath: ASTPath<JSXOpeningElement>, propName: string): boolean {
-  const attributes = elementPath.node?.attributes;
-  if (!attributes) return false;
```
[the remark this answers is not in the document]
YossiSaadi: a mistake 😶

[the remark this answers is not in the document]
YossiSaadi: the `find` of jscodeshift is taking care of it instead of the condition, it would return an empty collection instead of manipulating the code directly

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -1,31 +1,129 @@
...
+    .filter(attr => {
+      const attrName = attr.node.name.type === "JSXIdentifier" ? attr.node.name.name : "";
+      return propNames.includes(attrName);
+    });
+}
+
+/**
+ * Removes a prop from a JSX element
+ */
+export function removeProp(j: JSCodeshift, elementPath: ASTPath<JSXElement>, ...propsNames: string[]): void {
```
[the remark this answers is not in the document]
YossiSaadi: not spreading, but rather removing infinite number of props by accepting infinite amount of propNames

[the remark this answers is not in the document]
YossiSaadi: ```findProps(j, elementPath, ...propsNames).remove();```
can definitely get the job done, nice catch!

packages/codemod/src/utils/prop-utils.ts
YossiSaadi: just one of the js ways to make sure all items in array are equal, I do not want a compare function, but to know if all are equal

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -1,31 +1,129 @@
...
+}
+
+/**
+ * Gets the computed value of a prop
+ */
+function getPropValue(j: JSCodeshift, prop: JSXAttribute): undefined | boolean | Literal["value"] | string {
+  const value = prop.value;
+
+  // boolean flag value (e.g. <Button disabled />)
+  if (value === null) {
```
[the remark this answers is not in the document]
YossiSaadi: it would get into the
```if (value?.type === "JSXExpressionContainer") {```
`null` only means that it is a boolean `true` flag prop (without "" or {}, declaring `null` must be inside {})

File: packages/codemod/transformations/core/v2-to-v3/__tests__/avatar-component-migration.test.ts
Comment (YossiSaadi) [reply packages/codemod/transformations/core/v2-to-v3/__tests__/avatar-component-migration.test.ts:1]: we have one [here](https://github.com/mondaycom/vibe/pull/2215/files/f8359adb2cf22c45f295298220d9383f32ded3c5#diff-18a0432986436950ba966db9f0ba660b075cbb91f73b297508e8ef1da1df1878R28)

File: packages/codemod/src/utils/prop-utils.ts (hunk 2 above)
Comment (YossiSaadi) [reply packages/codemod/src/utils/prop-utils.ts:123]: it actually does catch arrays and objects, but in a very high level way
e.g. this would work:
```
const someArray = [1, 2, 3];
<MyComponent myArray={someArray} myNewArray={someArray} />
```

but for now, this won't work
```
const someArray = [1, 2, 3];
<MyComponent myArray={someArray} myNewArray={[1, 2, 3]} />
```
and honestly I'm not sure this **should** work as this can have different meaning in code, stuff that are done on purpose

so right now we're catching
<MyComponent **someFlag** **someFlag={true}** **someFlag={false}** **someString={"hello"}** someString="hello" />
which I think are more than enough for now

packages/codemod/src/utils/prop-utils.ts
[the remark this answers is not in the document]
shaharzil: Sorry, missed what you are sending, I thought it was only one array 🤦‍♀️

packages/codemod/src/utils/prop-utils.ts
shaharzil: But is it possible to have more then 2 items in this specific array?

packages/codemod/src/utils/prop-utils.ts
shaharzil: Yes, this is what I meant, if it is the same value but different pointer. 
I think this is fine for now, but some of our props' types are array so we need to keep that in mind for later

packages/codemod/transformations/core/v2-to-v3/__tests__/avatar-component-migration.test.ts
shaharzil: Are you sure that the link is for the correct line? it is different from what I wrote..

packages/codemod/src/utils/prop-utils.ts
YossiSaadi: yep, it can include a lot of props in theory (probably 99.99% of cases it won't, unless we'll encounter some very weird use case, but it's good to keep it generic)

packages/codemod/src/utils/prop-utils.ts
YossiSaadi: yes so for now it will check exact match
also cases like
```onClick={() => doSomething()} and onClick={doSomething}```
won't be caught as equals, but that is also cool I think, until we'll encounter somewhere where a specific use case is needed

packages/codemod/transformations/core/v2-to-v3/__tests__/avatar-component-migration.test.ts
YossiSaadi: @shaharzil 
image

This is what I've tagged
Is this opening a different line for you?

YossiSaadi: ohh I missed what you meant
ok will add one

shaharzil (approved):

PR #2220, opened 2024-07-10
feat: add debounce rate prop to combobox

moro256:
Add prop to combobox to allow changing its debounce rate


packages/core/src/components/Combobox/Combobox.tsx
```diff
@@ -151,7 +152,8 @@ const Combobox: React.FC<ComboboxProps> & {
       maxOptionsWithoutScroll,
       defaultFilter: defaultFilterValue = "",
       searchInputAriaLabel = "Search for content",
-      "data-testid": dataTestId
+      "data-testid": dataTestId,
+      debounceRate = 400
```
talkor, line 156: Let's remove the default value as it gets if from the Search component

talkor (approved):
Thanks!

PR #2221, opened 2024-07-10
feat(textField): Add option for tooltip on icon

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6965420722

Added an option for displaying a tooltip on the icon, regardless if it is clickable or not.

Screenshot 2024-07-15 at 13 37 42


packages/core/src/components/LegacySearch/LegacySearch.tsx
```diff
@@ -113,6 +113,7 @@ const LegacySearch: VibeComponent<SearchProps, unknown> & {
         role="search"
         loading={loading}
         underline={type === SearchType.UNDERLINE}
+        iconsNames={iconNames}
```
YossiSaadi, line 116: remove the eslint for unused var at [line 78](https://github.com/mondaycom/vibe/pull/2221/files#diff-f06badecccb1655de11b529c2bad5b3f9d88fe48331bb3f3857dbe8588583818R79)
also, should remove the `layout` from the Search fields now, right?

YossiSaadi: or is it cool that it is the way it is right now and it is ok to leave it?

packages/core/src/components/TextField/TextField.module.scss
```diff
@@ -111,6 +111,23 @@
...
 }
 
+.textField .inputWrapper .tooltipContainer {
+  position: absolute;
+  top: 50%;
+  right: 0;
+  width: 32px;
+  height: 20px;
+  transform: translateY(-50%);
+}
```
YossiSaadi, line 121: 🙃
We'll deprecate this TextField one day 🤞🏼

packages/core/src/components/TextField/__tests__/__snapshots__/TextField.snapshot.test.js.snap
```diff
@@ -37,6 +37,7 @@ exports[`TextField renders correctly when disabled 1`] = `
         value=""
       />
       <div
+        aria-label=""
```
YossiSaadi, line 40: it sound like the prop should be `undefined` if aria-label is empty, isn't it? and then the snapshots won't get an empty string aria label (and maybe in general it is better)

packages/core/src/components/TextField/TextField.tsx
```diff
@@ -100,6 +101,8 @@ export interface TextFieldProps extends VibeComponentProps {
    * When true, component is controlled by an external state
    */
   controlled?: boolean;
+  iconTooltipContent?: string;
+  secondaryTooltipContent?: string;
```
YossiSaadi, line 105: those are maybe problematic (the usage of them)
does the tooltip necessarily needs to be empty if not passed this prop? we should be backwards compatible and have the tooltip show the label, no?

YossiSaadi (commented):
Seems cool by me, awesome work!
I had some questions regarding the snapshots result and the usage of the new props which a bit worries me about backward compatibility

packages/core/src/components/LegacySearch/LegacySearch.tsx
rivka-ungar: It would be a breaking change to remove the `layout` now, no?

packages/core/src/components/TextField/TextField.module.scss
rivka-ungar: Hopefully sooner than later

packages/core/src/components/TextField/TextField.tsx
rivka-ungar: The tooltip was only added now, it didn't exist in this component before, we should add it to display the label?

packages/core/src/components/LegacySearch/LegacySearch.tsx
YossiSaadi: meant in `SearchDefaultIconNames` but on 2nd thought it is gonna be removed soon with the entire Search component, so no rush on that one

YossiSaadi: meant in `SearchDefaultIconNames` but on 2nd thought it is gonna be removed soon with the entire `LegacySearch` component, so no rush on that one

packages/core/src/components/TextField/TextField.tsx
YossiSaadi: but the behavior before existed

we do have tooltips today, aren't we?

YossiSaadi: clarifyied on DM, didn't recalled we did not had tooltips up until today on TextField 🫠

packages/core/src/components/TextField/__tests__/__snapshots__/TextField.snapshot.test.js.snap
```diff
@@ -338,7 +356,9 @@ exports[`TextField renders correctly with another type 1`] = `
         onMouseLeave={[Function]}
         role="button"
         tabIndex={-1}
-      />
+      >
+
```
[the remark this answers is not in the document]
rivka-ungar: I added an accidental space in the code, fixed now

packages/core/src/components/TextField/__tests__/__snapshots__/TextField.snapshot.test.js.snap
rivka-ungar: Fixed

packages/core/src/components/TextField/__tests__/__snapshots__/TextField.snapshot.test.js.snap
```diff
@@ -57,7 +59,9 @@ exports[`TextField renders correctly when disabled 1`] = `
...
```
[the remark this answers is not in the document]
rivka-ungar: Done

packages/core/src/components/TextField/__tests__/__snapshots__/TextField.snapshot.test.js.snap
```diff
@@ -1510,7 +1510,7 @@ exports[`TextField renders correctly with value 1`] = `
         onMouseEnter={[Function]}
         onMouseLeave={[Function]}
         role="button"
-        tabIndex={-1}
```
YossiSaadi, line 1513: is that on purpose?

YossiSaadi (approved):

packages/core/src/components/TextField/__tests__/__snapshots__/TextField.snapshot.test.js.snap
rivka-ungar: Fixed

PR #2227, opened 2024-07-14
feat(icons): Add Item Height Double Icon

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7020616472


talkor (approved):

packages/style/src/Icons/iconsMetaData.js
```diff
@@ -4,6 +4,13 @@ const VIEW = "View";
 export const iconsMetaData = [
   // DO NOT REMOVE THIS FOLLOWING LINE!
   // plop_marker:icon_metadata
+  {
+      name: "Item Height Double",
```
talkor, line 8: The name should be without spaces @rivka-ungar

PR #2229, opened 2024-07-14
chore(Tipseen): remove deprecated props [breaking]

talkor:
https://monday.monday.com/boards/3532714909/pulses/6965588862

### DIFF
--- a/packages/core/docs/vibe-3-changelog.md
+++ b/packages/core/docs/vibe-3-changelog.md
@@ -153,6 +153,7 @@
 - `showDelay` changed default to 100
 - `justify` removed [codemod]
 - `justifyTypes` removed [codemod]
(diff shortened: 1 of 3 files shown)


YossiSaadi (approved):
Can you please add in the changelog what the codemod should do in this case?

talkor:
> Can you please add in the changelog what the codemod should do in this case?

I don't think we should add a codemod here since some customization is no longer supported and requires developer's attention

YossiSaadi:
> > Can you please add in the changelog what the codemod should do in this case?
> 
> I don't think we should add a codemod here since some customization is no longer supported and requires developer's attention

I think we should consider injecting a TODO comment on everywhere we're not planning to touch with the codemod

PR #2231, opened 2024-07-14
feat(TextField): send native event if applicable in onChange

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7035581219

### DIFF
--- a/packages/core/src/components/TextField/TextField.tsx
+++ b/packages/core/src/components/TextField/TextField.tsx
@@ -11,7 +11,6 @@ import React, {
 } from "react";
 import useDebounceEvent from "../../hooks/useDebounceEvent";
 import Icon from "../Icon/Icon";
-import { backwardCompatibilityForProperties } from "../../helpers/backwardCompatibilityForProperties";
 import Loader from "../Loader/Loader";
 import Text from "../Text/Text";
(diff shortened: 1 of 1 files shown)


packages/core/src/components/TextField/TextField.tsx
```diff
@@ -176,11 +179,12 @@ const TextField: VibeComponent<TextFieldProps, unknown> & {
     );
 
     const onChangeCallback = useCallback(
-      (value: string) => {
+      (value: string, e?: React.ChangeEvent<HTMLInputElement>) => {
         if (isRequiredAndEmpty && value) {
           setIsRequiredAndEmpty(false);
         }
-        onChange(value, { target: inputRef.current });
+        const event = e || { target: inputRef.current };
```
talkor, line 186: Can e be ever undefined?

talkor (approved):

packages/core/src/components/TextField/TextField.tsx
YossiSaadi: It would be undefined when the "clear" button is clicked on the Search variant (if you're asking me, special cases like this, should be same as I did in the new Search - in the Search component and not in the Input, which should be as generic as possible.)

PR #2232, opened 2024-07-14
chore(TextField): remove requiredAsterisk prop

talkor:
https://monday.monday.com/boards/3532714909/pulses/7020874956


packages/core/src/components/TextField/__stories__/TextField.mdx
```diff
@@ -123,8 +123,6 @@ Users can insert text.
 
 ### Required input field
 
-Use the <code>requiredAsterisk</code> prop to indicate that a field is required. When set to <code>true</code>, an asterisk (\*) will appear next to the label.
```
YossiSaadi, line 126: shouldn't be changed to `required`? and not to be removed entirely?

talkor: `required` already exists

YossiSaadi: I meant in this story

talkor: Oh, I think it's self explanatory so I removed it entirely, WDYT?

YossiSaadi: sounds good, it also contains the code with required so sound cool

packages/core/src/components/FieldLabel/FieldLabel.tsx
```diff
@@ -43,8 +43,7 @@ const FieldLabel: FC<FieldLabelProps> = forwardRef(
         />
         <label htmlFor={labelFor} ref={ref} className={cx(styles.labelComponentText, labelClassName)}>
           {labelText}
-          {requiredAsterisk && <span className={styles.requiredAsterisk}> *</span>}{" "}
-          {/* Render asterisk based on the prop */}
+          {required && <span className={styles.required}> *</span>}
```
shaharzil, line 46: I see that you removed the space after the asterisk {" "}, why dont we need it anymore?

talkor: It has no effect, this is master today:
Screenshot 2024-07-17 at 12 18 15
Probably it was added by Prettier. Plus there's no other element on this block

shaharzil (approved):

PR #2233, opened 2024-07-14
fix(DialogContentContainer): fix size medium spacing

talkor:
https://monday.monday.com/boards/3532714909/pulses/7020937327


packages/core/src/components/DialogContentContainer/DialogContentContainer.module.scss
```diff
@@ -7,7 +7,7 @@
 }
 
 .sizeMedium {
-  padding: var(--spacing-small);
+  padding: var(--spacing-medium);
```
shaharzil, line 10: Don't we need to add it also to the change log?

shaharzil (approved):

PR #2238, opened 2024-07-16
feat(codemod): add plop capabilities to auto-generate common codemods

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7025352164


packages/codemod/plop/component/test.hbs
```diff
@@ -0,0 +1,19 @@
...
+import { defineInlineTest } from "jscodeshift/src/testUtils";
+
+function prependImport(source: string): string {
+  return `
+    import { {{pascalCase componentName}} } from "monday-ui-react-core";
+    ${source}
+  `;
+}
+
+describe("{{pascalCase componentName}} component migration", () => {
```
talkor, line 11: Does it pass right after the generation?

packages/codemod/plop/component/transform-remove-props.hbs
```diff
@@ -0,0 +1,26 @@
...
+  wrap,
+  getCoreImportsForFile,
+  getComponentNameOrAliasFromImports,
+  findComponentElements,
+  removeProp
+} from "../../../src/utils";
+import { TransformationContext } from "../../../types";
+
+/**
+ * 1. TODO: What this codemod does?
```
talkor, line 11: Why the numeric list? Is there another thing to do?

talkor (approved):
Super cool!!

packages/codemod/plop/component/test.hbs
YossiSaadi: actually no, but maybe it's good that it won't pass, so it would attract attention for the plop user

packages/codemod/plop/component/transform-remove-props.hbs
YossiSaadi: on some of the codemods we're doing more than one action
as we continute with Vibe3, probably most of the codemods would include more than one action

PR #2239, opened 2024-07-17
docs(storybook): enable ui regression testing with themes

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/4999786657


talkor (approved):
🥳

PR #2251, opened 2024-07-22
feat(Tipseen): New paddings for tipseen

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/6947400555


YossiSaadi (approved):

packages/core/src/components/Tipseen/TipseenBasicContent.module.scss
```diff
@@ -1,16 +1,17 @@
 @mixin tipseen-content-padding {
-  padding: var(--spacing-medium);
+  padding: var(--spacing-medium) 20px 20px 20px;
+
+  // this is because the designer wanted the padding top
+  // to be 12px in case the tipseen have both content and image
+  &:not(:first-child) {
+    padding-top: 12px;
+  }
```
YossiSaadi, line 8: only one question, why did you changed it?

shaharzil: It felt like a better location (since the mixin name related to the padding)

YossiSaadi: It sound like the mixin is just very redundant, is it even exported from Vibe?

shaharzil: I agree
It is not exported,do you think it will be best if I will just remove it?

YossiSaadi: if for sure not exported, let's remove it, yes
if it takes time to check it so leave it as is and create a task for future check 💪🏼

PR #2265, opened 2024-07-24
fix(Dropdown): nanoid 4+ supports only ESM, we should still support CJS

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7086495216

@rivka-ungar @talkor


talkor (approved):

YossiSaadi:
prerelease failed for bad branch name

PR #2266, opened 2024-07-24
feat: add `searchInputRef` prop to Combobox

eladdo92:


eladdo92:
@YossiSaadi  Sure! 
We want to use the combobox with an external button for adding labels (see image).
When the user clicks on this button, we want to (add the new label obviously and then) focus back to the input. In addition we also want to trigger the same "add" logic when the user press Enter in the search-input (after typing something).

Screenshot 2024-07-24 at 16 40 27

Both focusing & listening to Enter can be achieved using this ref :)

BTW, we also want to clear the input after adding a new label (by clicking the button or pressing Enter).
To achieve that we will need a way to control the value of the input, I'll work on another PR to allow it

doronbrikman (approved):

RonNachmany (approved):

talkor:
@eladdo92 let's talk about it

talkor (approved):

PR #2273, opened 2024-07-25
chore(Label): deprecate enums

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7021050514

### DIFF
--- a/packages/core/src/components/Label/Label.types.ts
+++ b/packages/core/src/components/Label/Label.types.ts
@@ -3,6 +3,10 @@ import Text from "../Text/Text";
 
 export type Sizes = "small" | "medium";
 
+export type LabelKind = "fill" | "line";
+
+export type LabelColor = "primary" | "dark" | "negative" | "positive";
+
 export const mapSizesToTextSize: Record<Sizes, TextType> = {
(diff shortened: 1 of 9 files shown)


packages/core/src/components/Label/Label.types.ts
```diff
@@ -3,6 +3,10 @@ import Text from "../Text/Text";
 
 export type Sizes = "small" | "medium";
 
+export type LabelKind = "fill" | "line";
+
+export type LabelColor = "primary" | "dark" | "negative" | "positive";
+
 export const mapSizesToTextSize: Record<Sizes, TextType> = {
```
talkor, line 10: I don't think we should put this here as it's not a type, plus I don't think we should export it

packages/core/src/components/Label/Label.types.ts
```diff
@@ -3,6 +3,10 @@ import Text from "../Text/Text";
 
 export type Sizes = "small" | "medium";
```
talkor, line 4: I'm a bit worried about this, we have too many generic types like this

talkor (approved):

Issue #2106, opened 2024-05-06
[Feat] `onClear` attribute for the `Search` component


YossiSaadi:
### Discussed in https://github.com/mondaycom/vibe/discussions/2101

<div type='discussions-op-text'>

<sup>Originally posted by **eduardo-wrkt** May  3, 2024</sup>
Hi everyone!

I've been using the library for almost a year now and I love it, its super clean and easy to use.

One thing that I think would be super useful, is a `onClear` attribute in the **Search** component. Currently I have to add an event listener to run a function when the user clicks in the **clear button** of the Search input. With this new attribute, that wouldn't be necessary anymore.

I suppose this would be a small lift but also a major improvement for the component.

That's it. Thanks in advance.</div>

YossiSaadi:
@eduardo-wrkt feel free to take it :)
Thank you!

YossiSaadi:
Awesome work!!

PR #2272, opened 2024-07-25
feat: add onClear prop to Search component

eduardo-wrkt:
Hi everyone!

Attending the idea discussion [#2101](https://github.com/mondaycom/vibe/discussions/2101), I've added the `onClear` prop to the `Search` component. It's a simple change, but at the same time, a useful one.

I decided to just add a verification inside of the function that clears the search input value once the users clicks the clear button, keeping the existing logic untouched and adding another step before clearing the input value.

I hope this change make sense to you guys!

### Checklist
- [x] I have read the [Contribution Guide](../packages/core/CONTRIBUTING.md) for this project.

Thanks in advance!

PR #2276, opened 2024-07-25
feat(Search): Add `onClear` prop+testing+story for `Search`

itaim18:

I took on myself to solve #2106 , additionaly I added it to the docs + usecase for storybook+ testing.

I hope this attempt will go through

pedro-monkey-puppet

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -3,6 +3,23 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.125.0](https://github.com/mondaycom/vibe/compare/monday-ui-react-core@2.124.1...monday-ui-react-core@2.125.0) (2024-07-25)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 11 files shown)


YossiSaadi (changes requested):
Hey @itaim18! Thanks for your awesome contribution!
Unfortunately, #2272 was created about 2 hours before this PR regarding the same issue #2106.

Any chance you can merge from `master` and add your test? It would help us a lot!

Two more things:
- We do not feel, currently, a need to add a story for this use case, so we can pass on it for now
- In general, if you add a default of `NOOP` (which we prefer to not do anymore), you do not need to have optional chaining (`onClear?.()`), as it would always be a function. 

You can change the version that's on `master` right now (`if onClear` at the beginning) to yours (`onClear?.()` at the end) without the `NOOP` default

Thanks again and looking forward for your ping to re-review :)

YossiSaadi (commented):
seem like `onClear` is destructed twice, you'll need to remove the first prop and remove the `if onClear`
same for in the types

YossiSaadi (approved):
Awesome work @itaim18 !
Thanks for another contribution for Vibe, keep rocking it 💪🏼

PR #2278, opened 2024-07-28
chore(Skeleton): deprecate enums

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7021051396


packages/core/src/components/Skeleton/Skelton.types.ts
```diff
@@ -0,0 +1,3 @@
+export type SkeletonType = "circle" | "rectangle" | "text";
+
+export type TextSkeletonSize = "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "small";
```
talkor, line 3: What about the 'custom'?

rivka-ungar: The custom is included in SkeletonSizeType, it should also be added here?

talkor: TextSkeletonSize shouldn't be exported, no?

talkor (approved):

packages/core/src/components/Skeleton/Skelton.types.ts
rivka-ungar: Your right, removed the export

PR #2281, opened 2024-07-29
feat(Menu): allow setting submenu position to left

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7110141799

### DIFF
--- a/packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
+++ b/packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
@@ -26,17 +26,17 @@ describe("<MenuItem />", () => {
 
   const submenuPositions = [
     {
-      submenuPosition: "left-start",
+      forceSubmenuOpenLeft: true,
       expectedPosition: "left-start"
     },
     {
-      submenuPosition: undefined,
+      forceSubmenuOpenLeft: undefined,
       expectedPosition: "right-start"
     }
   ];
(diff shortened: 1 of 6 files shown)


YossiSaadi (commented):
I still think we can have a "position" prop to be able to extend this behavior. It should default to right and have options of right or left.
internally you can do exactly what you do right now, but check if the position is left (the undefined behavior should be position right)
let's discuss it
@talkor wdyt?

packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
```diff
@@ -24,6 +24,49 @@ describe("<MenuItem />", () => {
     expect(queryAllByText("Sub Item")).toHaveLength(0);
   });
 
+  const submenuPositions = [
+    {
+      submenuPosition: "left",
+      expectedPosition: "left-start"
+    },
+    {
+      submenuPosition: undefined,
```
YossiSaadi, line 33: is it still like that?

YossiSaadi (approved):
LGTM!

packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
rivka-ungar: yes, it is checking the case where the user doesnt send the prop

PR #2285, opened 2024-07-29
docs: fix stories dark mode

talkor:

Need it for the vibe3 branch, to accept chromatic changes without compromising about this


packages/core/.storybook/preview-head.html
```diff
@@ -17,6 +17,7 @@
   body {
     font-family: "Figtree", sans-serif;
     box-sizing: border-box;
+    color: var(--sb-primary-text-color);
```
talkor, line 20: Not sure why we use the --sb prefix, we can use our tokens, but it's out of scope

YossiSaadi (approved):

PR #2292, opened 2024-07-30
fix: Add support for generic types in the TableVirtualizedBody component

daniilvo:
## Overview
TableVirtualizedBody doesn't support generic types. Which means that if you have typescript configured in the project with `"strict": true`, you will get an error like this:
image

## Changes:
- Added support for generic types in `TableVirtualizedBody` component, to make it more type-safe
- Converted `Table.stories.js` file to `.tsx` + necessary types updates (without logic changes)

### P.S. 
At the moment vibe repository is NOT strictly typed (`"strict": true` in `tsconfig.json`). Therefore, TableVirtualizedBody component works fine in storybook even without my changes. Which makes almost impossible to catch that kind of bug during development.

To avoid similar bugs in future, make Vibe type-safe and basically make it possible to use Vibe safely in the projects which are strictly typed, I highly encourage you to start gradually migrating to `"strict": true` typescript setting. It might be a long process, but it worth it.

### DIFF
--- a/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
+++ b/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import Avatar from "../../../Avatar/Avatar";
 import { Calendar, Doc, Status } from "../../../Icon/Icons";
 import { LabelColor } from "../../../Label/LabelConstants";
-import { TableVirtualizedRow } from "../../TableVirtualizedBody/TableVirtualizedBody";
 import { ITableColumn } from "../Table";
 
 export const doAndDontIconsRuleColumns = [
--- a/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
+++ b/packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
@@ -118,7 +118,7 @@ export const emailTableData = [
   }
 ];
 
-export const emailColumns = [
+export const emailColumns: ITableColumn[] = [
   {
     id: "sentOn",
     title: "Sent on",
@@ -258,14 +258,14 @@ export const highlightableTableData = [
   }
 ];
 
-export const priorityColumnToLabelColor = {
+export const priorityColumnToLabelColor: { [key: string]: LabelColor } = {
   Urgent: LabelColor.NEGATIVE,
   High: LabelColor.DARK,
   Normal: LabelColor.PRIMARY,
   Low: LabelColor.POSITIVE
 };
 
-export const statusColumnToLabelColor = {
+export const statusColumnToLabelColor: { [key: string]: LabelColor } = {
   Sent: LabelColor.POSITIVE,
   Queued: LabelColor.DARK,
   Failed: LabelColor.NEGATIVE,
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
```diff
@@ -325,7 +325,7 @@ export const virtualizedScrollTableColumns: ITableColumn[] = [
   }
 ];
 
-export function sort(columnId: string, sortState: "asc" | "desc" | "none", tableData: []) {
+export function sort<T>(columnId: keyof T, sortState: "asc" | "desc" | "none", tableData: T[]) {
```
kapusj, line 328: how about adding some default? Wouldn't it break when no type is provided (all of the existing code using this method)

packages/core/src/components/Table/Table/__stories__/Table.stories.tsx
```diff
@@ -402,14 +403,14 @@ export const Borders = {
...
+    const [sorting, setSorting] = useState<{[key: string]: "asc" | "desc" | "none"}>({});
 
-    const onSort = (columnId, sortState) => {
+    const onSort = (columnId: string, sortState: "asc" | "desc" | "none") => {
       setSorting({
         [columnId]: sortState
       });
 
-      setTableData(sort(columnId, sortState, tableData));
+      setTableData(sort(columnId as keyof typeof tableData[number], sortState, tableData));
```
kapusj, line 413: why not to use the generic type? It was added to `sort` in this PR

packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
daniilvo: It's needed to provide better type-safety for the `columnId` prop. It will not allow you to pass just any random string, only real keys of `T` is allowed.

It is not necessary to add default value for the `T`, since it will be automatically infered from the provided `tableData` prop

packages/core/src/components/Table/Table/__stories__/Table.stories.tsx
daniilvo: In this case generic type is automatically determined from the provided `tableData` object

packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
```diff
@@ -307,7 +307,7 @@ export const scrollTableColumns = [
   }
 ];
 
-export const virtualizedScrollTableData: TableVirtualizedRow[] = [...new Array(5000)].map((_, index) => ({
+export const virtualizedScrollTableData = [...new Array(5000)].map((_, index) => ({
```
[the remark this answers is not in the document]
daniilvo: `TableVirtualizedRow` is very generic type (`Record<string, unknown> & { id: string }`). And by keeping this type here, you are basically erasing the information about `num` and `text` fields. All you will know for sure when accessing `virtualizedScrollTableData` variable is that it ONLY has `id` field.

However, if we delete the explicit type declaration, we will let typescript automatically determine the type and all keys will present in our variable (`id`, `num` and `text`):
image

packages/core/src/components/Table/Table/__stories__/Table.stories.helpers.tsx
daniilvo: @talkor do you still think default value necessary in this case? 1) It's an internal storybook-related file, it is not exposed from the library 2) type if determined automatically from the specified props in this case

talkor (approved):
Thanks!

PR #2293, opened 2024-07-31
feat: added placeholder to textarea

orihassid2507:


packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -29,7 +29,8 @@ const TextArea = forwardRef(
       onChange,
       "aria-label": ariaLabel,
       required,
-      resize = true
+      resize = true,
+      placeholder = ""
```
talkor, line 33: Let's remove the default value so it won't be rendered as an empty attribute. It will be undefined when not provided and won't be added as an empty attribute

PR #2297, opened 2024-08-01
feat(combobox): Add render action to combobox

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/7041131964

### DIFF
--- a/packages/core/src/components/Combobox/Combobox.tsx
+++ b/packages/core/src/components/Combobox/Combobox.tsx
@@ -24,6 +24,8 @@ import {
   COMBOBOX_LISTBOX_ID
 } from "./components/ComboboxConstants";
 import styles from "./Combobox.module.scss";
+import IconButton from "../IconButton/IconButton";
+import MenuButton from "../MenuButton/MenuButton";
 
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Combobox/Combobox.tsx
```diff
@@ -115,6 +117,14 @@ export interface ComboboxProps extends VibeComponentProps {
...
   debounceRate?: number;
   searchInputRef?: React.RefObject<HTMLInputElement>;
+  /**
+   * Render additional action within the right section of search component.
+   */
+  renderAction?: React.ReactElement<typeof IconButton | typeof MenuButton>;
+  /**
+   * If true, hides the additional action when the user types in the search input.
+   */
+  hideRenderActionOnInput?: boolean;
```
YossiSaadi, line 127: btw, did they requested this behavior?
if no, we don't have to supply them with that, to not overload the api with stuff that they don't necessarily need right now

shaharzil: Meirav do want me to add this prop

PR #2299, opened 2024-08-01
chore(Loader): deprecate enums

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7021050773

### DIFF
--- a/packages/core/src/__tests__/__snapshots__/exports.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/exports.test.ts.snap
@@ -106,8 +106,6 @@ exports[`exports should export all components 1`] = `
   "Tooltip",
   "VirtualizedGrid",
   "VirtualizedList",
-  "mapLoaderColorsToColors",
-  "mapSizesToLoaderSize",
   "useGridKeyboardNavigationContext",
 ]
 `;
--- a/packages/core/src/__tests__/__snapshots__/exports.test.ts.snap
+++ b/packages/core/src/__tests__/__snapshots__/exports.test.ts.snap
@@ -106,6 +106,8 @@ exports[`exports should export all components 1`] = `
   "Tooltip",
   "VirtualizedGrid",
   "VirtualizedList",
+  "mapLoaderColorsToColors",
+  "mapSizesToLoaderSize",
   "useGridKeyboardNavigationContext",
 ]
 `;
--- a/packages/core/src/components/Loader/Loader.types.ts
+++ b/packages/core/src/components/Loader/Loader.types.ts
@@ -1,5 +1,5 @@
 export type LoaderColors = "primary" | "secondary" | "onPrimary" | "dark";
 
-export type LoaderSizes = "xs" | "small" | "medium" | "large";
+type LoaderSizes = "xs" | "small" | "medium" | "large";
 
 export type LoaderSize = LoaderSizes | number;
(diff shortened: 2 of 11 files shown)


packages/core/src/__tests__/__snapshots__/exports.test.ts.snap
```diff
@@ -106,6 +106,8 @@ exports[`exports should export all components 1`] = `
   "Tooltip",
   "VirtualizedGrid",
   "VirtualizedList",
+  "mapLoaderColorsToColors",
+  "mapSizesToLoaderSize",
```
talkor, line 110: I don't think we need to export these

packages/core/src/components/Loader/LoaderConstants.ts
```diff
@@ -1,15 +1,21 @@
...
-  SMALL = 24,
-  MEDIUM = 40,
-  LARGE = 64
+  XS = "xs",
+  SMALL = "small",
+  MEDIUM = "medium",
+  LARGE = "large"
 }
 
 export type LoaderSize = typeof LoaderSizes | number;
```
talkor, line 21: Let's move it to the types file

packages/core/src/components/Loader/Loader.types.ts
```diff
@@ -0,0 +1,19 @@
...
+  medium: 40,
+  large: 64
+};
+
+export const mapLoaderColorsToColors: Record<LoaderColors, string> = {
+  primary: "primary-color",
+  secondary: "secondary-text-color",
+  onPrimary: "text-color-on-inverted",
+  dark: "primary-text-color"
+};
```
talkor, line 19: Let's move these to the component maybe?

packages/core/src/components/Loader/Loader.types.ts
```diff
@@ -0,0 +1,19 @@
+export type LoaderColors = "primary" | "secondary" | "onPrimary" | "dark";
+
+export type LoaderSizes = "xs" | "small" | "medium" | "large";
```
talkor, line 3: Is it used anywhere else? Do we actually need to export it?
It might be confusing if we export both

talkor (changes requested):
Great but let's just re-adjust the exported types

packages/core/src/components/Loader/LoaderConstants.ts
rivka-ungar: I added it in the types file, but forgot to remove it from here. Removed now

packages/core/src/__tests__/__snapshots__/exports.test.ts.snap
rivka-ungar: Removed

packages/core/src/components/Loader/Loader.types.ts
rivka-ungar: Done

talkor (approved):
Awesome :)

packages/core/src/components/Loader/Loader.types.ts
rivka-ungar: @talkor I think for now we should just export it as it is used also for the const in the component and when we combine all the size types we will anyhow get rid of it. wdyt?

talkor: NP

PR #2302, opened 2024-08-04
chore(TextField): deprecate enums

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7021052006

### DIFF
--- a/packages/core/src/components/TextField/TextField.types.ts
+++ b/packages/core/src/components/TextField/TextField.types.ts
@@ -1,0 +1,14 @@
+export type TextFieldTextType =
+  | "text"
+  | "password"
+  | "search"
+  | "date"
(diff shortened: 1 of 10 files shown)


packages/core/src/components/TextField/TextField.types.ts
```diff
@@ -0,0 +1,14 @@
+export type TextFieldTextType =
```
talkor, line 1: If it's not too much trouble I think it would be better to be called TextFieldType, but not critical

talkor (approved):

packages/core/src/components/TextField/TextField.types.ts
rivka-ungar: Done

PR #2305, opened 2024-08-04
chore(Dropdown): deprecate enums

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7021050216


packages/core/src/components/Dropdown/Dropdown.types.ts
```diff
@@ -274,3 +272,11 @@ export interface DropdownComponentProps extends CustomMenuBaseProps, CustomOptio
 }
 
 export type DropdownProps = DropdownComponentProps;
+
+export type DropdownChipColors = "PRIMARY" | "NEGATIVE" | "POSITIVE";
```
talkor, line 276: Can we change these to lowercase?

rivka-ungar: Yes, I will change it when I deprecate the chips enums

talkor (approved):

PR #2309, opened 2024-08-04
feat(Dropdown): remove xxs and xs sizes

talkor:
https://monday.monday.com/boards/3532714909/pulses/7139919911


rivka-ungar (approved):

TheMikeyRoss:
why? :(

talkor:
> why? :(

Hey @TheMikeyRoss! What's your concern? :) These sizes where an old mistake and they have no effect at the moment (they both behave just like the "small" variant). This change will also be available only in the next major version. If you have any other concern about this change let us know please

PR #2314, opened 2024-08-04
feat(Button): remove deprecated sizes support

talkor:
https://monday.monday.com/boards/3532714909/pulses/7139936914


packages/core/src/components/Button/ButtonConstants.ts
```diff
@@ -26,19 +26,6 @@ export const OLD_BUTTON_SIZES = {
 
 export type Size = (typeof SIZES)[keyof typeof SIZES] | keyof typeof OLD_BUTTON_SIZES;
 
-// Support old sizes (backward compatible)
-export const getActualSize = (size?: Size) => {
```
shaharzil, line 30: No need to add it to the change log?

talkor: I don't think anyone knew about it but will add anyway 😅

shaharzil (approved):

PR #2317, opened 2024-08-05
chore(chips): deprecate enums

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7021049553


packages/core/src/utils/colors-vars-map.ts
```diff
@@ -224,24 +224,37 @@ export enum ContentColorByName {
...
+ */
 export const ElementAllowedColor = {
   ...ContentColorByName,
   ...StateSelectedColor,
   ...StateSelectedHoverColor
 };
 
+/**
+ * @deprecated
+ */
```
talkor, line 257: Why? It's a type, I think we want it

packages/core/src/components/Chips/Chips.tsx
```diff
@@ -48,8 +49,7 @@ export interface ChipsProps extends VibeComponentProps {
   iconClassName?: string;
   /** ClassName for left or right avatar */
   avatarClassName?: string;
-  // TODO Vibe 3.0: filter ElementAllowedColor.DARK_INDIGO, ElementAllowedColor.BLACKISH from colors which are valid for Chips
-  color?: ElementColor;
+  color?: Exclude<ElementAllowedColor, "dark_indigo" | "blackish">;
```
talkor, line 52: We need to document it

packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.jsx
```diff
@@ -7,7 +7,6 @@ import Counter from "../../../Counter/Counter";
 import Dialog from "../../../Dialog/Dialog";
 import DialogContentContainer from "../../../DialogContentContainer/DialogContentContainer";
 import Chips from "../../../Chips/Chips";
-import { DROPDOWN_CHIP_COLORS } from "../../DropdownConstants";
```
talkor, line 10: Delete it or needed?

talkor (approved):

packages/core/src/components/Chips/Chips.tsx
rivka-ungar: Today it is excluded from the storybook, do we still need to document it?

packages/core/src/utils/colors-vars-map.ts
rivka-ungar: Since this type is using the deprecated enums, I created a new type that is exported. Should we still keep this one? I thought it would be confusing

packages/core/src/components/Chips/Chips.tsx
talkor: I think we do, if it can be used anyhow it's a valid api

packages/core/src/utils/colors-vars-map.ts
talkor: Oh got it, so let's keep it this way

packages/core/src/components/Dropdown/components/MultiValueContainer/MultiValueContainer.jsx
rivka-ungar: It is still in use

packages/core/src/components/Chips/Chips.tsx
rivka-ungar: Added

PR #2319, opened 2024-08-05
fix(menuButton): fix scale for disabled menu button

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/7080313485

### DIFF
--- a/packages/core/src/components/MenuButton/MenuButton.module.scss
+++ b/packages/core/src/components/MenuButton/MenuButton.module.scss
@@ -73,6 +73,7 @@
   border-color: var(--disabled-background-color);
   color: var(--disabled-text-color);
   cursor: not-allowed;
+  pointer-events: none;
 }
 
 .disabled:hover {


packages/core/src/components/MenuButton/MenuButton.module.scss
```diff
@@ -73,6 +73,7 @@
   border-color: var(--disabled-background-color);
   color: var(--disabled-text-color);
   cursor: not-allowed;
+  pointer-events: none;
```
YossiSaadi, line 76: I think it is a bit risky
what if we would just not apply the css for focus and hover in disabled mode?
```
.wrapper:focus:not(.disabled),
.wrapper:hover:not(.disabled) {
  background-color: var(--primary-background-hover-color);
}
```

YossiSaadi (approved):
after a talk decided to go with the pointer-events, since this is the same way we're using in Button, and we didn't had special issues

PR #2328, opened 2024-08-06
chore(Icon): Remove clickable & onClick props

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/6887644148


packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
```diff
@@ -4137,13 +4075,11 @@ exports[`Dropdown renders correctly without clear button 1`] = `
           onTouchEnd={[Function]}
         >
           <svg
-            aria-hidden={false}
-            className="icon clickable"
+            aria-hidden={true}
```
talkor, line 4078: Is that ok?

talkor (approved):
Awesome, just a question about the aria-hidden in Dropdown

packages/core/src/components/Dropdown/__tests__/__snapshots__/Dropdown.snapshot.test.js.snap
rivka-ungar: Currently it has tabIndex -1 and no aria-label, so it is not focusable and should have aria-hidden

PR #2335, opened 2024-08-07
feat(MenuItem): Add all label capabilities to menu item

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7077366107

### DIFF
--- a/packages/core/src/components/Menu/MenuItem/MenuItem.tsx
+++ b/packages/core/src/components/Menu/MenuItem/MenuItem.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/jsx-props-no-spreading */
-import React, { AriaAttributes, ForwardedRef, ReactElement, forwardRef, useMemo, useRef } from "react";
+import React, { AriaAttributes, ForwardedRef, ReactElement, forwardRef, useMemo, useRef, useCallback } from "react";
 import { DialogPosition } from "../../../constants";
 import Tooltip, { TooltipProps } from "../../../components/Tooltip/Tooltip";
 import Icon from "../../../components/Icon/Icon";
@@ -17,7 +17,7 @@ import { SubmenuPosition } from "./MenuItem.Types";
 
 export interface MenuItemProps extends VibeComponentProps {
   title?: string;
-  label?: string;
+  label?: string | React.ReactElement<typeof Label>;
   icon?: SubIcon;
   iconType?: IconType;
   iconBackgroundColor?: string;
@@ -116,6 +116,16 @@ const MenuItem: VibeComponent<MenuItemProps | MenuItemTitleComponentProps> & {
       return title;
     }, [disableReason, disabled, title, tooltipContent]);
 
+    const renderLabel = useCallback(() => {
+      if (!label) return;
+      if (typeof label === "string") {
+        return <Label kind={Label.kinds.LINE} text={label} />;
+      }
+      if (React.isValidElement(label) && label.type === Label) {
+        return label;
+      }
+    }, [label]);
+
     return (
       <BaseMenuItem
         key={key}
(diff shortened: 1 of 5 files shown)


packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -116,6 +116,16 @@ const MenuItem: VibeComponent<MenuItemProps | MenuItemTitleComponentProps> & {
...
 
+    const renderLabel = useCallback(() => {
+      if (!label) return;
+      if (typeof label === "string") {
+        return <Label kind={Label.kinds.LINE} text={label} />;
+      }
+      if (React.isValidElement(label) && label.type === Label) {
+        return label;
+      }
+    }, [label]);
```
YossiSaadi, line 127: I think `useMemo` can be a better candidate here
and you can render it afterwards without calling the function, just as {renderLabel}

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -17,7 +17,7 @@ import { SubmenuPosition } from "./MenuItem.Types";
 
 export interface MenuItemProps extends VibeComponentProps {
   title?: string;
-  label?: string;
+  label?: string | React.ReactElement<typeof Label>;
```
YossiSaadi, line 20: that's a great solution! I initially thought about adding a labelProps, but that's way better if we already want to allow everything!

packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
```diff
@@ -67,6 +68,26 @@ describe("<MenuItem />", () => {
...
+    await waitFor(() => expect(labelElement).toBeTruthy());
+  });
+
+  it("should render the Label component with props when pass a component", async () => {
+    const labelText = "Label Text";
+    const { getByTestId } = render(
+      <MenuItem title={title} label={<Label text={labelText} color="dark" kind="line" />} />
+    );
+    const labelElement = getByTestId("label");
+    await waitFor(() => expect(labelElement).toBeTruthy());
```
YossiSaadi, line 84: is `waitFor` needed here?

packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
```diff
@@ -67,6 +68,26 @@ describe("<MenuItem />", () => {
...
+    const { getAllByTestId } = within(labelElement);
+    const labelTextElement = getAllByTestId("text")[0];
```
YossiSaadi, line 86: does it have more than one test id of text? if so, we should take a look into it, sound like a bug

YossiSaadi (commented):
had one suggestion.
great solution!

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
rivka-ungar: You're right, I changed it

packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
rivka-ungar: Nope, removed!

YossiSaadi (approved):
Awesome job!

packages/core/src/components/Menu/MenuItem/__tests__/MenuItem.test.js
YossiSaadi: (not in this PR of course)

rivka-ungar: Yes, In the label there are 2 nested text elements, not sure why maybe for the animation?

PR #2338, opened 2024-08-08
chore re-organized types

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7160133818


packages/core/src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -11,10 +11,11 @@ import InfoIcon from "../Icon/Icons/components/Info";
 import { IconType as IconTypeEnum } from "../Icon/IconConstants";
 import { AttentionBoxType as AttentionBoxTypeEnum } from "./AttentionBoxConstants";
 import { AttentionBoxType } from "./AttentionBox.types";
-import { SubIcon, VibeComponentProps, withStaticProps, ElementContent } from "../../types";
+import { VibeComponentProps, withStaticProps, ElementContent } from "../../types";
 import Text from "../Text/Text";
 import Flex from "../Flex/Flex";
 import styles from "./AttentionBox.module.scss";
+import { SubIcon } from "../Icon";
```
YossiSaadi, line 18: is it out of "types" in purpose?

packages/core/src/components/Avatar/Avatar.tsx
```diff
@@ -5,15 +5,16 @@ import cx from "classnames";
...
 import { AvatarSize as AvatarSizeEnum, AvatarType as AvatarTypeEnum } from "./AvatarConstants";
 import { AvatarSize, AvatarType } from "./Avatar.types";
 import { AvatarBadge, AvatarBadgeProps } from "./AvatarBadge";
 import { AvatarContent } from "./AvatarContent";
 import Tooltip, { TooltipProps } from "../Tooltip/Tooltip";
 import ClickableWrapper from "../Clickable/ClickableWrapper";
-import { SubIcon, VibeComponentProps, withStaticProps } from "../../types";
+import { VibeComponentProps, withStaticProps } from "../../types";
 import styles from "./Avatar.module.scss";
+import { SubIcon } from "../Icon";
```
YossiSaadi, line 17: same, and for the rest usages

packages/core/src/components/Box/Box.types.ts
YossiSaadi, line 1: it is a bit weird it is inside Box, isn't it? because in theory it is a very general type (we should maybe try to make this kind of general types file, to save us redundant re-writing)

packages/core/src/components/Chips/Colors.types.ts
YossiSaadi, line 1: this file also exports a function, do you think it can benefit to have a utils file as well and send it to there? so we would better separate types from utils

packages/core/src/components/Icon/Icon.tsx
```diff
@@ -3,11 +3,11 @@ import cx from "classnames";
 import React, { CSSProperties, forwardRef, Ref } from "react";
 import useMergeRef from "../../hooks/useMergeRef";
 import { IconType as IconTypeEnum } from "./IconConstants";
-import { IconType } from "./Icon.types";
+import { IconType, MouseEventCallBack, SubIcon } from "./Icon.types";
```
YossiSaadi, line 6: also sound like something very generic that should be in types rather in Icon.types (MouseEventCallBack)

packages/core/src/components/Icon/Icon.types.ts
```diff
@@ -1 +1,9 @@
+import React from "react";
+import { IconSubComponentProps } from "./Icon";
+
 export type IconType = "svg" | "font" | "src";
+
+// Custom type for all mouse event callbacks
+export type MouseEventCallBack = (event: React.MouseEvent<HTMLElement | SVGElement>) => void;
```
YossiSaadi, line 7: I think it shouldn't be a different type. I think this should be inlined:
```
onClick?: (event: React.MouseEvent<HTMLElement | SVGElement>) => void;
```

YossiSaadi: since it is being used only once I don't see much value in making a type for it

packages/core/src/components/Icon/__tests__/allIcons.test.ts
```diff
@@ -7,9 +6,10 @@ import {
   getClickableScreenReaderAccessProps,
   getClickableIconScreenReaderAccessProps
 } from "../../../helpers/screenReaderAccessHelper";
+import { ArrayLastElement, Helpers } from "./helpers";
 
 type IconPath = `/icons/${keyof typeof allIcons}`;
-type SplittedIconPath = SplitString<IconPath, "/">;
+type SplittedIconPath = Helpers<IconPath, "/">;
```
YossiSaadi, line 12: is this intended? should it be called Helpers? I don't understand what exactly the helpers type does, but maybe a more meaningful name like SplitString can better explain what it does

YossiSaadi: I'm not actually familiar with what's going on there, but Helpers sound too generic

packages/core/src/components/Toggle/Toggle.types.ts
```diff
@@ -1 +1 @@
-export type Sizes = "small" | "medium";
+export type ToggleSizes = "small" | "medium";
```
YossiSaadi, line 1:
```suggestion
export type ToggleSize = "small" | "medium";
```
wdyt?

packages/core/src/types/ArrayLastElement.ts
YossiSaadi, line 1: was it unused?

YossiSaadi (commented):
I'm not sure about some of the stuff
let's talk about it, unless it was already discussed and decided

packages/core/src/components/Icon/Icon.types.ts
YossiSaadi: maybe even better:
```
onClick?: MouseEventHandler<HTMLElement | SVGElement>
```

PR #2339, opened 2024-08-08
chore(Tooltip): keyboard a11y by default

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7157053422


packages/core/docs/vibe-3-changelog.md
```diff
@@ -218,6 +218,7 @@ codemod: `avatar-component-migration`
 - `position` changed to only accept "top, right, bottom, left" [codemod]
 - `justify` removed [codemod]
 - `arrowPosition` removed [codemod]
+- `addKeyboardHideShowTriggersByDefault` make default true [codemod]
```
YossiSaadi, line 221: what should the codemod do? 
- remove the prop if exists?
- add it as false if it does not exist?

YossiSaadi (approved):
Awesome! had a question regarding the codemod behavior, if you can explain there

PR #2342, opened 2024-08-08
chore: change domain

talkor:
https://monday.monday.com/boards/3532714909/pulses/7177869985


packages/storybook-blocks/.storybook/theme.js
```diff
@@ -4,7 +4,7 @@ import logo from './logo.png';
 export default create({
   base: 'light',
   brandImage: logo,
-  brandUrl: 'https://style.monday.com', // TODO set brand url
+  brandUrl: 'https://mondaycom.github.io/vibe-storybook-components',
```
YossiSaadi, line 7: Do we publish new versions there? I'm not sure, I think we only publish core

YossiSaadi (approved):
Holy.. 🥹
Exciting!

I think we should update the monday developers docs regarding the change of both repo name and the new link (even though both changes are redirected from original to new).

packages/storybook-blocks/.storybook/theme.js
talkor: Oh yeah we do but we use `https://style.monday.com/storybook-blocks`, I'll change it, good catch

YossiSaadi: had no idea, lol
we should find how it's done and apply this logic for the /v2 docs (can also be nice to create one for /v1 on the way, one-time deploy)

YossiSaadi:
btw, in continuation to that one
we should also update the monday developers docs regarding the @vibe changes

talkor:
> btw, in continuation to that one we should also update the monday developers docs regarding the @vibe changes

Good point! Will make sure to update these

PR #2345, opened 2024-08-11
chore: remove backwards sizes support

talkor:
https://monday.monday.com/boards/3532714909/pulses/7202857183


YossiSaadi (approved):
I swear that each PR like this one gets me excited 😅

PR #2346, opened 2024-08-11
build: fix targets for storybook

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7202885063


YossiSaadi (approved):
bummer. sorry for that.
did it fixed the issue? we need to check why internal tools for storybook need vibe's storybook config

PR #2348, opened 2024-08-11
chore: remove enumPropNamesArray

talkor:
No longer needed since we use types, the props table is now showing the types instead of enums
https://monday.monday.com/boards/3532714909/pulses/7213849746


rivka-ungar (approved):
So much cleaner now!

PR #2354, opened 2024-08-12
chore(Modal): revert removal of unmountOnClose prop

talkor:
https://monday.monday.com/boards/3532714909/pulses/6887875560


packages/core/src/components/Modal/Modal.tsx
```diff
@@ -99,6 +99,7 @@ const Modal: FC<ModalProps> & { width?: typeof ModalWidthEnum } = ({
   closeButtonAriaLabel = "Close",
   contentSpacing = false,
   zIndex = 10000,
+  unmountOnClose = true,
```
shaharzil, line 102: I can't comment on line 85 but maybe we need to change the comment explaining about this prop

shaharzil (approved):

PR #2355, opened 2024-08-12
docs: fix type checking in storybook

talkor:
https://monday.monday.com/boards/3532714909/pulses/6272877557
For example:
Screenshot 2024-08-12 at 21 11 22

It will also fail storybook build and prevent releasing corrupted versions


packages/core/src/hooks/useSwitch/index.ts
talkor, line 1: @YossiSaadi I had to revert this change, there were serious type issues otherwise. Maybe we can find other solution but I'm also ok with how it is now

packages/core/src/components/Steps/StepsHeader.tsx
```diff
@@ -15,9 +15,9 @@ export interface StepsHeaderProps extends VibeComponentProps {
   onChangeActiveStep: (e: React.MouseEvent, stepIndex: number) => void;
   stepsCount: number;
   areNavigationButtonsHidden: boolean;
-  backButtonProps: ButtonProps;
-  nextButtonProps: ButtonProps;
-  finishButtonProps: ButtonProps;
+  backButtonProps: Partial<ButtonProps>;
+  nextButtonProps: Partial<ButtonProps>;
+  finishButtonProps: Partial<ButtonProps>;
```
talkor, line 20: I changed it in a different PR since Button's children is now mandatory but I think making it a Partial is a better solution instead of providing `children: ''` as it already have defaults

packages/core/src/components/Skeleton/__stories__/Skeleton.stories.tsx
```diff
@@ -12,41 +12,37 @@ export default {
...
 const skeletonTemplate = (args: SkeletonProps) => {
+  return <Skeleton {...args} />;
+};
+
+export const Overview = {
+  render: skeletonTemplate.bind({}),
+  name: "Overview"
+};
+
+export const ComplexExample = () => {
```
talkor, line 23: Had to extract the overview example to a different story to make the overview story a real playable example, need to verify with designers about the new story

packages/core/src/hooks/useSwitch/index.ts
YossiSaadi: Not sure I understand what exactly was reverted here 🫣

packages/core/.storybook/main.ts
```diff
@@ -39,6 +39,12 @@ const config: StorybookConfig = {
   docs: {
     autodocs: false
   },
+  typescript: {
```
YossiSaadi, line 42: It read that simple? 🤦🏻‍♂️

packages/core/src/hooks/useSwitch/index.ts
talkor: [This](https://github.com/mondaycom/vibe/pull/2341)

packages/core/.storybook/main.ts
talkor: Looks like it haha

packages/core/src/components/Skeleton/__stories__/Skeleton.stories.tsx
shaharzil, line 1: Maybe add the live edit if we're already touching this file?

shaharzil (approved):

PR #2357, opened 2024-08-13
ci(prerelease): use commit sha for prerelease version naming

talkor:
https://monday.monday.com/boards/3532714909/pulses/7086730463
https://monday.monday.com/boards/3532714909/pulses/7086725168

Screenshot 2024-08-15 at 16 42 02

It looks much cleaner IMO, and the purpose of such version is clearer (alpha versions). Plus it's based on the commit sha so it'll be unique and can be published multiple times per PR

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -35,7 +35,7 @@ jobs:
         id: generate-versions
         run: |
           commit_sha=$COMMIT_SHA
-          preid="test-${commit_sha:0:5}"
+          preid="alpha-${commit_sha:0:5}"
 
           full_output=$(yarn lerna version --exact --conventional-commits --conventional-prerelease --json --no-changelog --no-push --preid "$preid" -y)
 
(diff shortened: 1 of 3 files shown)


.github/workflows/prerelease.yml
```diff
@@ -31,10 +34,10 @@ jobs:
       - name: Generate new versions
         id: generate-versions
         run: |
-          preid=${{ github.ref_name }}
-          normalized_pre_id=$(echo "$preid" | tr -dc '[:alnum:]\n\r' | tr '[:upper:]' '[:lower:]')
+          commit_sha=$COMMIT_SHA
+          preid="test-${commit_sha:0:5}"
```
shaharzil, line 38: Is it common to use the word "test" for prerelease versions? I think it is a bit weird

talkor: It's not common, the most common pre-releases names are `alpha`, `beta` and `rc`. According to the [semver](https://semver.org/#spec-item-9) specs it's valid to have any alphanumeric values, so it's up to us. I think it's way better than using the branch name, especially when it can become long, unclear, and confusing. We can decided on some other structure but I don't think the branch name is a good choice
Screenshot 2024-08-14 at 14 50 40

talkor: @shaharzil changed to `alpha`

shaharzil (approved):

PR #2359, opened 2024-08-14
chore(Box): deprecate enums

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7225220262

### DIFF
--- a/packages/core/src/components/Box/Box.types.ts
+++ b/packages/core/src/components/Box/Box.types.ts
@@ -4,10 +4,6 @@ export type Border = "border";
 
 export type BorderColor = "borderColorUiBorderColor" | "borderColorLayoutBorderColor";
 
-export type Rounded = "roundedSmall" | "roundedMedium" | "roundedBig";
-
-export type Shadow = "shadowXs" | "shadowSmall" | "shadowMedium" | "shadowLarge";
-
 export type BackgroundColor =
   | "bgPrimaryBackgroundColor"
   | "bgSecondaryBackgroundColor"
@@ -20,3 +16,7 @@ export type BoxTextColor = "textPrimaryTextColor" | "textTextColorOnInverted" | 
 export type BoxPropSize = "xs" | "small" | "medium" | "large" | "xl" | "xxl" | "xxxl";
 
 export type BoxPropSizeWithAuto = "auto" | BoxPropSize;
+
+export type Shadow = Extract<BoxPropSize, "xs" | "small" | "medium" | "large">;
+
+export type Rounded = "small" | "medium" | "big";
--- a/packages/core/src/components/Box/Box.types.ts
+++ b/packages/core/src/components/Box/Box.types.ts
@@ -8,34 +8,6 @@ export type Rounded = "roundedSmall" | "roundedMedium" | "roundedBig";
 
 export type Shadow = "shadowXs" | "shadowSmall" | "shadowMedium" | "shadowLarge";
 
-export type Margin = "mAuto" | "mXs" | "mSmall" | "mMedium" | "mLarge" | "mXl" | "mXxl" | "mXxxl";
-
-export type MarginX = "mxAuto" | "mxXs" | "mxSmall" | "mxMedium" | "mxLarge" | "mxXl" | "mxXxl" | "mxXxxl";
-
-export type MarginY = "myAuto" | "myXs" | "mySmall" | "myMedium" | "myLarge" | "myXl" | "myXxl" | "myXxxl";
-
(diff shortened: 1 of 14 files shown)


packages/core/src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,22 @@
+export type Disabled = "opacityDisabled";
+
+export type Border = "border";
```
talkor, line 3: Wasn't it removed?

packages/core/src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,22 @@
+export type Disabled = "opacityDisabled";
```
talkor, line 1: What is this for? Can we avoid it? It doesn't seem to be used

packages/core/src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type Border = "border";
+
+export type BorderColor = "borderColorUiBorderColor" | "borderColorLayoutBorderColor";
```
talkor, line 5: Any chance we can get rid of the `border` prefix here? Not a priory but if it's small maybe it's worth it

packages/core/src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type BackgroundColor =
```
talkor, line 7: Same with `bg` prefix

packages/core/src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,22 @@
...
+  | "bgPrimaryBackgroundColor"
+  | "bgSecondaryBackgroundColor"
+  | "bgGreyBackgroundColor"
+  | "bgAllgreyBackgroundColor"
+  | "bgInvertedColorBackground";
+
+export type BoxTextColor = "textPrimaryTextColor" | "textTextColorOnInverted" | "textSecondaryTextColor";
```
talkor, line 14: Same with `text`

packages/core/src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type BoxPropSize = "xs" | "small" | "medium" | "large" | "xl" | "xxl" | "xxxl";
+
+export type BoxPropSizeWithAuto = "auto" | BoxPropSize;
```
talkor, line 18:
```suggestion
type BaseBoxSize = "xs" | "small" | "medium" | "large" | "xl" | "xxl" | "xxxl";

export type BoxSize = "auto" | BaseBoxSize;
```

packages/core/src/components/Box/Box.types.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type Shadow = Extract<BoxPropSize, "xs" | "small" | "medium" | "large">;
+
+export type Rounded = "small" | "medium" | "big";
```
talkor, line 22:
```suggestion
export type RoundedSize = "small" | "medium" | "big";
```

talkor (commented):
Amazing work!!
Had some questions/sugestion

packages/core/src/components/Box/Box.types.ts
rivka-ungar: Yep, forgot to remove it here

packages/core/src/components/Box/Box.types.ts
rivka-ungar: Yep, removed it

packages/core/src/components/Box/Box.types.ts
rivka-ungar: Yes!

packages/core/src/components/Box/Box.types.ts
rivka-ungar: removed

packages/core/src/components/Box/Box.types.ts
rivka-ungar: removed

packages/core/src/components/Box/Box.types.ts
rivka-ungar: The BaseBoxSize should not be exported?

talkor: Only if it's used for a certain prop

talkor (approved):
Excellent!

packages/core/src/components/Box/Box.types.ts
rivka-ungar: Yes, it is used by a bunch of the props

PR #2365, opened 2024-08-18
chore(Counter): codemod for migration of props

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7232289370


talkor (approved):
💯

PR #2373, opened 2024-08-19
Bring back static props that were deleted

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7203148935


talkor (approved):
Just need to adjust the changelog

PR #2378, opened 2024-08-21
Support small table row height

monadav:
Adding small (32px) table row height support

https://monday.monday.com/boards/3118265115/pulses/7270897414


YossiSaadi (changes requested):
```typescript
  padding-inline: var(--spacing-small);

  &.small {
    padding-block: var(--spacing-xs);
  }
```

please add that one to `TableRowMenu.module.scss`

YossiSaadi (approved):
Looking good! Thanks for contributing!

PR #2383, opened 2024-08-25
feat: layering floating components on modals

talkor:
https://monday.monday.com/boards/3532714909/pulses/5481337413


packages/core/.storybook/preview-body.html
```diff
@@ -1,2 +1 @@
 <div id="dropdown-portal"></div>
-<div id="tooltips-container"></div>
```
talkor, line 2: Leading principle: remove the need to render floatings (tooltips, tipseens, dialogs) in a certain container and render them in the body (with z-index givent to the `monday-style-dialog-content-wrapper` class which is used in DialogContent), unless specifically given "containerSelector" which will adjust the rendering to that node. On top of modals they will be rendered in the modal's node, meaning same z-index as the modal.

talkor: In a different PR we can export the layer provider along with instructions how to use it elsewhere

packages/core/src/components/Dialog/Dialog.tsx
```diff
@@ -180,7 +181,6 @@ export default class Dialog extends PureComponent<DialogProps, DialogState> {
     open: false,
     animationType: Dialog.animationTypes.EXPAND,
     preventAnimationOnMount: false,
-    containerSelector: "body",
```
talkor, line 183: It will fallback to the body already

packages/core/src/components/MenuButton/MenuButton.tsx
```diff
@@ -148,7 +148,7 @@ const MenuButton: VibeComponent<MenuButtonProps> & {
       dialogOffset = MOVE_BY,
       dialogPosition = Dialog.positions.BOTTOM_START,
       dialogClassName,
-      dialogPaddingSize = DialogContentContainer.sizes.MEDIUM,
+      dialogPaddingSize = DialogContentContainer.sizes.SMALL,
```
talkor, line 151: Notices we haven't changed it and it should have been fixed

packages/core/src/components/Modal/Modal.module.scss
```diff
@@ -27,7 +27,6 @@
 }
 
 .dialog {
-  z-index: 2;
```
talkor, line 30: Not needed, and makes additional z-index issues inside the modal which increases complexity for no reason

packages/core/src/components/Modal/Modal.tsx
```diff
@@ -163,32 +164,34 @@ const Modal: FC<ModalProps> & { width?: typeof ModalWidthEnum } = ({
...
 
   const dialog = ReactDOM.createPortal(
-    <div
-      {...attr.container}
-      className={cx(styles.container, classNames.container)}
-      data-testid={dataTestId || getTestId(ComponentDefaultTestId.MODAL, id)}
-      style={{ "--monday-modal-z-index": zIndex }}
-    >
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
+    <LayerProvider layerRef={{ current: instance?.$el }}>
```
talkor, line 167: This is how the `a11y-dialog` gives the ref 🤷🏼‍♂️

packages/core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -174,10 +173,6 @@ export default class Tooltip extends PureComponent<TooltipProps> {
     this.wasShown = false;
   }
 
-  getContainer() {
```
talkor, line 177: It can be provided externally but no need to adjust it to tooltips-container here anymore

packages/core/docs/vibe-3-changelog.md
talkor, line 1: The changes are documented in a different PR of the migration guide

shaharzil (approved):

PR #2384, opened 2024-08-25
chore: Codemod infra for changing a prop value

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7266875176
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7268604397
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7291311198


packages/codemod/src/utils/prop-utils.ts
```diff
@@ -123,6 +150,29 @@ export function migratePropsNames(
...
 
+export function updatePropValues(
+  j: JSCodeshift,
+  elementPath: ASTPath<JSXElement>,
+  propName: string,
+  valuesMapping: Record<
+    string,
+    {
+      value: string | number | boolean;
+      type: typeof memberExpression | typeof literal;
```
YossiSaadi, line 161:
```suggestion
      type: MemberExpression | Literal
```
?

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -79,12 +81,37 @@ function getPropValue(j: JSCodeshift, prop: JSXAttribute): undefined | boolean |
       // e.g. <Button text={true} /> or <Button text={"text"} />
       return value.expression.value;
     }
-  }
 
+    if (value.expression.type === "MemberExpression") {
+      // e.g. <Button border={Box.borders.DEFAULT} />
+      return j(value.expression).toSource();
+    }
```
YossiSaadi, line 88: awesome!

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -123,6 +150,29 @@ export function migratePropsNames(
   });
 }
 
+export function updatePropValues(
```
YossiSaadi, line 153:
```suggestion
export function updatePropsValue(
```
or
```suggestion
export function updatePropsValues(
```

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -79,12 +81,37 @@ function getPropValue(j: JSCodeshift, prop: JSXAttribute): undefined | boolean |
...
+  newValue: {
+    value: string | number | boolean;
+    type: typeof memberExpression | typeof literal;
+  }
+): void {
+  if (typeof newValue.value !== "string") {
+    const newValueIsTrue = typeof newValue.value === "boolean" && newValue.value;
+    attributePath.node.value = newValueIsTrue ? null : j.jsxExpressionContainer(j.literal(newValue.value));
+  } else {
+    if (newValue.type === memberExpression) {
```
YossiSaadi, line 106:
```suggestion
    if (newValue.type === "MemberExpression") {
```

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -79,12 +81,37 @@ function getPropValue(j: JSCodeshift, prop: JSXAttribute): undefined | boolean |
...
   // can be very complex, we'll have to compare strings
   return value ? j(value).toSource() : undefined;
 }
 
+export function setPropValue(
+  j: JSCodeshift,
+  attributePath: ASTPath<JSXAttribute>,
+  newValue: {
+    value: string | number | boolean;
+    type: typeof memberExpression | typeof literal;
```
YossiSaadi, line 99:
```suggestion
    type: MemberExpression | Literal
```

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -79,12 +81,37 @@ function getPropValue(j: JSCodeshift, prop: JSXAttribute): undefined | boolean |
...
+  }
+): void {
+  if (typeof newValue.value !== "string") {
+    const newValueIsTrue = typeof newValue.value === "boolean" && newValue.value;
+    attributePath.node.value = newValueIsTrue ? null : j.jsxExpressionContainer(j.literal(newValue.value));
+  } else {
+    if (newValue.type === memberExpression) {
+      const objectValue = j(`${newValue.value}`).find(j.ExpressionStatement).get().node.expression;
+      attributePath.node.value = j.jsxExpressionContainer(objectValue);
```
YossiSaadi, line 108: might worth to add a protection
something like 
```suggestion
      const objectValue = j(`${newValue.value}`).find(j.ExpressionStatement).get()?.node?.expression;
      attributePath.node.value = j.jsxExpressionContainer(objectValue);
```
or
```suggestion
      const expression = j(`${newValue.value}`).find(j.ExpressionStatement);
      if (!expression.get()) return;
      attributePath.node.value = j.jsxExpressionContainer(expression.get().node.expression);
```

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -79,12 +81,37 @@ function getPropValue(j: JSCodeshift, prop: JSXAttribute): undefined | boolean |
...
+    } else if (newValue.type === literal) {
```
YossiSaadi, line 109:
```suggestion
    } else if (newValue.type === "Literal") {
```

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -7,7 +7,9 @@ import {
   JSXExpressionContainer,
   JSXIdentifier,
   JSXOpeningElement,
+  literal,
   Literal,
+  memberExpression,
```
YossiSaadi, line 12: can use the MemberExpression for your needs instead of this builder

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -7,7 +7,9 @@ import {
   JSXExpressionContainer,
   JSXIdentifier,
   JSXOpeningElement,
+  literal,
```
YossiSaadi, line 10: can use the Literal for your needs instead of this builder

packages/codemod/plop/component/transform-update-props.hbs
```diff
@@ -19,7 +19,7 @@ function transform({ j, root, filePath }: TransformationContext) {
   if (!elements.length) return;
 
   elements.forEach(elementPath => {
-    migratePropsNames(j, elementPath, filePath, componentName, { {{#each propsMapping}}{{@key}}: "{{this}}", {{/each}} });
+    migratePropsNames(j, elementPath, filePath, componentName, { {{#each propsMapping}}{{@key}}: "{{this}}"{{#unless @last}}, {{/unless}}{{/each}} });
```
YossiSaadi, line 22: oh 😮

YossiSaadi (commented):
Awesome work!! I had a few minor suggestions

packages/codemod/src/utils/prop-utils.ts
rivka-ungar: The function updates one prop and many value, so not sure I understand why it should be changed?

packages/codemod/src/utils/prop-utils.ts
rivka-ungar: Added!

packages/codemod/src/utils/prop-utils.ts
YossiSaadi: oh, you are right. so same as for the static prop update comment - it shouldn't be a Record of a string to another Record, but rather to accept those inlined or in another structure. as the Record would always have one key and one value, right?

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -104,7 +104,7 @@ export function setPropValue(
     attributePath.node.value = newValueIsTrue ? null : j.jsxExpressionContainer(j.literal(newValue.value));
   } else {
     if (newValue.type === memberExpression) {
-      const objectValue = j(`${newValue.value}`).find(j.ExpressionStatement).get().node.expression;
+      const objectValue = j(`${newValue.value}`).find(j.ExpressionStatement).get()?.node?.expression;
       attributePath.node.value = j.jsxExpressionContainer(objectValue);
```
YossiSaadi: how does `j.jsxExpressionContainer(objectValue);` acts if `objectValue` is null or undefined? maybe early return here anyway if it is accidentally breaks

packages/codemod/src/utils/prop-utils.ts
rivka-ungar: The `valuesMapping` record can receive many values, the propName is a separate param, so that it has the option of updating several values on one prop.

YossiSaadi: if we're talking on the same prop it can only have one value at a time, so there would always by one key
or am I missing something? can you give an example?

rivka-ungar: Sure. For example I have a prop called size, and I want to update both the small and the medium values to be large.
I would call it like this:
`    updatePropValues(j, elementPath, "size", {
      small: {
        value: "large",
        type: j.Literal
      },
      medium: {
        value: "large",
        type: j.Literal
      }
    });`

YossiSaadi: Ohhh ok I got you
I wonder if this should be in a separate calls to `updatePropValues`
but its cool for now, let's discuss it when we encounter such use case in the future

thanks for clarifying!!

rivka-ungar: I assumed it would be more efficient in a case like this, so that you get the prop only once and then check if its current value matches any of the values that need to change. As supposed to checking it multiple times. Right now we don't have such a case, so do you think it will be better to just support one value and change it in the future if needed?

YossiSaadi: I'm not sure actually
We can further discuss it when such a need raises, for now I wouldn't change it and remove an ability

packages/codemod/transformations/core/v2-to-v3/box-component-migration.ts
```diff
@@ -22,7 +23,7 @@ function transform({ j, root }: TransformationContext) {
     updatePropValues(j, elementPath, "border", {
       "Box.borders.DEFAULT": {
         value: true,
-        type: j.memberExpression
+        type: MemberExpression
```
YossiSaadi, line 26: you can also take this one from j

packages/codemod/transformations/core/v2-to-v3/dialog-content-container-component-migration.ts
```diff
@@ -22,7 +23,7 @@ function transform({ j, root }: TransformationContext) {
     updatePropValues(j, elementPath, "size", {
       "DialogContentContainer.sizes.MEDIUM": {
         value: "DialogContentContainer.sizes.SMALL",
-        type: j.memberExpression
+        type: MemberExpression
```
YossiSaadi, line 26: same, from j, just to not import more stuff in the transformations

I wonder if we should do it also for all the utils, I didn't do it initially because I didn't had `j` in all the utils (didn't needed it, like in the getCoreImports)

YossiSaadi (approved):

PR #2390, opened 2024-08-26
Chore(ColorPicker/ColorPickerContent): Codemod for migration of static props

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7295552042


YossiSaadi (approved):
Awesome!

PR #2395, opened 2024-08-28
chore: codemods for /types entry

talkor:
https://monday.monday.com/boards/3532714909/pulses/7303260067

### DIFF
--- a/packages/codemod/transformations/core/v2-to-v3/__tests__/type-imports-migration.test.ts
+++ b/packages/codemod/transformations/core/v2-to-v3/__tests__/type-imports-migration.test.ts
@@ -17,4 +17,32 @@ describe("Rename type imports migration", () => {
     `import { VibeComponentProps } from "monday-ui-react-core";`,
     "should not modify the import path if it's already 'monday-ui-react-core'"
   );
+
+  defineInlineTest(
+    transform,
+    {},
+    `import { SomeOtherType } from "another-library";`,
+    `import { SomeOtherType } from "another-library";`,
+    "should not change unrelated imports"
+  );
+
+  defineInlineTest(
+    transform,
+    {},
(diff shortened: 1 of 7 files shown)


packages/codemod/types/paths.types.ts
```diff
@@ -1,3 +1,3 @@
-import { CORE_IMPORT_PATH, CORE_NEXT_IMPORT_PATH } from "../src/consts";
+import { CORE_IMPORT_PATH, CORE_NEXT_IMPORT_PATH, CORE_TYPES_IMPORT_PATH } from "../src/consts";
 
-export type VibeCorePath = typeof CORE_IMPORT_PATH | typeof CORE_NEXT_IMPORT_PATH;
+export type VibeCorePath = typeof CORE_IMPORT_PATH | typeof CORE_NEXT_IMPORT_PATH | typeof CORE_TYPES_IMPORT_PATH;
```
YossiSaadi, line 3: I think I made a mistake here
Should it be this instead?:
```suggestion
export type VibeCorePath = CORE_IMPORT_PATH | CORE_NEXT_IMPORT_PATH | CORE_TYPES_IMPORT_PATH;
```

packages/codemod/transformations/core/v2-to-v3/__tests__/type-imports-migration.test.ts
YossiSaadi, line 1: same as the packages renaming, is this enough tests?
e.g. doesn't change unrelated imports, changes multiple imports, only change required when both types and core are in use, etc.

I'm saying it since we should be extra careful imo with codemods, as a wrong change/failure is a critical issue for the flow of running all the codemods

YossiSaadi (approved):
Looks cool, I think we can benefit from more tests with everyday cases and some edge cases

PR #2396, opened 2024-08-28
chore: packages rename codemod

talkor:
https://monday.monday.com/boards/3532714909/pulses/7303260027
https://monday.monday.com/boards/3532714909/pulses/7303260029


YossiSaadi (approved):
Same here, I think we can benefit from more tests of common use cases and some specific edge cases that might occur

PR #2398, opened 2024-08-28
chore: Featured icon codemod

talkor:
https://monday.monday.com/boards/3532714909/pulses/7314583577

### DIFF
--- a/packages/codemod/transformations/core/v2-to-v3/featured-icon-migration.ts
+++ b/packages/codemod/transformations/core/v2-to-v3/featured-icon-migration.ts
@@ -1,0 +1,29 @@
+import { wrap, getComponentNameOrAliasFromImports, getImports } from "../../../src/utils";
+import { TransformationContext } from "../../../types";
+import { CORE_ICONS_IMPORT_PATH } from "../../../src/consts";
+
+/**
+ * 1. Replace Featured icon with Upgrade icon
+ */
+function transform({ j, root }: TransformationContext) {
+  const imports = getImports(root, CORE_ICONS_IMPORT_PATH);
+
+  if (!imports.length) return;
+
+  const componentName = getComponentNameOrAliasFromImports(j, imports, "Featured");
+  if (!componentName) return;
+
+  imports.forEach(path => {
+    path.node.specifiers?.forEach(spec => {
+      if (j.ImportSpecifier.check(spec)) {
+        if (spec.imported.name === "Featured") {
+          spec.imported.name = "Upgrade";
+        }
+      }
+    });
+  });
+
+  root.find(j.JSXExpressionContainer).find(j.Identifier, { name: componentName }).replaceWith(j.identifier("Upgrade"));
+}
+
+export default wrap(transform);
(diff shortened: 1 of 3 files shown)


packages/codemod/transformations/core/v2-to-v3/featured-icon-migration.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+function transform({ j, root }: TransformationContext) {
+  const imports = getImports(root, CORE_ICONS_IMPORT_PATH);
+
+  if (!imports.length) return;
+
+  const componentName = getComponentNameOrAliasFromImports(j, imports, "Featured");
+  if (!componentName) return;
+
+  imports.forEach(path => {
```
rivka-ungar, line 16: Would using `updateImportSpecifierName` work here instead?

packages/codemod/transformations/core/v2-to-v3/featured-icon-migration.ts
```diff
@@ -0,0 +1,29 @@
...
+    path.node.specifiers?.forEach(spec => {
+      if (j.ImportSpecifier.check(spec)) {
+        if (spec.imported.name === "Featured") {
+          spec.imported.name = "Upgrade";
+        }
+      }
+    });
+  });
+
+  root.find(j.JSXExpressionContainer).find(j.Identifier, { name: componentName }).replaceWith(j.identifier("Upgrade"));
```
rivka-ungar, line 26: Can use `updatePropValues` instead?

YossiSaadi, line 26: The following are stuff I **think** would happen, but worth adding as tests to make sure they do not occur:

this would replace also
```
<SomeComponent myProp={"Featured"} />
```
to this
```
<SomeComponent myProp={"Upgrade"} />
```

and also this
```
<SomeComponent myProp={Featured} />
```
to this
```
<SomeComponent myProp={Upgrade} />
```

basically everything that is inside a prop which is inside `{}` and includes the name Featured.
e.g.
```
<AnotherComponent thisIsAnObject={{ oneAttr: "someValue", Featured: true }} />
```

can you add tests with those cases and check it?

YossiSaadi: I think I get why he took that approach. 
`updatePropValues` is for a specific component.
We cannot know where `Upgrade` is being used, it can be used as a prop in a lot kinds of components that get `SubIcon` type as a prop

talkor:
@rivka-ungar @YossiSaadi I feel like it's too risky for a very non-important change, let's drop it?

packages/codemod/transformations/core/v2-to-v3/featured-icon-migration.ts
YossiSaadi: yep, exactly

YossiSaadi:
> @rivka-ungar @YossiSaadi I feel like it's too risky for a very non-important change, let's drop it?

I agree
We are releasing a great migration guide, writing it there as a breaking change is more than enough.
Once someone's code would fail I expect the dev to check the migration guide.
We should maybe write stuff which are breaking non-gracefully, like that one, at the first sections of the migration guide, the stuff that has codemod can be written at a later section

talkor:
Closing for now

PR #2404, opened 2024-08-29
chore(MenuButton): remove   showTooltipOnlyOnTriggerElement

talkor:


packages/core/src/components/MenuButton/__tests__/__snapshots__/MenuButton.test.tsx.snap
```diff
@@ -1,13 +1,8 @@
...
 
 exports[`MenuButton renders correctly with .dummy-class-name 1`] = `
-<button
-  aria-disabled={false}
-  aria-expanded={false}
-  aria-haspopup="true"
-  aria-label="Menu"
-  className="wrapper dummy-class-name sizeSmall"
-  data-testid="menu-button"
+<span
```
talkor, line 4: @YossiSaadi I'm not sure about this change, is it ok a11y wise? Do you remember?

packages/core/src/components/MenuButton/MenuButton.tsx
```diff
@@ -366,10 +361,7 @@ const MenuButton: VibeComponent<MenuButtonProps> & {
       </Tooltip>
     );
 
-    if (showTooltipOnlyOnTriggerElement) {
-      return dialogNode(tooltipNode(triggerElementNode));
-    }
-    return tooltipNode(dialogNode(triggerElementNode));
+    return dialogNode(tooltipNode(triggerElementNode));
```
YossiSaadi, line 364: you can inline it I guess and remove the consts, if I remember correctly I've only added them for `showTooltipOnlyOnTriggerElement` functionality

packages/core/src/components/MenuButton/__tests__/__snapshots__/MenuButton.test.tsx.snap
YossiSaadi: why did it happen?
I don't think there's an a11y issue here, but I'm not sure that it should have happened
can you compare to the snapshot before the `showTooltipOnlyOnTriggerElement` change? if the snapshot looked exactly as it is right now?

packages/codemod/transformations/core/v2-to-v3/tooltip-component-migration.ts
```diff
@@ -20,6 +21,7 @@ function transform({ j, root }: TransformationContext) {
 
   elements.forEach(elementPath => {
     removeProp(j, elementPath, "withMaxWidth");
+    removeProp(j, elementPath, "showTooltipOnlyOnTriggerElement");
```
YossiSaadi, line 24:
```suggestion
    removeProp(j, elementPath, "withMaxWidth", "showTooltipOnlyOnTriggerElement");
```

should also work, should save an iteration of `O(n^2)` (probably negligible, so for your consideration)

YossiSaadi (approved):
Looks cool! not sure about the snapshot, we should compare the snapshots before the `"showTooltipOnlyOnTriggerElement"` first introduced

PR #2411, opened 2024-09-04
 chore: change codemod to use tsx parser

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7359123505

### DIFF
--- a/packages/codemod/src/utils/prop-utils.ts
+++ b/packages/codemod/src/utils/prop-utils.ts
@@ -7,7 +7,6 @@ import {
   JSXExpressionContainer,
   JSXIdentifier,
   JSXOpeningElement,
-  Literal,
   MemberExpression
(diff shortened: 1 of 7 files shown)


packages/codemod/src/utils/prop-utils.ts
```diff
@@ -69,14 +68,18 @@ export function getPropValue(j: JSCodeshift, prop: JSXAttribute): undefined | bo
   }
 
   // string value (e.g. <Button text="Click me" />)
-  if (value?.type === "Literal") {
-    return value.value;
+  if (value?.type === "StringLiteral") {
+    return value?.value;
```
YossiSaadi, line 72: needed? if we're in StringLiteral I think it would always have `value.value`

YossiSaadi (approved):
💯

PR #2413, opened 2024-09-04
feat(ModalHeader): header component to be used in modal

YossiSaadi:
### ModalHeader 🌐

image

Props:
Name | Type | Description
-- | -- | --
title* | string | 
description | string \| React.ReactNode | 
descriptionIcon | SubIcon \| { name: SubIcon; className?: string } | Can only be passed when description is passed

Usage example:
```typescript
<ModalHeader
  title="I'm title"
  description="I'm description"
  descriptionIcon={Table}
/>

// or for more complex usages:
<ModalHeader
  title="I'm title"
  description={<Text>This can include <a>Everything</a></Text>}
  descriptionIcon={Table}
/>

// this ends up as ts-error
<ModalHeader
  title="I'm title"
  descriptionIcon={Table}
/>
```

https://monday.monday.com/boards/3532715121/pulses/7359892692


packages/core/src/components/ModalNew/ModalHeader/ModalHeader.module.scss
```diff
@@ -0,0 +1,14 @@
+.header {
+  width: 100%;
+  align-self: flex-start;
+
+  .title {
+    width: 100%;
```
rivka-ungar, line 6: Does the title need to leave space for the header actions to guarantee that it won't overlap as it is absolute?

packages/core/src/components/ModalNew/ModalHeader/ModalHeader.tsx
```diff
@@ -0,0 +1,52 @@
...
+        {description && (
+          <Flex>
+            {descriptionIcon && (
+              <Icon
+                icon={typeof descriptionIcon === "object" ? descriptionIcon.name : descriptionIcon}
+                className={cx(styles.descriptionIcon, typeof descriptionIcon === "object" && descriptionIcon.className)}
+              />
+            )}
+            {typeof description === "string" ? (
+              <Text element="span" type={Text.types.TEXT1}>
```
rivka-ungar, line 39: Just wondering why this needs to be a span if it is inside flex

packages/core/src/components/ModalNew/ModalHeader/ModalHeader.module.scss
YossiSaadi: Every layout would be responsible to manage the spacing eacb component needs, so the inline end would have padding on basic layout only

packages/core/src/components/ModalNew/ModalHeader/ModalHeader.tsx
YossiSaadi: Hmm good point, it wasn't initially a flex, but then I thought that people can pass node which would be a block and it would break the inline with the icon.
Do you think I should change it or leave it as is?

packages/core/src/components/ModalNew/ModalHeader/ModalHeader.module.scss
rivka-ungar: Got it

packages/core/src/components/ModalNew/ModalHeader/ModalHeader.tsx
rivka-ungar: It is not critical, so whatever you prefer :)

rivka-ungar (approved):

packages/core/src/components/ModalNew/ModalHeader/ModalHeader.tsx
YossiSaadi: Cool, if I'll see it creates an issue for some reason, I'll remove it.
Thanks!

YossiSaadi: Cool, if I'll see it creates an issue for some reason, I'll remove it.
Thanks!

PR #2417, opened 2024-09-05
feat(Modal): modal component with basic functionality

YossiSaadi:
### Modal 🌐 

image

This should be a canvas-like component that should only take care of the logic and not what to render inside (except the top actions).

Props:

Name | Type | Default | Description
-- | -- | -- | --
show* | boolean |   |  
size | 🌐 ModalSize: "small" \| "medium" \| "large" | "medium" |  
closeButtonTheme | ModalTopActionsProps["color"]: "light" \| "dark" |   | 
closeButtonAriaLabel | ModalTopActionsProps["closeButtonAriaLabel"]: string |   |  
onClose | ModalTopActionsProps["onClose"]: React.MouseEventHandler<HTMLDivElement> |   |  
renderHeaderAction | ModalTopActionsProps["renderAction"]: (color?: ModalTopActionsButtonColor) => React.ReactElement<typeof MenuButton \| typeof IconButton> |   |  
children | React.ReactNode |   | Can be anything.

Usage example:
```typescript
<Modal
  show={shouldShowModal}
  size="small"
  closeButtonTheme="light"
  closeButtonAriaLabel="Close"
  onClose={() => console.log("Closing")}
  renderHeaderAction={(color) => <IconButton icon={Feedback} color={color} />}
  triggerElementRef={buttonRef}
>
  {/* children */}
</Modal>
```

https://monday.monday.com/boards/3532715121/pulses/7360583166


packages/core/src/components/ModalNew/Modal/Modal.types.tsx
```diff
@@ -0,0 +1,15 @@
+import { VibeComponentProps } from "../../../types";
+import React from "react";
+import { ModalTopActionsProps } from "../ModalTopActions/ModalTopActions.types";
+
+export type ModalSize = "small" | "medium" | "large";
+
+export interface ModalProps extends VibeComponentProps {
+  show: boolean;
+  size?: ModalSize;
+  closeButtonTheme?: ModalTopActionsProps["color"];
```
YossiSaadi, line 10: @shlomitc suggests deciding on a different structure for this, that we would apply on all our components.
Right now the pattern is also "[prefix]Theme" for Tipseen for example. (also for close button theming).
let's discuss if we want to change it.

packages/core/src/components/ModalNew/Modal/Modal.module.scss
```diff
@@ -0,0 +1,12 @@
+.overlay {
+  position: fixed;
+  inset: 0;
+  background-color: rgba(41, 47, 76, 0.7);
```
talkor, line 4: Is it ok? I think we recently changed it to be more adjustable to the themes, no?

packages/core/src/components/ModalNew/Modal/Modal.types.tsx
talkor: Not sure I understand

talkor (approved):

packages/core/src/components/ModalNew/Modal/Modal.module.scss
YossiSaadi: I'll create a separate task for it

PR #2420, opened 2024-09-05
feat(Modal): use titleId and descriptionId for appropriate arias, allow setting those values through context

YossiSaadi:
### Implementing Modal's arias for title and description

- Set `aria-labelledby` for the Modal's title element id
- Set `aria-describedby` for the Modal's description element id

The reason for it to be a setter, rather than passing a plain value to the Context and then use it (without states), is because we cannot tell if a description would be provided to the Modal by consumer.
So in ModalHeader we should call the setters according to the ids supplied there.

See https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/#rps_label for further best practices regarding `Modal`'s arias

Next steps:
- Consume the context inside `ModalHeader`
- Set the id for the title and description elements inside `ModalHeader`, using the `Modal`'s `id` prop (to have unique ids):
	- `[my-modal-id]_label`
	- `[my-modal-id]_desc`

https://monday.monday.com/boards/3532715121/pulses/7367625236

### DIFF
--- a/packages/core/src/components/ModalNew/Modal/Modal.tsx
+++ b/packages/core/src/components/ModalNew/Modal/Modal.tsx
@@ -28,8 +28,8 @@ const Modal = forwardRef(
     }: ModalProps,
     ref: React.ForwardedRef<HTMLDivElement>
   ) => {
-    const [titleId, setTitleId] = useState<string>("");
-    const [descriptionId, setDescriptionId] = useState<string>("");
+    const [titleId, setTitleId] = useState<string>();
+    const [descriptionId, setDescriptionId] = useState<string>();
 
     const setTitleIdCallback = useCallback((id: string) => setTitleId(id), []);
     const setDescriptionIdCallback = useCallback((id: string) => setDescriptionId(id), []);
@@ -53,8 +53,8 @@ const Modal = forwardRef(
             data-testid={dataTestId || getTestId(ComponentDefaultTestId.MODAL, id)}
             role="dialog"
             aria-modal
-            aria-labelledby={titleId || undefined}
-            aria-describedby={descriptionId || undefined}
+            aria-labelledby={titleId}
+            aria-describedby={descriptionId}
           >
             <ModalTopActions
               renderAction={renderHeaderAction}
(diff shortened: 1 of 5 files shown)


YossiSaadi:
I'm really not 100% sure, and would love to get some feedback regarding that

packages/core/src/components/ModalNew/Modal/Modal.tsx
```diff
@@ -21,32 +24,48 @@ const Modal = forwardRef(
       onClose,
       children,
       className,
-      id,
       "data-testid": dataTestId
     }: ModalProps,
     ref: React.ForwardedRef<HTMLDivElement>
   ) => {
+    const [titleId, setTitleId] = useState<string>("");
+    const [descriptionId, setDescriptionId] = useState<string>("");
```
talkor, line 32: If you dont give it a default value you wouldn't need the `undefined` in lines 56-57, no?

packages/core/src/components/ModalNew/Modal/Modal.types.tsx
```diff
@@ -5,6 +5,7 @@ import { ModalTopActionsProps } from "../ModalTopActions/ModalTopActions.types";
 export type ModalSize = "small" | "medium" | "large";
 
 export interface ModalProps extends VibeComponentProps {
+  id: string;
```
talkor, line 8: Why is it now mandatory? Also, it won't guarantee uniqueness as users can pass same id to multiple modals

packages/core/src/components/ModalNew/Modal/Modal.tsx
YossiSaadi: it will set the attribute with empty string, I want it to not be shown at all
image
image

talkor: I meant useState<string>()

packages/core/src/components/ModalNew/Modal/Modal.types.tsx
YossiSaadi: because I don't have `useId` in React 16 😅
I wanted to generate an id for title and subtitle which are:
```
[my-modal-id]_label
[my-modal-id]_desc
```
I can just implement some generate id logic in ModalHeader as a substitute. wdyt?

talkor: Can it have a default without an id and if id is provided - use it for these ids?

packages/core/src/components/ModalNew/Modal/Modal.tsx
YossiSaadi: oh
good point, lol

PR #2422, opened 2024-09-08
feat(Dropdown): add option for divider in dropdown

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7178692867

Adding option for divider in dropdown

Screenshot 2024-09-08 at 15 40 44

### DIFF
--- a/packages/core/src/components/Dropdown/Dropdown.styles.js
+++ b/packages/core/src/components/Dropdown/Dropdown.styles.js
@@ -371,11 +371,10 @@ const group =
           content: '""',
           position: "absolute",
           bottom: "0",
-          left: "50%",
-          transform: "translateX(-50%)",
-          width: `calc(100% - (${getCSSVar("spacing-small")} * 2))`,
+          left: "var(--spacing-small)",
+          width: "calc(100% - (var(--spacing-small) * 2))",
           height: "1px",
-          backgroundColor: getCSSVar("layout-border-color")
+          backgroundColor: "var(--layout-border-color)"
         }
       }
     })
--- a/packages/core/src/components/Dropdown/Dropdown.styles.js
+++ b/packages/core/src/components/Dropdown/Dropdown.styles.js
@@ -373,7 +373,7 @@ const group =
           bottom: "0",
           left: "50%",
           transform: "translateX(-50%)",
-          width: "93%",
+          width: `calc(100% - (${getCSSVar("spacing-small")} * 2))`,
           height: "1px",
           backgroundColor: getCSSVar("layout-border-color")
         }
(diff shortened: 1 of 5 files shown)


packages/core/src/components/Dropdown/Dropdown.styles.js
```diff
@@ -358,15 +358,32 @@ const indicatorSeparator = () => () => ({
...
+    ...(withGroupDivider && {
+      ":not(:last-child)": {
+        position: "relative",
+        paddingBottom: "8px",
+        "::after": {
+          content: '""',
+          position: "absolute",
+          bottom: "0",
+          left: "50%",
+          transform: "translateX(-50%)",
```
YossiSaadi, line 375:
```suggestion
          left: var(--spacing-small),
```

Can also work

packages/core/src/components/Dropdown/Dropdown.styles.js
```diff
@@ -358,15 +358,32 @@ const indicatorSeparator = () => () => ({
...
+          width: `calc(100% - (${getCSSVar("spacing-small")} * 2))`,
```
YossiSaadi, line 376:
```suggestion
          width: `calc(100% - (var(--spacing-small) * 2))`,
```
I'm not 100% sure about css-in-js, but shouldn't this work?

packages/core/src/components/Dropdown/Dropdown.styles.js
```diff
@@ -358,15 +358,32 @@ const indicatorSeparator = () => () => ({
...
+          height: "1px",
+          backgroundColor: getCSSVar("layout-border-color")
```
YossiSaadi, line 378:
```suggestion
          backgroundColor: "var(layout-border-color)"
```
same as the above comments, shouldn't this work?

packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -110,7 +110,8 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
       popupsContainerSelector,
       filterOption,
       menuPosition = Dropdown.menuPositions.ABSOLUTE,
-      "data-testid": dataTestId
+      "data-testid": dataTestId,
+      withGroupDivider = false
```
YossiSaadi, line 114: do we have any way to put it in the supplied options array instead of as a prop?

YossiSaadi (commented):
nice solution, had a few thoughts and questions, lmk what you think

packages/core/src/components/Dropdown/Dropdown.tsx
rivka-ungar: We wanted to keep it similar to the combobox api that has `withCategoriesDivider`. Also if it is in the options array will have to send the prop in each group separately.

YossiSaadi: oh you're right
good call 💪🏼

YossiSaadi (approved):
💯

PR #2425, opened 2024-09-09
Codemod update prop name

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7322266599


packages/codemod/src/utils/__tests__/prop-utils.test.ts
```diff
@@ -28,6 +28,18 @@ describe("Prop Utils", () => {
...
+        source: `<Component someProp="value"><div someOtherProp="value"></div></Component>`,
+        propName: "someOtherProp",
+        expected: false,
+        description: "should return false if prop doesn't exist on parent element"
+      },
+      {
+        source: `<Component someOtherProp="value"><div someOtherProp="value"></div></Component>`,
+        propName: "someOtherProp",
+        expected: true,
+        description: "should return true if prop exists on parent element"
```
YossiSaadi, line 42: can you have one also for the opposite case? like the top two, but with nested
```suggestion
      {
        source: `<Component someProp="value"><div someOtherProp="value"></div></Component>`,
        propName: "someOtherProp",
        expected: false,
        description: "should return false if prop doesn't exist on parent element"
      },
      {
        source: `<Component someProp="value"><div someOtherProp="value"></div></Component>`,
        propName: "someProp",
        expected: true,
        description: "should return true if prop exists on parent element"
      },
      {
        source: `<Component someOtherProp="value"><div someOtherProp="value"></div></Component>`,
        propName: "someOtherProp",
        expected: true,
        description: "should return true if prop exists on parent element"
```

packages/codemod/src/utils/prop-utils.ts
```diff
@@ -15,25 +15,26 @@ import { logPropMigrationError } from "./report-utils";
...
-    .find(JSXOpeningElement)
-    .find(JSXIdentifier)
-    .forEach(identifierPath => {
-      if (propsNamesMappingOldToNew[identifierPath.node.name]) {
-        identifierPath.node.name = propsNamesMappingOldToNew[identifierPath.node.name];
-      }
-    });
+  propCollection.forEach(attr => {
+    const propName = attr.node.name.name;
+    if (typeof propName !== "string") return;
```
YossiSaadi, line 23: we should tackle this one day and check throughly on which cases we don't have `.name.name` as a string (out of scope for current phase IMO)

YossiSaadi (approved):
Awesome! 💯

packages/codemod/src/utils/__tests__/prop-utils.test.ts
rivka-ungar: Added

PR #2429, opened 2024-09-10
feat(Modal): handle onClose and show logic of component

YossiSaadi:
### `onClose` and `show` logic

- Also fix types for `Modal` and `ModalTopActions` to be more specific for each use case

https://monday.monday.com/boards/3532715121/pulses/7368472599


packages/core/src/components/ModalNew/ModalTopActions/ModalTopActions.types.ts
```diff
@@ -13,5 +13,5 @@ export interface ModalTopActionsProps {
   renderAction?: (color?: ModalTopActionsButtonColor) => React.ReactElement<typeof MenuButton | typeof IconButton>;
   color?: ModalTopActionsColor;
   closeButtonAriaLabel?: string;
-  onClose?: React.MouseEventHandler<HTMLDivElement>;
+  onClose?: (e: React.MouseEvent<HTMLButtonElement>) => void;
```
talkor, line 16: Nitpick: Why just e and not event?

packages/core/src/components/ModalNew/Modal/Modal.tsx
```diff
@@ -43,27 +42,57 @@ const Modal = forwardRef(
...
+        if (!show) return;
+        onClose(e);
+      },
+      [onClose, show]
+    );
+
+    useKeyEvent({
+      callback: onEscClick,
+      capture: true,
+      keys: ["Escape"]
```
talkor, line 64: I think we have consts for these

packages/core/src/components/ModalNew/Modal/__tests__/Modal.test.tsx
```diff
@@ -27,6 +28,16 @@ describe("Modal", () => {
     expect(getByTestId("modal")).toHaveAttribute("aria-modal", "true");
   });
 
+  it("does not render when 'show' is false", () => {
```
talkor, line 31:
```suggestion
  it("should not render when 'show' is false", () => {
```

talkor (approved):

packages/core/src/components/ModalNew/Modal/__tests__/Modal.test.tsx
YossiSaadi: I've wrote all the tests not in a `should` style (by mistake, accidentally ignored Gebeta's chosen style)
planning on changing all the tests names later 😓

packages/core/src/components/ModalNew/ModalTopActions/ModalTopActions.types.ts
YossiSaadi: hmm no special reason
will change it

packages/core/src/components/ModalNew/Modal/Modal.tsx
YossiSaadi: Will change for now, but I think we should make `useKeyEvent` use string keys as well, same as we switched from static props to inline string props

can be a non-breaking change after we release Vibe3

talkor: But it's not type-safe, no?

YossiSaadi: it can be
we can extract the keys strings from React's TS I believe

PR #2434, opened 2024-09-12
feat(Modal): keep focus inside opened modal, return focus to last element

YossiSaadi:
### Focus lock

- Focus should be locked inside Modal and to not let "Tab" click to focus stuff outside of the Modal
https://monday.monday.com/boards/3532714909/pulses/7360933300

- Focus should return to last element that was opened before modal opened
https://monday.monday.com/boards/3532715121/pulses/7391253003

### DIFF
--- a/packages/core/src/components/ModalNew/Modal/Modal.tsx
+++ b/packages/core/src/components/ModalNew/Modal/Modal.tsx
@@ -1,6 +1,7 @@
 import React, { forwardRef, useCallback, useMemo, useState } from "react";
 import cx from "classnames";
 import { RemoveScroll } from "react-remove-scroll";
+import FocusLock from "react-focus-lock";
 import { getTestId } from "../../../tests/test-ids-utils";
 import { ComponentDefaultTestId } from "../../../tests/constants";
 import styles from "./Modal.module.scss";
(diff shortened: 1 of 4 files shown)


packages/core/src/components/ModalNew/Modal/Modal.tsx
```diff
@@ -78,24 +79,26 @@ const Modal = forwardRef(
...
-          >
-            <ModalTopActions
-              renderAction={renderHeaderAction}
-              color={closeButtonTheme}
-              closeButtonAriaLabel={closeButtonAriaLabel}
-              onClose={onClose}
-            />
-            {children}
-          </div>
+          <FocusLock returnFocus>
```
talkor, line 82: We need to use it also in Dialogs in the future

packages/core/src/components/ModalNew/Modal/__tests__/Modal.test.tsx
```diff
@@ -129,6 +129,71 @@ describe("Modal", () => {
     expect(mockOnClose).toHaveBeenCalled();
   });
 
+  it("traps focus inside the modal when opened", () => {
```
talkor, line 132: Nice tests!!

talkor (approved):

packages/core/src/components/ModalNew/Modal/Modal.tsx
YossiSaadi: yep, we need to maybe implement it in all the system modals and old modal, and therefore all the locks would use the same lib and can live in harmony, until all would migrate to the new modal

PR #2436, opened 2024-09-15
feat: init new package @vibe/icons

rivka-ungar:


packages/icons/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "skipLibCheck": true,
+    "noImplicitAny": true,
+    "baseUrl": ".",
+    "module": "ESNext",
+    "declaration": true,
+    "declarationDir": "dist",
+    "outDir": "dist",
+    "rootDir": ".",
+    "allowSyntheticDefaultImports": true,
+    "plugins": [{ "name": "typescript-plugin-css-modules" }]
```
YossiSaadi, line 17: not sure needed, this package does not include css modules

packages/icons/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "lib": ["esnext", "dom"],
+    "target": "es6",
+    "moduleResolution": "node",
+    "esModuleInterop": true,
+    "jsx": "react",
+    "skipLibCheck": true,
+    "noImplicitAny": true,
```
YossiSaadi, line 9: add also `strict` mode please 🙏🏼

packages/icons/tsconfig.json
```diff
@@ -0,0 +1,26 @@
...
+    "allowSyntheticDefaultImports": true,
+    "plugins": [{ "name": "typescript-plugin-css-modules" }]
+  },
+  "include": [
+    "src/**/*"
+  ],
+  "exclude": [
+    "node_modules",
+    "dist/**/*.js"
+  ]
```
YossiSaadi, line 25: we might also want to exclude `__tests__` pattern here

packages/icons/tsconfig.json
YossiSaadi, line 9: you can also add here `allowJs: false` and convert the metadata file to `.ts`

YossiSaadi: notice that since you're only including `src` folder in this file, so we won't need, yet, to change the rest of the scripts (though we should sometime soon)

packages/icons/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
```
YossiSaadi, line 2: can you also please add `isolatedModules: true`

YossiSaadi (commented):
Awesome job! 🤩

packages/icons/README.md
```diff
@@ -0,0 +1,21 @@
...
+To install, use the following command:
+
+```bash
+npm install @vibe/icons
+```
+
+## Usage
+
+```javascript
+import { Add } from "@vibe/icons";
```
talkor, line 16: I'd choose something more classic than Add :) Might be confusing

packages/icons/README.md
```diff
@@ -0,0 +1,21 @@
+# @vibe/icons
+
+This package includes all of the icons of monday.com's Vibe Design System
```
talkor, line 3: Let's add a link to the catalog?

talkor: Worth mentioning that it contains React icons and SVGs

packages/icons/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "plop": "plop",
+    "build": "yarn clean && yarn build:react-icons && yarn build:esm",
+    "build:esm": "rollup -c",
+    "build:react-icons": "svg2react-icon --typescript --keep-colors src/svg/ src/react/ --no-sub-dir",
+    "clean": "rm -rf dist",
+    "clean:icons": "rm -rf src/react",
+    "lint": "yarn validate-icons",
+    "lint:fix": "yarn lint -- --fix",
+    "validate-icons": "node scripts/validate-meta.js && node scripts/icons/validate-icons-colors.js",
+    "generate-meta": "node scripts/generate-meta.js"
```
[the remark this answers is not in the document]
talkor: I believe we need it though, need to be exported

packages/icons/package.json
```diff
@@ -0,0 +1,40 @@
...
+  },
+  "devDependencies": {
+    "rollup": "^2.79.1",
+    "rollup-plugin-terser": "^7.0.2",
+    "rollup-plugin-typescript2": "^0.34.1",
+    "@rollup/plugin-node-resolve": "^15.0.1",
+    "svg2react-icon": "^3.1.178",
```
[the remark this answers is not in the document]
talkor: Let's open a ticket :)

packages/icons/plopfile.js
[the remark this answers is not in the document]
talkor: Yes I used plop 5, it handled esm a bit different

packages/icons/scripts/copy-public-folder.sh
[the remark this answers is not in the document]
talkor: I think we can delete it

packages/icons/scripts/generate-meta.js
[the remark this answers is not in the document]
talkor: I think we do need it, no? Many asked before for a list of all the icons, a json with all the metadata would be cool IMO

packages/icons/scripts/icons/utils.js
```diff
@@ -0,0 +1,41 @@
+// Files on which the rule doesn't apply
+const EXCEPTIONAL_ICONS = ["ConnectedDoc.svg"];
```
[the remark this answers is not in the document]
talkor: I talked about it with them and decided to keep it for now :(

packages/icons/tsconfig.json
talkor: Why?

talkor (commented):
Yes!!!! Very cool

packages/icons/package.json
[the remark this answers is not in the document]
YossiSaadi: why? it isn't being called anywhere
does it need to be called on-demand?

packages/icons/scripts/generate-meta.js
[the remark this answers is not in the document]
YossiSaadi: when I talked to @orrgottlieb, he suggested that if someone wants a json with all the meta, he can import the "@vibe/icons/lazy" with * and take the keys, as he would anyway probably need to use them after getting the names

packages/icons/package.json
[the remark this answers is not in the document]
YossiSaadi: Created one for next sprint

packages/icons/tsconfig.json
YossiSaadi: I know it's not in the `include`, but anyway `dist` and `node_modules` are excluded, as it is a better practice to be explicit here I believe
and tests usually need a different config
that's the reason the `core`'s config was so overloaded, as it took care for stuff related to the library itself, and storybook, and tests. so we had to, in rollup, do overrides for the publish. it's better to separate concerns

packages/icons/package.json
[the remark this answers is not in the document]
talkor: Yes, on build IMO

packages/icons/scripts/generate-meta.js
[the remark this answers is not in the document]
talkor: What about tags, description etc? Or if they need the svgs?

packages/icons/package.json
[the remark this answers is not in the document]
YossiSaadi: it isn't being called anywhere, not even on build or workflow (not even today, before the change)

packages/icons/tsconfig.json
talkor: I think it can hide errors if you have a different config for tests, and if you want to remove them from dist you can exclude in rollup config

packages/icons/package.json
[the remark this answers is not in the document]
talkor: Let's call it than

packages/icons/scripts/generate-meta.js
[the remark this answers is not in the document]
YossiSaadi: if they need the svgs they will be able to take it from "@vibe/icons/svg" in the near future
about tags and others that's a good point, but we need to check if they really need it

packages/icons/package.json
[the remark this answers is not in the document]
talkor: I think we can benefit from it with zero side effects

packages/icons/tsconfig.json
YossiSaadi: We need to inherit the tsconfig of the library's published version and add stuff that are related to tests only in the dedicated config
that's how we stay reliable to the app config but still have specific test-related config in a different one
but usually we need to add stuff which are not related to the lib in the tsconfig to not fail tests

packages/icons/package.json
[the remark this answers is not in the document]
YossiSaadi: so it needs to be part of the build command, but I'm not sure about the effect it will have

packages/icons/scripts/copy-public-folder.sh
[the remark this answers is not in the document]
rivka-ungar: Deleted

packages/icons/README.md
rivka-ungar: Changed it to "close", is that better?

packages/icons/scripts/generate-meta.js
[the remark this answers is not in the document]
rivka-ungar: After discussing it, decided to keep it

packages/icons/README.md
rivka-ungar: I added about the react icons, right now it doesn't export the svgs, when we add the svgs will add here too

packages/icons/vitest.config.ts
```diff
@@ -0,0 +1,16 @@
...
+    globals: true,
+    environment: "node",
+    include: ["**/__tests__/**/*.test.ts"],
+    clearMocks: true,
+    typecheck: {
+      enabled: true
+    },
+    onConsoleLog(log) {
+      if (log.includes("Component uses both")) return false;
+    }
```
YossiSaadi, line 14: relevant only for the codemod package

packages/icons/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "allowSyntheticDefaultImports": true,
+    "allowJs": false,
+    "isolatedModules": true
+  },
+  "include": [
+    "src/**/*"
+  ],
+  "exclude": [
+    "node_modules",
+    "dist/**/*.js"
```
YossiSaadi, line 26: add the ignore `__test__` pattern in case we'll have tests inside `src`

packages/icons/src/iconsMetaData.ts
```diff
@@ -1,7 +1,15 @@
 const BASIC = "Basic";
 const PLATFORM = "Platform";
 const VIEW = "View";
-export const iconsMetaData = [
+type icon = {
```
YossiSaadi, line 4: use PascalCase for the type

YossiSaadi:
@rivka-ungar please ping me to re-review once you think we're cool

packages/icons/src/svg/CloseRound.svg
YossiSaadi, line 1: Create a task for design + dev to fix the icon

YossiSaadi (approved):

packages/icons/README.md
```diff
@@ -16,6 +17,14 @@ npm install @vibe/icons
 import { Close } from "@vibe/icons";
 ```
 
+Alternatively, you can import the raw SVG files
+
+```javascript
+import { Close as CloseSvg } from "@vibe/icons/raw";
```
YossiSaadi, line 23: I'm wondering if it should be the initial thing we said "raw" or just "svg"
@rivka-ungar @talkor wdyt?

packages/icons/README.md
```diff
@@ -16,6 +17,14 @@ npm install @vibe/icons
 import { Close } from "@vibe/icons";
```
YossiSaadi, line 17: Maybe we should state here that this path will serve React components, same as you did for the raw one

packages/icons/package.json
```diff
@@ -17,38 +17,48 @@
...
+      ],
+      "raw": [
+        "./dist/svg/index.d.ts"
       ]
     }
   },
   "scripts": {
     "plop": "plop",
-    "build": "yarn clean && yarn build:react-icons && rollup -c",
+    "build": "yarn clean && yarn build:react-icons && rollup -c && yarn generate-svg-index",
```
YossiSaadi, line 39: I think the generate should happen before the build by rollup
is it? or it doesn't matter?
```suggestion
    "build": "yarn clean && yarn build:react-icons && yarn generate-svg-index rollup -c",
```

YossiSaadi: oh I think the order matters as it should copy in rollup the index as well, right?

YossiSaadi (approved):
Was thinking about an additional thing
we also need to remove (not sure if in this PR?) the dependency of the icons in the core package, so it won't export it

packages/icons/package.json
rivka-ungar: I generate it in the dist after it is copied. Would it be better to change it to generate it in the svg folder and then add it to the gitignore and have it copied with the rest of the files?

rivka-ungar:
> Was thinking about an additional thing we also need to remove (not sure if in this PR?) the dependency of the icons in the core package, so it won't export it

Yes, I will have a separate pr that removes the icons components from core and uses it only from @vibe/icons. I will also need to remove the svgs from style and all the related scripts.

packages/icons/package.json
YossiSaadi: sounds right to me

packages/icons/README.md
rivka-ungar: Added

packages/icons/src/svg/Mute.svg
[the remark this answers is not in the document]
rivka-ungar: Created

packages/icons/package.json
rivka-ungar: Updated

packages/icons/package.json
```diff
@@ -0,0 +1,74 @@
+{
+  "name": "@vibe/icons",
+  "version": "0.0.4",
+  "description": "Vibe's icon library",
+  "repository": "git+https://github.com/mondaycom/vibe.git",
```
YossiSaadi, line 5: can you please change to this?
```suggestion
  "repository": {
    "type": "git",
    "url": "git+https://github.com/mondaycom/vibe.git",
    "directory": "packages/icons"
  },
```

YossiSaadi (approved):
Yalla! let's merge it!

PR #2439, opened 2024-09-23
chore(icons): remove icons from core and use from package

rivka-ungar:


packages/style/README.md
```diff
@@ -23,12 +23,6 @@ or if you want to import it in your JS files
 import "monday-ui-style/dist/index.min.css";
 ```
 
-**Icons**: We expose each icon's SVG from the dist files so you can use them in your app.
-
-```javascript
-import IconName from "monday-ui-style/src/Icons/IconName.svg";
-```
-
```
talkor, line 31: We need to mention it in the changelog as well as this is breaking, and we'll have to release a major for this one (which is legit)

packages/style/scripts/__tests__/__snapshots__/validate-icons-colors.test.js.snap
talkor, line 1: It's all in the new package, right?

packages/core/package.json
```diff
@@ -89,6 +80,7 @@
   "homepage": "https://github.com/mondaycom/vibe#readme",
   "dependencies": {
     "@popperjs/core": "2.11.6",
+    "@vibe/icons": "^0.0.4",
```
talkor, line 83: Should be without a caret for the monorepo to take the correct version

talkor, line 83:
```suggestion
    "@vibe/icons": "0.0.4",
```

talkor (approved):
Yes!!

packages/style/scripts/__tests__/__snapshots__/validate-icons-colors.test.js.snap
rivka-ungar: Yes

packages/core/package.json
YossiSaadi, line 83: also should not be a dependency in my opinion, but rather a dev dependency
it should only be shown on our Storybook without exporting it again through core

packages/core/package.json
```diff
@@ -58,11 +50,10 @@
     "test:watch": "jest --watch",
     "test:update": "yarn test -u",
     "test:coverage": "yarn test -- --coverage",
-    "start": "yarn build:react-icons && yarn build:esm -- --watch",
-    "build": "yarn build:react-icons && yarn build:esm && yarn build:esm:mock-classnames",
+    "start": "yarn build:esm -- --watch",
+    "build": "yarn build:esm && yarn build:esm:mock-classnames",
```
YossiSaadi, line 54: 🥹

packages/core/package.json
YossiSaadi, line 1: worth checking if we have more deps that can be removed as a result

packages/core/package.json
YossiSaadi: oh never mind, forgot our components also consume it
I wonder how would this affect the bundle size then 👀

packages/core/types/monday-ui-style.d.ts
YossiSaadi, line 1: 💯

packages/style/README.md
YossiSaadi: let's discuss it and accumulate more stuff for a major
we should also take into consideration people who consume us **inside monday** through unpkg. they need to set themselves to a v0 tag instead of latest (which is our responsibility to add this kind of tag to a version of `style`)

YossiSaadi (approved):
Amazing work. For real. Congrats on the hard work you've put into the new package 👏🏼

packages/style/README.md
YossiSaadi: btw, we probably need to add `BREAKING` or `chore(icons)!:...`, so Lerna would know to release a major for `style`

YossiSaadi:
Notice you have missing icons which are on master

packages/style/README.md
rivka-ungar: As discussed I moved the changes of style to a different [pr](https://github.com/mondaycom/vibe/pull/2449)

PR #2442, opened 2024-09-25
fix(TextArea): Spread rest props to the native textarea tag

ziperfal:
**Main Changes:**
1. Spread `rest` props object to the native `textarea` tag in order to support native HTML API.

**Impact:**
1. A more robust component

**Related:** 
- https://monday.monday.com/boards/5896576795/pulses/7480825140
- https://monday.slack.com/archives/C02S6F817JQ/p1726471392017299

### DIFF
--- a/packages/core/src/components/TextArea/TextArea.tsx
+++ b/packages/core/src/components/TextArea/TextArea.tsx
@@ -33,7 +33,8 @@ const TextArea = forwardRef(
       required,
       maxLength,
       resize = true,
-      placeholder
+      placeholder,
+      ...rest
     }: TextAreaProps,
     ref: React.ForwardedRef<HTMLTextAreaElement>
   ) => {
@@ -76,6 +77,7 @@ const TextArea = forwardRef(
           aria-describedby={helpTextId ?? undefined}
           placeholder={placeholder}
           maxLength={maxLength}
+          {...rest}
         />
         {helpText && (
           <Text className={cx(styles.helpText)} color={Text.colors.INHERIT} id={helpTextId}>
--- a/packages/core/src/components/TextArea/TextArea.tsx
+++ b/packages/core/src/components/TextArea/TextArea.tsx
@@ -23,17 +23,10 @@ const TextArea = forwardRef(
       className,
       "data-testid": dataTestId,
       id,
-      name,
       disabled,
       readOnly,
-      value,
-      onChange,
-      onBlur,
-      "aria-label": ariaLabel,
       required,
-      maxLength,
       resize = true,
-      placeholder,
       ...rest
     }: TextAreaProps,
     ref: React.ForwardedRef<HTMLTextAreaElement>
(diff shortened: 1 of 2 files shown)


packages/core/src/components/TextArea/TextArea.tsx
YossiSaadi, line 1: Additionally, those can now be removed from the types file (only the types file, as we already get them on the `TextAreaNativeInputProps`) as they are extracted and handled inside the component and are not inside the spread:
- `rows`
- `disabled`
- `readOnly`
- `required`

Notice those should only be removed from the types declaration, and not from the destructure inside the component

YossiSaadi, line 1: Last, we have a value that we set which count on our logic
- `aria-invalid`

We need to Omit it from the `TextAreaNativeInputProps` I believe so it won't override our logic

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -33,7 +33,8 @@ const TextArea = forwardRef(
       required,
       maxLength,
       resize = true,
-      placeholder
+      placeholder,
+      ...rest
```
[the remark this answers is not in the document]
ziperfal: What ever you prefer. I am more familiar with the `rest` convention, but It is up to what you guys prefer.
Choose one and let me know :)

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -33,7 +33,8 @@ const TextArea = forwardRef(
       required,
       maxLength,
       resize = true,
-      placeholder
+      placeholder,
```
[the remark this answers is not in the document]
ziperfal: yes yes i just notice it too :) 

Thanks

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -76,6 +77,7 @@ const TextArea = forwardRef(
           aria-describedby={helpTextId ?? undefined}
           placeholder={placeholder}
           maxLength={maxLength}
+          {...rest}
```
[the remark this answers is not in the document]
ziperfal: I usually like to set default behavior and then allow the end users to override it if they explicitly send the same props.
In this case, I agree it is better to move the `...rest` up because there are "aliases" to some of the props. For example `aria-invalid={error}` and  `rows={numRows}` (instead of `rows={rows}`

Moved it up 👍

packages/core/src/components/TextArea/TextArea.tsx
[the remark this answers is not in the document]
YossiSaadi: leave it as `rest` for now, as it is internal we'll change it later if something's not clear

packages/core/src/components/TextArea/TextArea.tsx
ziperfal: Not sure i got this one. They are part of the native textarea API. 
Can you explain again what do you want me to change please?

YossiSaadi: We are re-delcaring those on the types file, but since they come from the native API, it is a bit redundant (the only change is the JSDoc, which I don't think is critical, we should stick to their declaration)

YossiSaadi: `TextAreaProps` interface re-declares stuff that arrive from the `TextAreaNativeInputProps`

ziperfal: I see. indeed redundant. 
Removed

YossiSaadi (approved):
Awesome work man, thanks for the contribution! 💯

YossiSaadi:
@ziperfal please run prettier on the types file, it fails the lint 🙃

PR #2447, opened 2024-09-29
feat(Dropdown): add scroll handlers props

asaf472monday:
<!-- Thank you for contributing!
Before we can review your submission, please fill the information below:

Please describe the changes you're making. Include the motivation for these changes, any additional context, and the impact on the project. If your changes are related to any open issues, please link to them here.:

regard my question in ask-vibe:
https://monday.slack.com/archives/C02S6F817JQ/p1727593060746269?thread_ts=1727359943.844519&cid=C02S6F817JQ

I added some events to establish an infinite scroll on dropdown
the description of the event and name are taken from here:
https://react-select.com/props

they also appear as () => void; but I'm not sure if they should be custom events like in onKeyDown, onFocus, onBlur


packages/core/src/components/Dropdown/Dropdown.types.ts
```diff
@@ -87,6 +87,18 @@ export interface DropdownComponentProps extends CustomMenuBaseProps, CustomOptio
    * Called when blurred
    */
   onBlur?: FocusEventHandler;
+  /**
+   * Called when scrolling
+   */
+  onScroll?: (event: any) => void;
```
YossiSaadi, line 93: it is suppose to be a regular HTML handler, so probably:
```suggestion
  onScroll?: UIEventHandler<HTMLDivElement>;
```

packages/core/src/components/Dropdown/Dropdown.types.ts
```diff
@@ -87,6 +87,18 @@ export interface DropdownComponentProps extends CustomMenuBaseProps, CustomOptio
...
+  /**
+   * Fired when the user scrolls to the bottom of the menu
+   */
+  onMenuScrollToBottom?: (event: WheelEvent) => void;
```
YossiSaadi, line 97: should it be:
```suggestion
  onMenuScrollToBottom?: (event: WheelEvent<HTMLDivElement>) => void;
```
?

packages/core/src/components/Dropdown/components/menu/menu.jsx
```diff
@@ -30,6 +30,7 @@ const Menu = ({ children, Renderer, selectProps, dropdownMenuWrapperClassName, i
   return (
     <components.Menu
       {...props}
+      innerProps={{ onScroll: props.onScroll }}
```
YossiSaadi, line 33: can you just make sure that `props` do not already include `innerProps`? if it does, it would override it, so we'll need to spread it in like this:
```suggestion
      innerProps={{ ...props.innerProps, onScroll: props.onScroll }}
```

Do not commit it, check before if the `props` does include it

YossiSaadi (commented):
awesome work 💪🏼 had a few comments

packages/core/src/components/Dropdown/Dropdown.types.ts
asaf472monday: TS2315: Type WheelEvent is not generic.
image

packages/core/src/components/Dropdown/components/menu/menu.jsx
asaf472monday: it has innerProps, added

packages/core/src/components/Dropdown/Dropdown.types.ts
YossiSaadi:
```suggestion
  onMenuScrollToBottom?: WheelEventHandler<HTMLDivElement>;
```

packages/core/src/components/Dropdown/components/menu/menu.jsx
YossiSaadi: cool, thanks 🙏🏼

YossiSaadi (approved):
Awesome contribution!!

PR #2449, opened 2024-09-29
chore(icons)!: remove icons from style

rivka-ungar:


YossiSaadi (approved):
it is exact same content as the other PR, right?

rivka-ungar:
> it is exact same content as the other PR, right?

Should be :)

PR #2450, opened 2024-09-29
feat(useWizard): add hook for managing wizard logic for different use cases

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7525076173

image


packages/core/src/hooks/useWizard/__stories__/useWizard.stories.module.scss
```diff
@@ -0,0 +1,6 @@
+.stepper {
```
talkor, line 1: I'd stick to the current terminology

talkor: Plus why not Flex?

packages/core/src/hooks/useWizard/__stories__/useWizard.stories.tsx
```diff
@@ -0,0 +1,129 @@
...
+          <Button onClick={next} disabled={!canGoNext}>
+            Next
+          </Button>
+        </Flex>
+      </>
+    );
+  }
+};
+
+export const WithCompletion = {
```
talkor, line 69: I think it's a bit redundant and can be in the overview story

packages/core/src/hooks/useWizard/useWizard.ts
```diff
@@ -0,0 +1,41 @@
+import { useCallback, useState } from "react";
+import { UseWizardProps, UseWizardReturnValue, WizardDirection } from "./useWizard.types";
+
+function useWizard({ initialStep = 0, stepCount, onStepChange, onComplete }: UseWizardProps): UseWizardReturnValue {
```
talkor, line 4: onComplete => onFinish? We use it in the Steps component

talkor (approved):
🥇

packages/core/src/hooks/useWizard/__stories__/useWizard.stories.module.scss
YossiSaadi: @talkor Notice it is being used inside
```
        <Steps
          className={styles.stepper}
```

I cannot use Flex

YossiSaadi: what do you mean about the terminology, which one

packages/core/src/hooks/useWizard/__stories__/useWizard.stories.tsx
YossiSaadi: agree

packages/core/src/hooks/useWizard/__stories__/useWizard.stories.module.scss
talkor: Oh ok nevermind, my bad, ignore

YossiSaadi:
@talkor I've fixed your comments

Issue #2225, opened 2024-07-11
[Bug]: Tooltip component is flickering


Ma1kovich:
### Vibe version

2.116.0

### Describe the bug

There is a special spot near or on a Tooltip's beak, which on hover starts hiding/showing the tooltip. In my app's enviroment it is flickering like crazy (many many times a second), and it is really easy to find the spot accidently, while mouse is moving over the button. In the StackBlitz enviroment the flickering is milder, but still noticable and disturbing.

Recording attached.
Notice that after I found the correct spot, I no longer move my mouse, but the cursor is changing and the tooltip is flickering.

We use `monday-ui-react-core` version `2.102.0`, in the StackBlitz enviroment it is set to `2.116.0`.

### Expected behavior

There should be no spot user can hover to achive the described behavior.

### Steps to reproduce

1. Add a Button with tooltip from Monday Vibe
2. Hover over a button near or on the tooltip's beak
3. Find a spot where it starts flickering and don't move the mouse

### Reproduction example link

https://vibe-bug-report-nsxkbf.stackblitz.io/

### Additional context, Screenshots

Recording: 
https://github.com/mondaycom/vibe/assets/39345609/ab5b425f-8d16-40e4-97cc-1634355b21fa

Hossein-Mirazimi:
I believe I may have found an issue that could be causing this bug:
![image](https://github.com/user-attachments/assets/4a473f18-48f6-44ac-9d07-0582db89c52e)
![image](https://github.com/user-attachments/assets/91584eed-d22a-4730-a0a6-74b7d82f694b)
Why are these elements positioned differently? I think they should have similar positions.
It seems that the correct position for the tooltip element is this one:
![image](https://github.com/user-attachments/assets/91584eed-d22a-4730-a0a6-74b7d82f694b)
Am I right?

If so, we should consider updating either the tooltip component or the dialog component to ensure consistency.

RitikBora:
@Ma1kovich  I have raised a pull request for this. Please let me know your thoughts as well 

Cheers :)

talkor:
Hi, thanks for opening this issue! In the latest major version (v4), we migrated the Tooltip to use Floating UI, and the issue no longer seems to reproduce. If you’re still able to reproduce it, please reopen the issue and attach a reproducible snippet. Thanks!

PR #2457, opened 2024-10-03
tooltip_flickering_fix

RitikBora:
This pull request has been raised for the issue #2225 pointed out by [Ma1kovich](https://github.com/Ma1kovich).

When tooltip component hovers over an element (Say Button), and you scroll near the area where the tooltip and the element meet. The tooltip starts to flicker.

I saw in the default moveby offset was {main: 4 , secondary: 0} which was not for enought for all top and bottom tooltips. 
For the rest is was ok.

So I added a little more main offset to increase the gap between the tooltip and the element for which tooltip is there. This solved the flickering issue for all top and bottom tooltips

I'm attaching a video with the fix for your reference. 

https://github.com/user-attachments/assets/bdddb831-b4de-418c-bf3a-58c9aa5c4b53

@talkor  let me know your thoughts on this. 

Thanks,
Ritik

### DIFF
--- a/packages/core/src/components/Tooltip/Tooltip.tsx
+++ b/packages/core/src/components/Tooltip/Tooltip.tsx
@@ -318,6 +318,7 @@ export default class Tooltip extends PureComponent<TooltipProps> {
       tip,
       arrowClassName,
       id,
+      position,
       "data-testid": dataTestId
     } = this.props;
 
@@ -329,9 +330,16 @@ export default class Tooltip extends PureComponent<TooltipProps> {
       return this.renderTooltipContent();
     }
 
+    let moveAmount : MoveBy = this.props.moveBy;
+
+    if (position && (position.includes("top") || position.includes("bottom"))) {
+        moveAmount = { main: 11, secondary: 0 };
+    }
+
     const content = this.renderTooltipContent;
     const dialogProps = {
       ...this.props,
+      moveBy: moveAmount,
       "data-testid": dataTestId || getTestId(ComponentDefaultTestId.TOOLTIP, id),
       startingEdge: justify,
       tooltip: tip,


packages/core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -329,9 +330,16 @@ export default class Tooltip extends PureComponent<TooltipProps> {
       return this.renderTooltipContent();
     }
 
+    let moveAmount : MoveBy = this.props.moveBy;
+
+    if (position && (position.includes("top") || position.includes("bottom"))) {
+        moveAmount = { main: 11, secondary: 0 };
```
talkor, line 336: How will it apply to smaller components/elements? I feel like it will distant the tooltip from the referencing element

talkor (commented):
Thanks for this PR @RitikBora ! Can you please check my comment? 
Also I feel like we'll have to test it a bit to make sure it behaves well in all scenarios

RitikBora:
Hey @talkor ,

Thank you for the feedback.

The default prop of {main: 4 , secondary: 0} seemed a little less to me and thats why I modified move by. But I do get your point here . It looks okay on the button but might not look good on some other elements

I'll try out a couple of things here.
1) Test it out with some smaller elements. Any leads to what all should a try would be appriated :)
2) Try a different approach where the space between is dynamic and just enough for the tooltip and element to be considered separate .

Let me know if there is anything else you'd like to suggest.

Thanks,
Ritik

talkor:
> Hey @talkor ,
> 
> 
> 
> Thank you for the feedback.
> 
> 
> 
> The default prop of {main: 4 , secondary: 0} seemed a little less to me and thats why I modified move by. But I do get your point here . It looks okay on the button but might not look good on some other elements
> 
> 
> 
> I'll try out a couple of things here.
> 
> 1) Test it out with some smaller elements. Any leads to what all should a try would be appriated :)
> 
> 2) Try a different approach where the space between is dynamic and just enough for the tooltip and element to be considered separate .
> 
> 
> 
> Let me know if there is anything else you'd like to suggest.
> 
> 
> 
> Thanks,
> 
> Ritik

That's awesome @RitikBora, thank you! I'll try to check it out and find some leads as well, and update here 🙏

talkor:
Closing for now, feel free to re-open if relevant again, thanks!

PR #2462, opened 2024-10-07
feat(ModalFooterWizard): wizard modal footer

YossiSaadi:
### ModalFooterWizard

image

https://monday.monday.com/boards/3532714909/pulses/7360993682

### DIFF
--- a/packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
+++ b/packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
@@ -9,6 +9,6 @@
 .stepDots {
   position: absolute;
   left: 50%;
-  translate: -50%;
+  transform: translateX(-50%);
   margin: 0;
 }
--- a/packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
+++ b/packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
@@ -1,0 +1,14 @@
+.primary {
+  order: 2;
+}
+
+.secondary {
+  order: 1;
+}
+
+.stepDots {
+  position: absolute;
+  left: 50%;
+  translate: -50%;
+  margin: 0;
+}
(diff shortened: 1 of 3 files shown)


packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
```diff
@@ -0,0 +1,14 @@
+.primary {
+  order: 2;
```
talkor, line 2: Why are you controlling the order?

packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
```diff
@@ -0,0 +1,14 @@
...
+}
+
+.secondary {
+  order: 1;
+}
+
+.stepDots {
+  position: absolute;
+  left: 50%;
+  translate: -50%;
```
talkor, line 12: transform: translate?

packages/core/src/components/Steps/StepsGalleryHeader.tsx
```diff
@@ -9,7 +10,7 @@ export interface StepsGalleryHeaderProps extends VibeComponentProps {
   activeStepIndex: number;
   stepsCount: number;
   onChangeActiveStep: (e: React.MouseEvent, stepIndex: number) => void;
-  stepDescriptionFunc: (stepIndex: number) => string;
+  stepDescriptionFunc?: (stepIndex: number) => string;
```
talkor, line 13: I'm not really fond of this name

talkor (approved):

packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
YossiSaadi: > it is a bad practice, but design wants the primary to be first tabbed order
and using positive tabindex is even a worse practice
so I've manipulated the render order

packages/core/src/components/ModalNew/footers/ModalFooterWizard/ModalFooterWizard.module.scss
YossiSaadi: it works basically, but you are right, it is not supported enough (104 according to caniuse)

packages/core/src/components/Steps/StepsGalleryHeader.tsx
YossiSaadi: lol
this should be re-written, I don't like the way this component works

PR #2466, opened 2024-10-09
feat: add inputAriaLabel to Dropdown component

joshpe-monday:

Adding the ability to provide an `ariaLabel` specifically for the input of the `Dropdown` component. This is fully backwards compatible, providing the aria label `"Dropdown input"` that it used to by default, but if you provide the `inputAriaLabel` prop then it will use that instead.


YossiSaadi (approved):
Looks cool :)

PR #2467, opened 2024-10-09
Support SSR injected styles

MosheZemah:
https://monday.monday.com/boards/4400799549/pulses/6799260682

### DIFF
--- a/packages/core/rollup/styleInject.ejs
+++ b/packages/core/rollup/styleInject.ejs
@@ -1,21 +1,25 @@
 function styleInject(css) {
-    if(typeof document === 'undefined') return;
-    const head = document.head || document.getElementsByTagName('head')[0];
     const id = "<%= hashValue %>_<%= version %>";
-    const styleExists = head.querySelector("#" + id);
-
-    if(styleExists) return;
-
-    const style = document.createElement('style');
-    style.id = id;
-
-    if (head.firstChild) {
-        head.insertBefore(style, head.firstChild);
-    } else {
-        head.appendChild(style);
+    if(typeof document !== 'undefined') {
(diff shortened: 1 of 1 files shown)


talkor (approved):

packages/core/rollup/styleInject.ejs
```diff
@@ -1,21 +1,25 @@
...
+        if (head.firstChild) {
+            head.insertBefore(style, head.firstChild);
+        } else {
+            head.appendChild(style);
+        }
+    
+        style.appendChild(document.createTextNode(css));
+    } else if (typeof globalThis !== 'undefined' && globalThis.injectedStyles) {
+        // support SSR environments
+        globalThis.injectedStyles[id] = css;
```
YossiSaadi, line 21: this is not a direct dependency of vibe, should it be? (it's not used anywhere in the code, only in your addition so might be good to add it as a direct dependency?)

packages/core/rollup/styleInject.ejs
```diff
@@ -1,21 +1,25 @@
...
+        style.id = id;
+    
+        if (head.firstChild) {
+            head.insertBefore(style, head.firstChild);
+        } else {
+            head.appendChild(style);
+        }
+    
+        style.appendChild(document.createTextNode(css));
+    } else if (typeof globalThis !== 'undefined' && globalThis.injectedStyles) {
```
orrgottlieb, line 19: why are you checking the injectedStyles object? 
and can you add a section about CSS and SSR to the README

MosheZemah: No need to check if globalThis exists? I did it just in case. Wasn't sure if needed

PR #2477, opened 2024-10-09
Fix ssr issues

MosheZemah:
https://monday.monday.com/boards/4400799549/pulses/6799260682

### DIFF
--- a/packages/core/src/components/Modal/Modal.tsx
+++ b/packages/core/src/components/Modal/Modal.tsx
@@ -13,6 +13,7 @@ import { getTestId } from "../../tests/test-ids-utils";
 import { ComponentDefaultTestId } from "../../tests/constants";
 import styles from "./Modal.module.scss";
 import { useWarnDeprecated } from "../../utils/warn-deprecated";
+import { isClient } from "../../utils/ssr-utils";
 
 export interface ModalProps {
   /**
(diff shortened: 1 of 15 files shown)


packages/core/src/components/Modal/Modal.tsx
```diff
@@ -170,40 +171,42 @@ const Modal: FC<ModalProps> & { width?: typeof ModalWidth } = ({
...
-    <div
-      {...attr.container}
-      className={cx(styles.container, classNames.container)}
-      data-testid={dataTestId || getTestId(ComponentDefaultTestId.MODAL, id)}
-      style={{ "--monday-modal-z-index": zIndex }}
-    >
-      {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
+  const dialog =
+    isClient() &&
+    ReactDOM.createPortal(
```
YossiSaadi, line 176: seem like you can remove this `ReactDOM.createPortal` from here
it calls `createPortal` at [line 209](https://github.com/mondaycom/vibe/pull/2477/files#diff-912bdffbe0842863767b158e48535904a784f7bebe85bcd4dfcf42ef32b05f6fR209) as well

at the current implementation it would not render the Modal at all when SSR

YossiSaadi (approved):
had one thing to fix, but approving

packages/core/src/components/Modal/Modal.tsx
YossiSaadi, line 176:
```suggestion
  const dialog =
```

PR #2485, opened 2024-10-10
Add placeholder for custom svg icon for SSR [prerelease]

MosheZemah:


YossiSaadi (approved):

packages/core/src/components/Modal/__tests__/__snapshots__/Modal.snapshot.test.js.snap
```diff
@@ -154,7 +154,6 @@ Snapshot Diff:
 +       <div
 +         class="typography primary start text text1Normal container"
 +         data-testid="modal-content"
-+         style="--text-clamp-lines: 1;"
```
YossiSaadi, line 157: seem like an unwanted change?

MosheZemah: I think it is ok - I've removed text-clamp if single ellipsis line

PR #2486, opened 2024-10-10
[fix #2476] docs: convert useClickableProps stories file to TS

### LINKED ISSUE #2476: docs: convert useClickableProps stories file to TS
Convert the `useClickableProps` stories file to TypeScript:
- Change the file extension to `stories.tsx`
- Resolve type issues (if there are), avoid using `any` and prefer existing types

## Related files
`packages/core/src/hooks/useClickableProps/__stories__/useClickableProps.stories.js`

---

[Monday ticket [internal]](https://monday.monday.com/boards/3532714909/pulses/7601091903)

ahammadabdullah:
First of all I changed the .prettierrc file and set `"endOfLine": "auto"` to avoid crlf issues. 
Then I Changed the useClickableProps.stories.js to useClickableProps.stories.tsx file as the issue mentioned. And defined types properly. 


packages/core/src/hooks/useClickableProps/__stories__/useClickableProps.stories.tsx
```diff
@@ -23,7 +31,7 @@ export const Overview = {
     );
 
     return (
-      <div {...clickableProps} className="monday-storybook-use-clickable-props">
+      <div {...clickableProps} className="monday-storybook-use-clickable-props" ref={ref}>
```
talkor, line 34: The ref is already provided in `clickableProps`

packages/core/src/hooks/useClickableProps/__stories__/useClickableProps.stories.tsx
```diff
@@ -1,17 +1,25 @@
-import React, { useCallback, useRef } from "react";
+import React, { useCallback, useRef, RefObject, MouseEvent, KeyboardEvent } from "react";
 import useClickableProps from "../useClickableProps";
 import "./useClickableProps.stories.scss";
 
+interface ClickableProps extends React.HTMLAttributes<HTMLElement> {
```
talkor, line 5: No need to re-create the interface, we already use such interface in the hook :) We can just import it

.prettierrc
```diff
@@ -1,5 +1,6 @@
 {
   "printWidth": 120,
   "trailingComma": "none",
-  "arrowParens": "avoid"
+  "arrowParens": "avoid",
+  "endOfLine": "auto"
```
talkor, line 5: Let's avoid such changes as they have an effect on the entire repo, plus it's out of the scope and we currently don't want changes to prettier config at the moment 🙏🏼

talkor (changes requested):
Thanks for the PR @ahammadabdullah ! Please see my comments

.prettierrc
ahammadabdullah: @talkor i understand. But without prettier config I'm getting CRLF errors. Every page I open I see error in every line. :) do you have any idea how can I solve this issue ?

packages/core/src/hooks/useClickableProps/__stories__/useClickableProps.stories.tsx
ahammadabdullah: image
getting this error with the imported type. Now can I change that interface?

ahammadabdullah: @talkor in the interface `  tabIndex?: string | number;` defined as this. But im getting error because it has to be number only. Either I cast the tabindex explicitly or change the type interface. What to do ?

talkor: @ahammadabdullah there seem to be an issue with the return type of the ref in useClickableProps which might be a breaking change, probably using a generic type will be best here but it's out of the scope. Let's close it for now, appreciate the effort! We have more issues to tackle if you'd like🙏🏼Thanks!

PR #2488, opened 2024-10-10
Bug/moshe/fix custom svg

MosheZemah:


YossiSaadi (commented):
it affects also non-server usages, is it safe? did you had the chance to thoroughly check this change?

PR #2489, opened 2024-10-10
feat(Modal): add pop from center/anchor animations

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7586333722
https://monday.monday.com/boards/3532714909/pulses/7586334945

### DIFF
--- a/packages/core/src/components/ModalNew/Modal/Modal.tsx
+++ b/packages/core/src/components/ModalNew/Modal/Modal.tsx
@@ -14,7 +14,11 @@ import { ModalProvider } from "../context/ModalContext";
 import { ModalContextProps } from "../context/ModalContext.types";
 import useKeyEvent from "../../../hooks/useKeyEvent";
 import { keyCodes } from "../../../constants";
-import { modalAnimationCenterPopVariants, modalAnimationOverlayVariants } from "../utils/animationVariants";
+import {
+  modalAnimationAnchorPopVariants,
+  modalAnimationCenterPopVariants,
+  modalAnimationOverlayVariants
+} from "../utils/animationVariants";
 
 const Modal = forwardRef(
   (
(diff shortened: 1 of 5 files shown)


packages/core/src/components/ModalNew/utils/animationVariants.ts
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+};
+
+export const modalAnimationAnchorPopVariants: Variants = {
+  enter: {
+    opacity: [0, 1, 1],
+    scale: [0.65, 0.65, 1],
+    top: "50%",
+    left: "50%",
```
rivka-ungar, line 51: Are the `top` & `left` needed here as they are defined already in the css?

packages/core/src/components/ModalNew/Modal/Modal.tsx
```diff
@@ -67,41 +74,54 @@ const Modal = forwardRef(
...
-          </FocusLock>
-        </RemoveScroll>
-      </ModalProvider>
+              <FocusLock returnFocus>
+                <motion.div
+                  variants={modalAnimationVariants}
+                  initial="exit"
+                  animate="enter"
+                  exit="exit"
+                  custom={anchorElementRef}
```
rivka-ungar, line 102: Does this need a fallback in case it is undefined, so it won't mess up the animation?

rivka-ungar (approved):
Very cool! Had 2 small questions

packages/core/src/components/ModalNew/Modal/Modal.tsx
YossiSaadi: checked it, if it is undefined then the pop from center variation would be applied and it doesn't use it
if it is being used from `custom`, then the pop from anchor variation would be applied, and it means that the anchor element ref isn't undefined

PR #2528, opened 2024-10-13
docs(Clickable): related component should use Text instead of inline string

### LINKED ISSUE #2507: Change `Clickable` catalog component to use `Text` component
In the file of:
`packages/core/src/storybook/components/related-components/descriptions/clickable-description/clickable-description.jsx`
which can be seen on our [Catalog](https://vibe.monday.com/?path=/docs/catalog--docs) page, we do not currently use Vibe's `Text` component, but instead use inline string.
It also causes theme issues where a story doesn't look right on dark themes.

```typescript
<Clickable className={classes["clickable-description"]} onClick={() => alert("clicked")}>
  <div>Button with custom appearance</div>
</Clickable>
```
--> Something like
```typescript
<Clickable className={classes["clickable-description"]} onClick={() => alert("clicked")}>
  <Text ellipsis={false}>Button with custom appearance</Text>
</Clickable>
```

- Notice the `ellipsis` as `false` as well.
- Notice to remove `color` from the appropriate css as well attached to the file

monday ticket:

Aejkatappaja:
We do not currently use Vibe's `Text` component, but instead use inline string.

This PR fix this, `Text` component is now properly used.

Resolves #2507

### DIFF
--- a/packages/core/src/storybook/components/related-components/descriptions/clickable-description/clickable-description.jsx
+++ b/packages/core/src/storybook/components/related-components/descriptions/clickable-description/clickable-description.jsx
@@ -3,12 +3,13 @@ import React, { useMemo } from "react";
 import { RelatedComponent } from "vibe-storybook-components";
 import Clickable from "../../../../../components/Clickable/Clickable";
 import classes from "./clickable-description.stories.module.scss";
+import { Text } from "../../../../../components";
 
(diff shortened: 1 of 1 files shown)


YossiSaadi (approved):
Awesome @Aejkatappaja, thanks!

PR #2534, opened 2024-10-13
docs(VirtualizedGrid): should use Text instead of inline string

### LINKED ISSUE #2505: Change `VirtualizedGrid` catalog component to use `Text` component
In the file of:
`packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js`
which can be seen on our [Catalog](https://vibe.monday.com/?path=/docs/catalog--docs) page, we do not currently use Vibe's `Text` component, but instead use inline string.
It also causes theme issues where a story doesn't look right on dark themes.

We need to change the wrapper `div` to be `Text` instead.
```typescript
<div
  style={{
    backgroundColor,
    height: item.height,
    width: item.width,
    display: "flex",
    alignItems: "center",
    justifyContent: "center"
  }}
>
  {item.value}
</div>
```
--> Something like
```typescript
<Text
  color={Text.colors.FIXED_DARK}
  style={{
    backgroundColor,
    height: item.height,
    width: item.width,
    display: "flex",
    alignItems: "center",
    justifyContent: "center"
  }}
>
  {item.value}
</Text>
```

Notice the `FIXED_DARK` color addition as well - we need it so it would be shown correct on all themes.

monday ticket:

GurneeshBudhiraja:
Changes made:
Replaced the `div` with the `Text` component in the `packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js` file.

Resolves #2505

### DIFF
--- a/packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js
+++ b/packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js
@@ -1,5 +1,5 @@
 import React from "react";
-import { Text } from "src/components/Text";
+import { Text } from "../../Text";
 
 export const generateItems = (height = 30, width = "100%", itemsCount) => {
   const items = [];
--- a/packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js
+++ b/packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js
@@ -1,4 +1,5 @@
 import React from "react";
+import { Text } from "src/components/Text";
 
 export const generateItems = (height = 30, width = "100%", itemsCount) => {
   const items = [];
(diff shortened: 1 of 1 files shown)


YossiSaadi (dismissed):
Awesome! thanks @GurneeshBudhiraja

packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js
```diff
@@ -1,4 +1,5 @@
 import React from "react";
+import { Text } from "src/components/Text";
```
YossiSaadi, line 2: just noticed this is being imported wrongly
did you ran Storybook to make sure it works?
the import should look something like
```
import { Text } from "../../Text";
```
as we do not have aliases yet in the project

YossiSaadi (changes requested):
Can you please fix my comment @GurneeshBudhiraja ?

packages/core/src/components/VirtualizedGrid/__stories__/VirtualizedGrid.stories.helpers.js
GurneeshBudhiraja: My bad.
Just fixed the import statement. Also ran storybook and tests to make sure everything is alright.

YossiSaadi (approved):
💯

PR #2540, opened 2024-10-15
Fix react select import since it is a commonJS module

MosheZemah:


YossiSaadi:
It sound like this can increase the bundle even more (and react-select is heavy anyway)
@talkor can you insert the moduled react select to vibe2 as well? Or Alternatively, @MosheZemah can you use vibe3.0?

talkor:
> It sound like this can increase the bundle even more (and react-select is heavy anyway) @talkor can you insert the moduled react select to vibe2 as well? Or Alternatively, @MosheZemah can you use vibe3.0?

We fixed it in the fork so this can be closed

PR #2554, opened 2024-10-20
docs: pause animation at first frame

talkor:
https://monday.monday.com/boards/3532714909/pulses/7598655120

### DIFF
--- a/packages/core/src/components/Toast/__stories__/Toast.stories.tsx
+++ b/packages/core/src/components/Toast/__stories__/Toast.stories.tsx
@@ -36,7 +36,8 @@ export const Overview = {
         content: "Button"
       }
     ]
-  }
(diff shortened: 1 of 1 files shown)


YossiSaadi (approved):
if this is relevant for all stories in this file (or at least does not affect them badly, if not relevant), you can move it to the default export of the meta as `parameters: { chromatic: { pauseAnimationAtEnd: false } }` as well

talkor:
> if this is relevant for all stories in this file (or at least does not affect them badly, if not relevant), you can move it to the default export of the meta as `parameters: { chromatic: { pauseAnimationAtEnd: false } }` as well

I thought about that but wanted to solve for now the exact stories that fail and continue monitoring if others fail too

PR #2555, opened 2024-10-20
fix: EditableText and EditableHeading overflowing in the catalog

### LINKED ISSUE #2471: docs: fix EditableText and EditableHeading overflowing in the catalog
In the [Catalog](https://vibe.monday.com/?path=/docs/catalog--docs) there are interactive examples of components.

The `EditableText` and `EditableHeading` examples are overflowing. See screenshots:
Screenshot 2024-10-09 at 18 28 12

Screenshot 2024-10-09 at 18 28 22

Need to make them stretch, for example like the Heading:
Screenshot 2024-10-09 at 18 28 44

## Related files
`packages/core/src/storybook/components/related-components/descriptions/editable-text-description.tsx`
`packages/core/src/storybook/components/related-components/descriptions/editable-heading-description.jsx`

---
[Monday ticket [internal]](https://monday.monday.com/boards/3532714909/pulses/5740467028)

mashazyu:

Resolves #2471

### Details

Based on [this tip](https://github.com/mondaycom/vibe/issues/2471#issuecomment-2409017206) I updated editableTypography style. 

The following checks pass locally:
- [X] tests 
- [X] linter 

To verify that changes work, I checked catalog locally and it looks like overflow works correctly (see screenshots below)
image
image

Questions:
- Are there other scenarios that I should test?
- In the component description of <Heading> there is information about overflow behavior, while there is no such info in the description for <EditableHeading>. Should it be added?

### DIFF
--- a/packages/core/src/components/EditableTypography/EditableTypography.module.scss
+++ b/packages/core/src/components/EditableTypography/EditableTypography.module.scss
@@ -1,7 +1,7 @@
 @import "~monday-ui-style/dist/mixins";
 
 .editableTypography {
-  display: inline-flex;
+  display: contents;
   min-width: 0;
   max-width: 100%;
   // Shifts the component to align the text with the container
--- a/packages/core/src/components/EditableTypography/EditableTypography.module.scss
+++ b/packages/core/src/components/EditableTypography/EditableTypography.module.scss
@@ -1,11 +1,12 @@
 @import "~monday-ui-style/dist/mixins";
 
 .editableTypography {
-  display: contents;
+  display: block;
   min-width: 0;
-  max-width: 100%;
   // Shifts the component to align the text with the container
   margin-left: -6px;
+  // Adds 6 px removed above to the element width to avoid unnecessary overflow
+  max-width: calc(100% + 6px);
   overflow: hidden;
   position: relative;
 
--- a/packages/core/src/components/EditableTypography/EditableTypography.module.scss
+++ b/packages/core/src/components/EditableTypography/EditableTypography.module.scss
@@ -1,7 +1,7 @@
 @import "~monday-ui-style/dist/mixins";
 
 .editableTypography {
-  display: block;
+  display: inline-flex;
   min-width: 0;
   // Shifts the component to align the text with the container
   margin-left: -6px;


packages/core/src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -1,7 +1,7 @@
 @import "~monday-ui-style/dist/mixins";
 
 .editableTypography {
-  display: inline-flex;
+  display: contents;
```
talkor, line 4: Let's use `display: block` instead as `contents` [might affect a11y](https://developer.mozilla.org/en-US/docs/Web/CSS/display#display_contents)

talkor (changes requested):
Thanks @mashazyu ! Please see my comment

packages/core/src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -1,11 +1,12 @@
 @import "~monday-ui-style/dist/mixins";
 
 .editableTypography {
-  display: inline-flex;
+  display: block;
```
mashazyu, line 4: **note:** 

I still changed display to block because it was requested.

packages/core/src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -1,11 +1,12 @@
...
   min-width: 0;
-  max-width: 100%;
   // Shifts the component to align the text with the container
   margin-left: -6px;
+  // Adds 6 px removed above to the element width to avoid unnecessary overflow
+  max-width: calc(100% + 6px);
```
mashazyu, line 9: **question:** 
@talkor 

First of all, thank you very much for the quick feedback and pointing out accessibility issues.

I didn't go with "display: block" initially because it didn't fix the overflowing issue. I looked further and realized that what's causing this issue is this negative margin-left in combination with max-width set to 100%. Adding missing 6px to max-width fixes the issue without chaning the display value.

Wdyt about this approach to fixing this issue and do you think the display value still needs to be updated?

talkor: Thanks @mashazyu ! Looks like a great solution. Since there's a change in the component we'll need to test it a bit to make sure it doesn't break anything. Assuming everything is ok I'll merge it. In the meantime I've added the hacktoberfest accepted tag. Thank you so much for this!

packages/core/src/components/EditableTypography/EditableTypography.module.scss
talkor: Seems like this causes some issues, see video attached:

https://github.com/user-attachments/assets/733a6231-3798-469c-afe8-b1c4a0dd0f05

talkor (changes requested):
Great direction! But it breaks the component

packages/core/src/components/EditableTypography/EditableTypography.module.scss
mashazyu: @talkor 

Thank you for testing! I think changing display to 'block' caused that behavior. I reverted this change and I this issue is fixed for me locally  - I verified it locally and ran lint & tests. Does proposed change fix the issue for you?

mashazyu:
> Great direction! But it breaks the component

Hello, @talkor !

I reverted back the change of display value and it seems the bug is fixed. Could you please check it from your end?

talkor (approved):
Thank you @mashazyu ! Will be merged in the next major version :)

mashazyu:
@talkor 

Thank you very much for the update! Great news!

talkor:
Hey @mashazyu ! I found a workaround that will not be breaking - changing only the container in the component example. In the `packages/core/src/storybook/components/related-components/descriptions/editable-heading-description.tsx` file you can replace the wrapping `div` with a flex container and width `width: 100%`, e.g:

```
    <Flex justify="center" style={{ width: "100%" }}>
        <EditableHeading value="Hello world" />
      </Flex>
```

Same for the editable text description file. Would you like to take it? :) We'll be able to push it right away instead of in a major version since that change will be only in the docs

mashazyu:
> Hey @mashazyu ! I found a workaround that will not be breaking - changing only the container in the component example. In the `packages/core/src/storybook/components/related-components/descriptions/editable-heading-description.tsx` file you can replace the wrapping `div` with a flex container and width `width: 100%`, e.g:
> 
> ```
>     <Flex justify="center" style={{ width: "100%" }}>
>         <EditableHeading value="Hello world" />
>       </Flex>
> ```
> 
> Same for the editable text description file. Would you like to take it? :) We'll be able to push it right away instead of in a major version since that change will be only in the docs

Hello @talkor ,

Sorry for delay and sure, I'd love to. 

A small clarifying question - I thought this issue is about library component, not the way it is displayed in Storybook. Am I incorrect and only the Storybook representation should be updated?

talkor:
> > Hey @mashazyu ! I found a workaround that will not be breaking - changing only the container in the component example. In the `packages/core/src/storybook/components/related-components/descriptions/editable-heading-description.tsx` file you can replace the wrapping `div` with a flex container and width `width: 100%`, e.g:
> > ```
> >     <Flex justify="center" style={{ width: "100%" }}>
> >         <EditableHeading value="Hello world" />
> >       </Flex>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Same for the editable text description file. Would you like to take it? :) We'll be able to push it right away instead of in a major version since that change will be only in the docs
> 
> Hello @talkor ,
> 
> Sorry for delay and sure, I'd love to.
> 
> A small clarifying question - I thought this issue is about library component, not the way it is displayed in Storybook. Am I incorrect and only the Storybook representation should be updated?

Hey @mashazyu, sorry for the misunderstanding, the issue is indeed about the docs and not the component

mashazyu:
> Hey @mashazyu, sorry for the misunderstanding, the issue is indeed about the docs and not the component

Thank you for the clarification, @talkor ! So no further changes are needed from my side at this point?

talkor:
> > Hey @mashazyu, sorry for the misunderstanding, the issue is indeed about the docs and not the component
> 
> Thank you for the clarification, @talkor ! So no further changes are needed from my side at this point?

Yes, so this PR can be closed, and if you want to make the changes in the docs as I described above that would be great!
Thanks!

PR #2556, opened 2024-10-20
feat(Modal): control autofocus and keyboard navigation focus

YossiSaadi:
- Always start autofocus cycle on content
- Skip top actions (close button + render action) from autofocusing on first focus cycle of modal

https://monday.monday.com/boards/3532714909/pulses/7360929204


rivka-ungar (approved):
💯

PR #2557, opened 2024-10-21
feat(TransitionView): component to be used on wizard-related logics to render a step with animation between steps

YossiSaadi:
WizardSlideshow result:
Oct-21-2024 15-43-02

Original design request:
Oct-21-2024 15-45-06

#### Note: name was changed from `WizardSlideshow` to `TransitionView`

https://monday.monday.com/boards/3532714909/pulses/7525075588


YossiSaadi:
@shaharzil @talkor I'm not a big favor of the name I gave to that component, but I think using just Wizard here is too confusing, as it is only responsible on rendering a single child with animation among given children. it does not do any other logic related to Wizard in general
Would love to get your opinion on that one

packages/core/src/components/WizardSlideshow/__tests__/WizardSlideshow.test.tsx
talkor, line 1: I personally don't like the naming. I think it should be a Carousel

packages/core/src/components/SlideTransition/SlideTransition.tsx
```diff
@@ -0,0 +1,27 @@
...
+        custom={direction}
+        variants={slideAnimationVariants}
+        initial="initial"
+        animate="enter"
+        exit="exit"
+        transition={slideAnimationTransition}
+        className={cx(styles.slide, className)}
+      >
+        {children}
+      </motion.div>
```
talkor, line 22: Very cool!

packages/core/src/components/WizardSlideshow/WizardSlideshow.tsx
```diff
@@ -0,0 +1,32 @@
...
+    ref: React.ForwardedRef<HTMLDivElement>
+  ) => {
+    return (
+      <div
+        id={id}
+        className={cx(styles.slideshow, className)}
+        data-testid={dataTestId || getTestId(ComponentDefaultTestId.WIZARD_SLIDESHOW, id)}
+        ref={ref}
+      >
+        <AnimatePresence initial={false} custom={direction}>
```
talkor, line 22: No initial animation?

packages/core/src/components/SlideTransition/SlideTransition.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return (
+      <motion.div
+        ref={ref}
+        custom={direction}
+        variants={slideAnimationVariants}
+        initial="initial"
+        animate="enter"
+        exit="exit"
+        transition={slideAnimationTransition}
+        className={cx(styles.slide, className)}
```
talkor, line 19: Does it really need a className?

talkor (approved):
Very cool
I personally don't like the naming, worth considering it
Also notice you are not exporting any of these components

packages/core/src/components/SlideTransition/SlideTransition.tsx
YossiSaadi: for now it would be an internal component, but in case we decide to expose a few animation components next that is an hachana lemazgan, I gave it all the vibe props 
`export interface SlideTransitionProps extends VibeComponentProps {`

packages/core/src/components/WizardSlideshow/__tests__/WizardSlideshow.test.tsx
YossiSaadi: I think carousel is more of a "never ending" thing where you see everything, no?

YossiSaadi: btw, it doesn't have to be only slide, this can - in the future - to accept a variation for animation and just be some kind of animation controller

talkor: So it's just Wizard, why the Slideshow? Slideshow sounds like PowerPoint

packages/core/src/components/SlideTransition/SlideTransition.tsx
talkor: Ok. Anyway notice you are not exporting the other component. Not sure if it's intended or not

YossiSaadi: nope, will do

packages/core/src/components/WizardSlideshow/__tests__/WizardSlideshow.test.tsx
YossiSaadi: suggesting TransitionView

packages/core/src/components/SlideTransition/SlideTransition.tsx
YossiSaadi: Done

packages/core/src/components/WizardSlideshow/__tests__/WizardSlideshow.test.tsx
talkor: Like it!

PR #2560, opened 2024-10-24
feat(ListItemAvatar): support Avatar props

### LINKED ISSUE #2008: [Feat] ListItemAvatar - Be able to provide all Avatar props
Currently `ListItemAvatar` component can get only a few props (e.g. `src`, `type`, `size`). It should be able to get all `Avatar` props (`AvatarProps`). It can be achieved by extending the `ListItemAvatarProps` to `AvatarProps` and spreading the avatar props into the component.

Monday ticket: https://monday.monday.com/boards/3532714909/pulses/5741029337

naorpeled:

Resolves #2008

### DIFF
--- a/packages/core/src/components/ListItemAvatar/ListItemAvatar.tsx
+++ b/packages/core/src/components/ListItemAvatar/ListItemAvatar.tsx
@@ -1,12 +1,12 @@
 import React, { forwardRef, useRef } from "react";
 import cx from "classnames";
 import useMergeRef from "../../hooks/useMergeRef";
-import { VibeComponent, VibeComponentProps, withStaticProps } from "../../types";
-import Avatar from "../Avatar/Avatar";
+import { VibeComponent, withStaticProps } from "../../types";
+import Avatar, { AvatarProps } from "../Avatar/Avatar";
 import { ListItemComponentType } from "../ListItem/ListItemConstants";
 import styles from "./ListItemAvatar.module.scss";
 
-export interface ListItemAvatarProps extends VibeComponentProps {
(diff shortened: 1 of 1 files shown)


packages/core/src/components/ListItemAvatar/ListItemAvatar.tsx
```diff
@@ -27,6 +36,7 @@ const ListItemAvatar: VibeComponent<ListItemAvatarProps> & { components?: typeof
           type={Avatar.types.IMG}
           size={Avatar.sizes.SMALL}
           className={cx(styles.avatar, avatarClassName)}
+          {...avatarProps}
```
naorpeled, line 39: should I omit `className` and `id` as they're applied to the `Component`?

naorpeled:
Hey @talkor,
the PR is ready for a review,
let me know what you think 🙏

packages/core/src/components/ListItemAvatar/ListItemAvatar.tsx
talkor: They are already omitted while spreading the props so no need

talkor (approved):
Excellent! Thank you @naorpeled!

packages/core/src/components/ListItemAvatar/ListItemAvatar.tsx
naorpeled: Oh, you're absolutely right, my bad 🤦

talkor:
Thank you so much for the effort! But we have decided to drop this ability as we are advancing the usage of the new List component which has different Avatar flexibility. Thanks!

PR #2566, opened 2024-10-28
feat(AttentionBox): add prop for enter animation

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7719404878

### DIFF
--- a/packages/core/src/components/AttentionBox/AttentionBox.tsx
+++ b/packages/core/src/components/AttentionBox/AttentionBox.tsx
@@ -36,6 +36,7 @@ export interface AttentionBoxProps extends VibeComponentProps {
   onClose?: (event: React.MouseEvent) => void;
   compact?: boolean;
   closeButtonAriaLabel?: string;
+  animateOnEnter?: boolean;
 }
 
 const AttentionBox: React.FC<AttentionBoxProps> & {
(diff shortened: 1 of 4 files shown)


packages/core/src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -71,7 +73,9 @@ const AttentionBox: React.FC<AttentionBoxProps> & {
 
   return (
     <aside
-      className={cx(styles.attentionBox, getStyle(styles, camelCase(`type-${type}`)), overrideClassName)}
+      className={cx(styles.attentionBox, getStyle(styles, camelCase(`type-${type}`)), overrideClassName, {
+        [styles.animateOnEnter]: animateOnEnter
```
YossiSaadi, line 77: should this class be removed after the animation ends? or does the animation just works once and you don't care about it afterwards?

YossiSaadi (approved):

packages/core/src/components/AttentionBox/AttentionBox.tsx
YossiSaadi: ok nm, I don't think it matters

packages/core/src/components/AttentionBox/AttentionBox.tsx
```diff
@@ -36,6 +36,7 @@ export interface AttentionBoxProps extends VibeComponentProps {
   onClose?: (event: React.MouseEvent) => void;
   compact?: boolean;
   closeButtonAriaLabel?: string;
+  entryAnimation?: boolean;
```
shaharzil, line 39: Please add a description to this prop :)

shaharzil (approved):

PR #2568, opened 2024-10-30
feat(Modal): css and html structure improvements, add footer shadow for layouts

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7741641864
https://monday.monday.com/boards/3532714909/pulses/7719009083


packages/core/src/components/ModalNew/ModalContent/__tests__/ModalContent.test.tsx
talkor, line 1: Why delete it?

packages/core/src/components/ModalNew/ModalTopActions/ModalTopActions.module.scss
```diff
@@ -2,4 +2,5 @@
   position: absolute;
   right: var(--top-actions-spacing);
   top: var(--top-actions-spacing);
+  z-index: 1;
```
talkor, line 5: It can interrupt tipseens/dialogs that are opened on top of the modal

packages/core/src/components/ModalNew/layouts/ModalSideBySideLayout/ModalSideBySideLayout.module.scss
talkor, line 1: Very nice using the grid!!

packages/core/src/components/ModalNew/layouts/ModalMedia.module.scss
```diff
@@ -1,8 +1,17 @@
...
+  max-width: 100%;
   max-height: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-shrink: 0;
   overflow: hidden;
+  position: relative;
+
+  > * {
```
talkor, line 11: This is usually bad for performance as css selectors are read from right to left. Any way to avoid it?

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/ModalBasicLayout.tsx
```diff
@@ -4,31 +4,38 @@ import { getTestId } from "../../../../tests/test-ids-utils";
...
+          id={id}
+          data-testid={dataTestId || getTestId(ComponentDefaultTestId.MODAL_NEXT_BASIC_LAYOUT, id)}
+        >
+          <div className={styles.header}>{header}</div>
+          {isContentScrolled && <Divider className={styles.divider} withoutMargin />}
+          <ModalLayoutScrollableContent onScroll={onScroll} className={styles.content}>
+            {content}
+          </ModalLayoutScrollableContent>
+        </Flex>
+        {isContentScrolled && <ModalFooterShadow />}
```
talkor, line 37: Nice

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/ModalBasicLayout.module.scss
```diff
@@ -5,15 +5,29 @@
...
   .divider {
     flex-shrink: 0;
   }
 
-  .content {
-    padding-inline: var(--modal-inline-padding);
+  /**
+   * The footer is optional, so we need to check if it exists before applying padding.
+   * If it exists (or "has" not supported), we apply padding to the content, if it doesn't, we don't apply padding.
+  */
```
talkor, line 20: Why not use context? Or data attrs?

packages/core/src/components/ModalNew/ModalContent/__tests__/ModalContent.test.tsx
YossiSaadi: it is a simple div that does nothing, doesn't have even css 🤷🏼 
logic has moved to inside layouts

packages/core/src/components/ModalNew/layouts/ModalMedia.module.scss
YossiSaadi: I can just target img and video tags instead

packages/core/src/components/ModalNew/ModalTopActions/ModalTopActions.module.scss
YossiSaadi: Hmm... I have no choice
Images overlaps it
and basically anything that the user puts as input

YossiSaadi: what do you suggest?

YossiSaadi: oh I actually have an idea
```
<ModalTopActions
  renderAction={renderHeaderAction}
  color={closeButtonTheme}
  closeButtonAriaLabel={closeButtonAriaLabel}
  onClose={onClose}
/>
{children}
```

-->
```
{children}
<ModalTopActions
  renderAction={renderHeaderAction}
  color={closeButtonTheme}
  closeButtonAriaLabel={closeButtonAriaLabel}
  onClose={onClose}
/>
```

🫠

YossiSaadi: created a task
https://monday.monday.com/boards/3532714909/pulses/7794144599

packages/core/src/components/ModalNew/layouts/ModalMedia.module.scss
YossiSaadi: I'm not sure how Lottie would behave anyway with the current CSS I applied

YossiSaadi: We should probably have Lottie as a dev dep for this kind of cases

talkor: Dev dep? Why? Plus I feel like that this css is creating a side effect for the users, don't you think? Because they can provide an img element and use some css on that element but you will most likely override their css

YossiSaadi: Their css should "win"
I'm using 0,1,0 specificity 
so if they override it, their css would be the one rendered
also, you suggested in a different PR to apply a predefined css
otherwise the ModalMedia is redundant as ModalContent

talkor: I meant predefined css for an image that you provided by a prop, not by children, but this was dropped, no? 
But anyway, I mean that targeting elements inside your component is creating a side effect. I don't think that the css is bad, it just should only affect the elements inside your component and not elements provided by the user. Plus you're applying style that may not fit to all elements like object-fit. I think the best combination would be to have an Image/Video components or maybe Media. This along with the ModalMedia will get along well. But probably out of this scope

YossiSaadi: I need to release the layouts with a solution for that
I cannot leave it with every user doing whatever they want, it would create inconsistency
let's have a talk about it tomorrow

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/ModalBasicLayout.module.scss
YossiSaadi: I've added the padding to always be visible, as the footer shouldn't be omit

PR #2574, opened 2024-11-04
feat(TextArea): Character count and maxLength

feclist:
Towards https://monday.monday.com/boards/3709356818/pulses/7684207310
Design [Figma](https://www.figma.com/design/U27gdArnWEFYL7XA9ZId5D/Submission-view?node-id=3919-12955&node-type=section&t=4707WZFZeZnWo4Wj-0)

Currently there is no visual feedback for the user when there is a max length set on the TextArea component. A character count should be displayed when there is a maxLength set to convey this information.

There are two attributes that influence the behaviour, when `allowExceedingLimit` is set the user is able to type more than the maximum allowed characters, but will see the input in an error state (see 2nd screenshot). When a `maxLength` only is set the user is strictly able to type no more than the allowed length.

image
image

### DIFF
--- a/packages/core/src/components/TextArea/TextArea.module.scss
+++ b/packages/core/src/components/TextArea/TextArea.module.scss
@@ -24,15 +24,7 @@
     color: var(--secondary-text-color);
 
     .helpText {
-      width: fit-content;
       display: flex;
-      flex-grow: 1;
-    }
-
-    .limitText {
-      width: fit-content;
-      flex-shrink: 0;
-      margin-left: auto;
     }
   }
 
--- a/packages/core/src/components/TextArea/TextArea.module.scss
+++ b/packages/core/src/components/TextArea/TextArea.module.scss
@@ -21,10 +21,7 @@
   }
 
   .subTextContainer {
-    display: flex;
-    gap: var(--spacing-xs);
     color: var(--secondary-text-color);
-    align-items: top;
 
     .helpText {
       width: fit-content;
--- a/packages/core/src/components/TextArea/TextArea.module.scss
+++ b/packages/core/src/components/TextArea/TextArea.module.scss
@@ -20,8 +20,22 @@
     }
   }
 
-  .helpText {
+  .subTextContainer {
+    display: flex;
+    gap: var(--spacing-xs);
     color: var(--secondary-text-color);
+    align-items: top;
+
+    .helpText {
+      width: fit-content;
+      display: flex;
+      flex-grow: 1;
+    }
+
+    .limitText {
+      width: fit-content;
+      flex-shrink: 0;
+    }
   }
 
   .textArea {
@@ -71,12 +85,14 @@
       border-color: var(--color-error);
     }
 
-    .helpText {
+    .helpText,
+    .limitText {
       color: var(--color-error);
     }
   }
 
   &.disabled {
+
     .textArea,
     .textArea:hover {
       color: var(--disabled-text-color);
@@ -90,6 +106,7 @@
   }
 
   &.readOnly {
+
     .textArea,
     .textArea:hover {
       background-color: var(--allgrey-background-color);
--- a/packages/core/src/components/TextArea/TextArea.module.scss
+++ b/packages/core/src/components/TextArea/TextArea.module.scss
@@ -26,6 +26,10 @@
     .helpText {
       display: flex;
     }
+
+    .limitText {
+      margin-left: auto;
+    }
   }
 
   .textArea {
--- a/packages/core/src/components/TextArea/TextArea.module.scss
+++ b/packages/core/src/components/TextArea/TextArea.module.scss
@@ -92,7 +92,6 @@
   }
 
   &.disabled {
-
     .textArea,
     .textArea:hover {
       color: var(--disabled-text-color);
@@ -106,7 +105,6 @@
   }
 
   &.readOnly {
-
     .textArea,
     .textArea:hover {
       background-color: var(--allgrey-background-color);
--- a/packages/core/src/components/TextArea/TextArea.module.scss
+++ b/packages/core/src/components/TextArea/TextArea.module.scss
@@ -35,6 +35,7 @@
     .limitText {
       width: fit-content;
       flex-shrink: 0;
+      margin-left: auto;
     }
   }
 
--- a/packages/core/src/components/TextArea/TextArea.tsx
+++ b/packages/core/src/components/TextArea/TextArea.tsx
@@ -5,6 +5,7 @@ import { ComponentDefaultTestId } from "../../tests/constants";
 import styles from "./TextArea.module.scss";
 import { TextAreaProps, TextAreaSize } from "./TextArea.types";
 import Text from "../Text/Text";
+import { Flex } from "../Flex";
 
 const DEFAULT_ROWS: Record<TextAreaSize, number> = {
   small: 3,
@@ -81,19 +82,19 @@ const TextArea = forwardRef(
           aria-describedby={helpTextId ?? undefined}
           onChange={handleOnChange}
         />
-        <div className={cx(styles.subTextContainer)}>
+        <Flex gap={Flex.gaps.XS} className={cx(styles.subTextContainer)}>
           {helpText && (
             <Text className={cx(styles.helpText)} color={Text.colors.INHERIT} id={helpTextId}>
               {helpText}
             </Text>
           )}
           {showCharCount && (
-            <div className={cx(styles.limitText)}>
+            <Flex className={cx(styles.limitText)}>
               {characterCount}
               {maxLength && `/${maxLength}`}
-            </div>
+            </Flex>
           )}
-        </div>
+        </Flex>
       </div>
     );
   }
--- a/packages/core/src/components/TextArea/TextArea.tsx
+++ b/packages/core/src/components/TextArea/TextArea.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef, useState } from "react";
+import React, { forwardRef, useCallback, useState } from "react";
 import cx from "classnames";
 import { getTestId } from "../../tests/test-ids-utils";
 import { ComponentDefaultTestId } from "../../tests/constants";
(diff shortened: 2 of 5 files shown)


packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -34,12 +37,19 @@ const TextArea = forwardRef(
     const numRows = rows || DEFAULT_ROWS[size];
     const helpTextId = helpText && `${id}-help-text`;
 
+    const [characterCount, setCharacterCount] = useState(rest.value?.length || 0);
+
+    const handleOnChange = useCallback((event: React.ChangeEvent<HTMLTextAreaElement>) => {
+      setCharacterCount(event.target.value.length);
+      onChange?.(event);
+    }, [onChange]);
```
feclist, line 45: Doing this as it seems like this component doesn't do any form of controlled input values, seemed overkill to implement it based on this feature.

packages/core/src/components/TextArea/TextArea.types.ts
```diff
@@ -50,4 +50,12 @@ export interface TextAreaProps extends TextAreaNativeInputProps, VibeComponentPr
    * Placeholder text to display when the textarea is empty.
    */
   placeholder?: string;
+  /**
+   * The allowed number of characters.
+   */
+  maxLength?: number;
```
feclist, line 56: Maybe my Storybook is weird, but I assume I need to define these types here for storybook to pick it up

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -64,12 +75,16 @@ const TextArea = forwardRef(
...
           aria-invalid={error}
           aria-describedby={helpTextId ?? undefined}
+          onChange={handleOnChange}
         />
-        {helpText && (
-          <Text className={cx(styles.helpText)} color={Text.colors.INHERIT} id={helpTextId}>
-            {helpText}
-          </Text>
-        )}
+        <div className={cx(styles.subTextContainer)}>
```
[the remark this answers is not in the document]
or-nuri-monday: +1

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -34,12 +37,19 @@ const TextArea = forwardRef(
...
+      onChange?.(event);
+    }, [onChange]);
+
     return (
       <div
         className={cx(
           styles.textAreaWrapper,
           {
-            [styles.error]: error,
+            [styles.error]: error || (maxLength && characterCount > maxLength),
```
or-nuri-monday, line 52: Should we show an error? I think the component should just block new inputs.

packages/core/src/components/TextArea/TextArea.types.ts
```diff
@@ -50,4 +50,12 @@ export interface TextAreaProps extends TextAreaNativeInputProps, VibeComponentPr
...
+  /**
+   * If true, the textarea will allow the user to exceed the character limit.
+   */
+  allowExceedingLimit?: boolean;
```
or-nuri-monday, line 60: Not sure we need this param.
If you set a limit, why would you want to allow it to be exceeded?

packages/core/src/components/TextArea/__tests__/TextArea.test.tsx
```diff
@@ -99,4 +100,57 @@ describe("TextArea", () => {
...
+      // Attempt to add more characters (should be prevented)
+      userEvent.type(input, "a");
+      expect(input).toHaveValue("12345");
+
+      // Remove a character (should be allowed)
+      userEvent.type(input, "{backspace}");
+      userEvent.type(input, "4");
+
+      expect(input).toHaveValue("12344");
+    });
```
or-nuri-monday, line 154: Let's please add a test that validates that we can't input more characters if we met the limit

packages/core/src/components/TextArea/TextArea.module.scss
```diff
@@ -90,11 +106,12 @@
...
+
     .textArea,
     .textArea:hover {
       background-color: var(--allgrey-background-color);
       border-color: transparent;
       resize: none;
     }
   }
-}
+}
```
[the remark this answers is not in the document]
feclist: The default vscode formatter removes empty lines at the end of the file 👀 

Happy to add it back in, if this is an expected default we should look into fixing the formatter 🤔

packages/core/src/components/TextArea/TextArea.types.ts
feclist: Per our conversation this morning, we want to allow the limit to be exceeded to enable a user to type out their original intended text such that they can more easily reduce it afterwards.

Followed the designs and AFAIK design is aligned on this 👍

packages/core/src/components/TextArea/__tests__/TextArea.test.tsx
feclist: https://github.com/mondaycom/vibe/pull/2574/files#diff-ab35f45e7f2a0be5ce7d34564c0a408202d6e74da27f72651a6863fce4e4a29cR110-R124

This test validates that!

packages/core/src/components/TextArea/TextArea.tsx
feclist: Once our character count exceeds the maxLength (which only happens when `allowExceedingLimit` is set to true) we display the input in an error state. Pulled this styling from the Figma designs.

packages/core/src/components/TextArea/TextArea.tsx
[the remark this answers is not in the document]
feclist: If both of you are very keen on using `<Flex />` here I'll add it, but it seems like a more verbose component than is actually useful.

I will have to write a scss class for the container regardless of the display rules, so might as well put all the css in one place.

Again, happy to get pushback on this :)

packages/core/src/components/TextArea/__tests__/TextArea.test.tsx
talkor, line 1: Nice tests!

talkor (changes requested):
Thanks @feclist ! Again, really cool addition. Some of the comments from #2576 are probably relevant here as well

packages/core/src/components/TextArea/TextArea.types.ts
```diff
@@ -50,4 +50,16 @@ export interface TextAreaProps extends TextAreaNativeInputProps, VibeComponentPr
    * Placeholder text to display when the textarea is empty.
    */
   placeholder?: string;
+  /**
+   * The allowed number of characters.
+   */
+  maxLength?: number;
+  /**
+   * If true, the textarea will allow the user to exceed the character limit.
```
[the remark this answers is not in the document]
feclist: textarea is not capitalised in any of the other comments in this file, happy to change all of them though

packages/core/src/components/TextArea/TextArea.module.scss
```diff
@@ -20,8 +20,23 @@
     }
   }
 
-  .helpText {
+  .subTextContainer {
+    display: flex;
+    gap: var(--spacing-xs);
     color: var(--secondary-text-color);
+    align-items: top;
```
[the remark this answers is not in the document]
feclist: Awkward 😂

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -26,20 +28,37 @@ const TextArea = forwardRef(
...
       resize = true,
+      showCharCount = false,
       ...rest
     }: TextAreaProps,
     ref: React.ForwardedRef<HTMLTextAreaElement>
   ) => {
     const numRows = rows || DEFAULT_ROWS[size];
     const helpTextId = helpText && `${id}-help-text`;
+    const allowExceedingMaxLengthTextId = allowExceedingMaxLength && `${id}-allow-exceeding-max-length`;
+    const inErrorState = error || (maxLength && value?.length > maxLength);
```
talkor, line 44: Nitpick:
```suggestion
    const isErrorState = error || (maxLength && value?.length > maxLength);
```

packages/core/src/components/TextArea/TextArea.types.ts
```diff
@@ -7,7 +7,23 @@ type TextAreaNativeInputProps = Omit<
   "role" | "aria-describedby" | "aria-invalid"
 >;
 
-export interface TextAreaProps extends TextAreaNativeInputProps, VibeComponentProps {
+/**
+ * This type ensures that allowExceedingMaxLength is only used when maxLength is provided.
+ */
+type MaxLengthProps =
```
talkor, line 13: As discussed, let's revert and use jsdoc comment to make the relation between these prop clearer

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -56,20 +75,32 @@ const TextArea = forwardRef(
...
           ref={ref}
+          value={value}
           disabled={disabled}
           readOnly={readOnly}
           required={required}
           rows={numRows}
           className={cx(styles.textArea, [styles[size]], { [styles.resize]: resize })}
           aria-invalid={error}
-          aria-describedby={helpTextId ?? undefined}
+          aria-describedby={[helpTextId, allowExceedingMaxLengthTextId].filter(id => !!id).join(" ") || undefined}
```
talkor, line 87: Worth memoizing it maybe

packages/core/src/components/TextArea/TextArea.tsx
```diff
@@ -56,20 +75,32 @@ const TextArea = forwardRef(
...
-          </Text>
-        )}
+        <Flex gap={Flex.gaps.XS} justify={Flex.justify.SPACE_BETWEEN} className={cx(styles.subTextContainer)}>
+          {helpText && (
+            <Text className={cx(styles.helpText)} color={Text.colors.INHERIT} id={helpTextId}>
+              {helpText}
+            </Text>
+          )}
+          {showCharCount && (
+            <Flex className={cx(styles.limitText)}>
```
talkor, line 97: This can only be a fragment, as the HiddenText has no dimensions

talkor (approved):
Excellent! Awesome addition! 🥳

PR #2581, opened 2024-11-07
fix: lazy loading icons name

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7802963361


YossiSaadi (approved):
Awesome!

PR #2585, opened 2024-11-13
chore: add d.ts file compilation to testkit

uziab:
- [ ] I have read the [Contribution Guide](../packages/core/CONTRIBUTING.md) for this project.

Resolves #


packages/testkit/package.json
```diff
@@ -17,13 +27,14 @@
...
+    "build": "tsc && tsc --project tsconfig.esm.json",
     "start-server": "yarn lerna run storybook --scope=monday-ui-react-core"
   },
   "bugs": {
     "url": "https://github.com/mondaycom/vibe/issues"
   },
   "devDependencies": {
-    "@playwright/test": "1.45.3"
+    "@playwright/test": "1.45.3",
+    "typescript": "^4.4.3"
```
laviomri, line 38: any reason for not using a newer version?

laviomri (approved):
looking good!

packages/testkit/package.json
uziab: used the same version as in the root project. I don't see a reason why not to use a newer version. @talkor do you?

packages/testkit/tsconfig.esm.json
YossiSaadi, line 1: same question, why the need for cjs came up? is this really needed?

packages/testkit/package.json
```diff
@@ -6,6 +6,16 @@
...
   ],
+  "module": "dist/esm/index.js",
+  "types": "dist/index.d.ts",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./dist/esm/index.js",
+      "types": "./dist/index.d.ts"
+    },
+    "./package.json": "./package.json"
```
YossiSaadi, line 17: this would create a @vibe/testkit/package.json entry
is this really needed?

YossiSaadi: if you just want to export the package.json, it gets exported by default anyway, so no need to declare it in `files` or in `exports`

YossiSaadi: see also https://www.npmjs.com/package/@vibe/testkit?activeTab=code

packages/testkit/package.json
```diff
@@ -17,13 +27,14 @@
...
-    "build": "tsc",
+    "build": "tsc && tsc --project tsconfig.esm.json",
     "start-server": "yarn lerna run storybook --scope=monday-ui-react-core"
   },
   "bugs": {
     "url": "https://github.com/mondaycom/vibe/issues"
   },
   "devDependencies": {
-    "@playwright/test": "1.45.3"
+    "@playwright/test": "1.45.3",
```
YossiSaadi, line 37: is `playwright`'s version locked by accident or is that intended? 
```suggestion
    "@playwright/test": "^1.45.3",
```

packages/testkit/tsconfig.json
```diff
@@ -1,28 +1,22 @@
...
-      "declaration": false,
-      "downlevelIteration": true,
-      "experimentalDecorators": true,
-      "noImplicitOverride": true,
-      "noImplicitReturns": true,
-    },
-    "include": [ "buttons", "inputs", "navigation", "pickers", "popover", "text", "utils", "BaseElement.ts", "index.ts"]
-  }
+  "compilerOptions": {
+    "lib": ["ESNext"],
```
YossiSaadi, line 3: Don't you need also "dom" in here?

YossiSaadi: for e2e in general I believe you do
I'm not sure for testkit but I think so as well

packages/testkit/tsconfig.json
```diff
@@ -1,28 +1,22 @@
...
+    "target": "es6",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "strict": true,
+    "experimentalDecorators": true,
```
YossiSaadi, line 9: really needed? do you use it?

packages/testkit/tsconfig.json
```diff
@@ -1,28 +1,22 @@
...
-      "noImplicitReturns": true,
-    },
-    "include": [ "buttons", "inputs", "navigation", "pickers", "popover", "text", "utils", "BaseElement.ts", "index.ts"]
-  }
+  "compilerOptions": {
+    "lib": ["ESNext"],
+    "target": "es6",
+    "module": "commonjs",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
```
YossiSaadi, line 7: do you import json files in your testkit? needed?

packages/testkit/tsconfig.json
```diff
@@ -1,28 +1,22 @@
...
+    "strict": true,
+    "experimentalDecorators": true,
+    "emitDecoratorMetadata": true,
```
YossiSaadi, line 10: same as above, really needed?

YossiSaadi (changes requested):
Had a few comments
also, in general, you're not using eslint, prettier, etc. can you add it sometime to the library? it might have ts, lint, or format errors that we're not catching currently

packages/testkit/package.json
laviomri: @YossiSaadi yes it should work this way, but that means that when you'll upgrade the TS version you'll have to deal with all packages in one go. If you want, you CAN set each package using its own TS version. This will allow you a more flexible approach

YossiSaadi: yep, exactly
we do want to have the root with packages that should have the same version across all the repos, but were not there yet
eslint, typescript, prettier, etc.

but I think for now, @uziab can use the one from root (or remove the declaration entirely, up to him to decide, not critical for me)

packages/testkit/package.json
```diff
@@ -17,13 +27,14 @@
   },
   "scripts": {
     "test:e2e": "npx playwright test",
-    "build": "tsc",
+    "build": "tsc && tsc --project tsconfig.esm.json",
```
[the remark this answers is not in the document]
uziab: we have clients that use commonjs. i.e the monolith. So we need it for now.

packages/testkit/package.json
uziab: removed

packages/testkit/package.json
uziab: @YossiSaadi yes it's locked on purpose since it is the version used currently in some of our repo's. we'd need to have a separate task for bumping versions, synchronously to avoid mismatch.

packages/testkit/tsconfig.json
uziab: No we wouldn't need "dom" because  the DOM interaction is abstracted by testkit itself. Playwright directly handles actions like clicking elements, asserting visibility, or navigating pages, without needing direct access to browser-native APIs.

packages/testkit/package.json
uziab: removed typescript

packages/testkit/tsconfig.json
uziab: removed

packages/testkit/tsconfig.json
uziab: removed

packages/testkit/package.json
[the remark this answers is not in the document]
YossiSaadi: the monolith can use esm as far as I know
@talkor, does it did any trouble to you with @vibe/core when you used esm only?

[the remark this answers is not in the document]
talkor: Nope, all good

packages/testkit/package.json
YossiSaadi: got ya, cool 💯

packages/testkit/package.json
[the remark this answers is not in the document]
YossiSaadi: can you separate the cases? @uziab 
have a release with esm only and add cjs if the need arises?

packages/testkit/__TESTS__/dropdown.spec.js
```diff
@@ -13,7 +13,9 @@ test.describe("dropdown Class with Storybook", () => {
   });
 
   test("set dropdown value", async ({ page }) => {
-    await page.waitForTimeout(3000);
+    // extended wait for interaction test to finish
+    await page.waitForTimeout(10000);
+    await DropDown.inputField.setText("");
```
YossiSaadi, line 18: please notice I've disabled this test for now in master branch, since it caused releases to fail

yarn.lock
YossiSaadi, line 1: not sure why the yarn.lock was changed here that significantly

YossiSaadi: please try to rollback to the version as in master and run `yarn` again, something isn't right here

packages/testkit/vibe-testkit-1.0.2.tgz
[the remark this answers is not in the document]
uziab: done

yarn.lock
uziab: done

packages/testkit/package.json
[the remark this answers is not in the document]
uziab: Sorry no, The monolith e2e project can't handle esm, so we would need cjs.

YossiSaadi (approved):
💯

PR #2588, opened 2024-11-15
fix(TextArea): Error state not correctly set when maxlength is exceeded

feclist:
Follow up on https://github.com/mondaycom/vibe/pull/2574

Somewhere in the process of implementing the above PR the error state wasn't triggering anymore once the max length was exceeded (when it is allowed to do so).

Before:
image

After:
image

This PR fixes the conditional logic and uses the `characterCount`  state variable to more reliably check against the `maxLength` prop.


talkor (commented):
Let's add a test for that please, to prevent future breaking of this feature

talkor (approved):

PR #2600, opened 2024-11-20
feat(Search): allow tracking Enter key press

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7890250116

### DIFF
--- a/packages/core/src/components/Search/Search.tsx
+++ b/packages/core/src/components/Search/Search.tsx
@@ -35,6 +35,7 @@ const Search = forwardRef(
       onFocus,
       onBlur,
       onClear,
+      onEnterKey,
       className,
       ariaExpanded,
       ariaHasPopup,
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Search/Search.tsx
```diff
@@ -35,6 +35,7 @@ const Search = forwardRef(
       onFocus,
       onBlur,
       onClear,
+      onEnterKey,
```
shlomitc, line 38: why aren't we just exposing `onKeyDown`?

YossiSaadi: I actually thought about doing it, but maybe we should make this more straightforward for cases like Enter, which is the one the team currently needs. I can't think of another key that someone might want to check for stuff like Search field

packages/core/src/components/Search/Search.tsx
```diff
@@ -53,14 +54,11 @@ const Search = forwardRef(
...
-      if (disabled) {
-        return;
-      }
-
+      if (disabled) return;
       inputRef.current?.focus?.();
       clearValue();
       onClear?.();
-    }, [disabled, clearValue]);
+    }, [disabled, clearValue, onClear]);
```
talkor, line 61: Needed?

talkor (approved):

packages/core/src/components/Search/Search.tsx
YossiSaadi: @shlomitc after reconsideration, went with your suggestion at the end, thanks!

PR #2601, opened 2024-11-20
chore: standardize testkit package with lint, config, plugins

YossiSaadi:


YossiSaadi:
@uziab can you please go over this and see which stuff you can fix where I've added eslint disable? stuff like `any` usage, `no-wait-for-timeout`, `no-constant-condition`, `prefer-web-first-assertions`, etc.

PR #2608, opened 2024-11-23
feat(Dropdown): add inputValue and blurInputOnSelect properties to allow editing options

eliyapl:
In this PR, I added two properties to the React Select component:  
1. inputValue
2. blurInputOnSelect
These properties allow the user to edit their search term after they have already performed a search. Currently, when the input field gains focus, the latest search is cleared, and the user can only start a new search. With these two properties, the user will have the option to edit their existing search term.
Properties added:
1. inputValue: Controls the current value of the input field programmatically, allowing the user to modify their search term.
2. blurInputOnSelect: If true, the input field will lose focus when an option is selected, providing better control over the input behavior.

Resolves #


talkor (approved):
Thanks!

PR #2619, opened 2024-12-02
fix(table): fix scroll delay in header on virtualized table

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7963926689

### DIFF
--- a/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
+++ b/packages/core/src/components/Table/TableHeader/TableHeader.module.scss
@@ -4,7 +4,7 @@
   position: sticky;
   top: 0;
   z-index: 2;
-  background-color: inherit;
+  background-color: var(--primary-background-color);
   min-width: 100%;
   width: fit-content;
 
--- a/packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
+++ b/packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
@@ -10,6 +10,7 @@ import { ComponentDefaultTestId } from "../../../tests/constants";
 import { RowHeights } from "../Table/TableConsts";
 import AutoSizer, { Size as AutoSizerSize } from "react-virtualized-auto-sizer";
 import { useTableRowMenu } from "../context/TableRowMenuContext/TableRowMenuContext";
+import { TableColumn } from "../Table/Table";
 
 export type TableVirtualizedRow = Record<string, unknown> & { id: string };
 
@@ -18,15 +19,27 @@ export interface TableVirtualizedBodyProps<T extends TableVirtualizedRow = Table
   items: T[];
   rowRenderer: (item: T) => JSX.Element;
   onScroll?: (horizontalScrollDirection: ScrollDirection, scrollTop: number, scrollUpdateWasRequested: boolean) => void;
+  columns?: TableColumn[];
+  headerRenderer?: (columns: TableColumn[]) => JSX.Element;
 }
 
 const TableVirtualizedBody = forwardRef(
   <T extends TableVirtualizedRow = TableVirtualizedRow>(
-    { items, rowRenderer, onScroll, id, className, "data-testid": dataTestId }: TableVirtualizedBodyProps<T>,
+    {
+      items,
+      rowRenderer,
+      onScroll,
+      columns,
+      headerRenderer,
+      id,
+      className,
+      "data-testid": dataTestId
+    }: TableVirtualizedBodyProps<T>,
     ref: React.ForwardedRef<HTMLDivElement>
   ) => {
     const { size, virtualizedListRef, onVirtualizedListScroll, markTableAsVirtualized } = useTable();
     const { resetHoveredRow } = useTableRowMenu();
+    const virtualizedWithHeader = !!columns && !!headerRenderer;
 
     const handleOuterScroll = useCallback(
       (e: Event) => {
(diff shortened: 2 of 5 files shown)


packages/core/src/components/Table/TableHeader/TableHeader.module.scss
```diff
@@ -4,7 +4,7 @@
   position: sticky;
   top: 0;
   z-index: 2;
-  background-color: inherit;
+  background-color: var(--primary-background-color);
```
YossiSaadi, line 7: need to check how this one behaves with the sticky - on scrolling horizontally and vertically, and hovering a row or a table header cell, doesn't it causing issues?

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
```diff
@@ -95,12 +120,22 @@ const TableVirtualizedBody = forwardRef(
               <List
                 itemSize={RowHeights[size]}
                 height={height}
-                itemCount={items.length}
+                itemCount={virtualizedWithHeader ? items.length + 1 : items.length}
```
YossiSaadi, line 123: nice trick

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
```diff
@@ -61,30 +74,42 @@ const TableVirtualizedBody = forwardRef(
...
         });
       },
-      [items, rowRenderer]
+      [items, rowRenderer, virtualizedWithHeader]
     );
 
     useEffect(() => {
-      markTableAsVirtualized();
-    }, [markTableAsVirtualized]);
+      if (!virtualizedWithHeader) markTableAsVirtualized();
```
YossiSaadi, line 101: why does it matter though?

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
```diff
@@ -61,30 +74,42 @@ const TableVirtualizedBody = forwardRef(
...
     );
 
     const itemRenderer = useCallback<ComponentType<ListChildComponentProps<TableVirtualizedRow>>>(
       ({ index, style: { width: _width, ...style } }) => {
-        const currentItem = items[index];
+        if (virtualizedWithHeader && index === 0) {
+          return null; //placeholder for virtualized with header
+        }
+        const currentIndex = virtualizedWithHeader ? index - 1 : index;
+        const currentItem = items[currentIndex];
```
YossiSaadi, line 89:
```suggestion
        if (virtualizedWithHeader && index === items.length) {
          return null; //placeholder for virtualized with header
        }
        const currentItem = items[index];
```

would this work as well? or do you have to make the header as the index 0? (I know it makes sense it's index 0, but is `innerElementType` just puts it before the first index anyway?)

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
```diff
@@ -95,12 +120,22 @@ const TableVirtualizedBody = forwardRef(
...
+                innerElementType={
+                  virtualizedWithHeader
+                    ? forwardRef(({ children, ...rest }, ref) => (
+                        <div ref={ref} {...rest}>
+                          {headerRenderer(columns)}
+                          {children}
+                        </div>
+                      ))
+                    : undefined
+                }
```
YossiSaadi, line 138: is this just lets you wrap everything that VirtualizedList is about to render?
also, do we absolutely need the wrapping div? and what's `rest`? is it coming from VirtualizedList?

YossiSaadi: Can we try Fragments there maybe?
or is it a must? (a yes is a legitimate answer 🤓)

YossiSaadi (approved):
Thank you.
I know how much frustration it caused and I really appreciate the work you've put into that one 👑

YossiSaadi:
Would also love to try it in Storybook if you have a live example to try locally

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
rivka-ungar: `innerElementType` puts it before the first index but if we don't keep an empty slot for it on index 0 it overlaps the first row.

YossiSaadi: got ya

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
rivka-ungar: Now that we don't need the sync for the header on the scroll it should act the same as non virtualized table, so there is no need to mark it as virtualized

packages/core/src/components/Table/TableHeader/TableHeader.module.scss
rivka-ungar: I checked on all scroll and hover states and it seems to be ok

packages/core/src/components/Table/TableVirtualizedBody/TableVirtualizedBody.tsx
rivka-ungar: The div exists in the default behavior of VirtualizedList, so we need to include it here to maintain the current structure. The rest props come from VirtualizedList

rivka-ungar:
> Would also love to try it in Storybook if you have a live example to try locally

I changed the current virtualized story to the new implementation

YossiSaadi:
> > Would also love to try it in Storybook if you have a live example to try locally
> 
> I changed the current virtualized story to the new implementation

Went over this, looks good 💪🏼

PR #2622, opened 2024-12-03
fix(Button): fix secondary disabled border color

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7801661751


YossiSaadi (approved):
It's weird that the request was for "text" color 🤔

PR #2625, opened 2024-12-04
fix: support vitest

talkor:


YossiSaadi (approved):
isn't `process.env.NODE_ENV === "test"` also in vitest?

talkor:
> isn't `process.env.NODE_ENV === "test"` also in vitest?

Yes but vitest can't handle `require`

YossiSaadi:
> > isn't `process.env.NODE_ENV === "test"` also in vitest?
> 
> Yes but vitest can't handle `require`

so maybe worth leaving the NODE_ENV one as well

YossiSaadi:
btw its really weird that this code should lay in the component itself, is it really a must in order to do lazy load?

talkor:
> > > isn't `process.env.NODE_ENV === "test"` also in vitest?
> > 
> > 
> > Yes but vitest can't handle `require`
> 
> so maybe worth leaving the NODE_ENV one as well

why? It was for jest because jest isn't very good with esm while vitest is quite the opposite. This check covers both

talkor:
> btw its really weird that this code should lay in the component itself, is it really a must in order to do lazy load?

unfortunately yes, do you have other idea?

PR #2628, opened 2024-12-05
fix: menu item a11y for disable items

doronbrikman:

https://monday.monday.com/boards/5751390280/pulses/5751647586

Resolves #
It will let screen readers reach disabled items and be aware of the disable reason


packages/core/src/components/Menu/Menu/utils/utils.ts
```diff
@@ -2,7 +2,7 @@ import { ReactElement } from "react";
 import { MenuChild } from "../MenuConstants";
 
 export function isMenuChildSelectable(child: MenuChild): boolean {
-  return !!child.type.isSelectable && !child.props.disabled;
+  return !!child.type.isSelectable;
```
rivka-ungar, line 5: Since we only want it to be selectable if it is disabled and has a tooltip (but should not be selectable when it is disabled without a tooltip), the check for disabled should stay, and there should be an additional check if has tooltip.

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -56,6 +56,8 @@ export interface MenuItemProps extends VibeComponentProps {
   splitMenuItem?: boolean;
   "aria-label"?: AriaAttributes["aria-label"];
   submenuPosition?: SubmenuPosition;
+  tooltipId?: string;
```
rivka-ungar, line 59: Instead of adding an id, can be implemented using the id from `tooltipProps`

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -56,6 +56,8 @@ export interface MenuItemProps extends VibeComponentProps {
...
+  shouldShowTooltip?: boolean;
```
rivka-ungar, line 60: Do we need this prop? there is already logic in the component for this [here](https://github.com/mondaycom/vibe/blob/669b9483da1b81348469063dbae04c8634fd985c/packages/core/src/components/Menu/MenuItem/MenuItem.tsx#L101).

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -130,6 +133,8 @@ const MenuItem: VibeComponent<MenuItemProps | MenuItemTitleComponentProps> & {
           className={className}
           disabled={disabled}
           selected={selected}
+          tooltipId={tooltipProps?.id}
+          shouldShowTooltip={!!shouldShowTooltip}
```
rivka-ungar, line 137: In order to simplify it it might be clearer to add `aria-disabled` and `aria-describedby` props instead of the tooltip specific props so that the logic can be done here.

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -121,6 +123,7 @@ const MenuItem: VibeComponent<MenuItemProps | MenuItemTitleComponentProps> & {
         content={shouldShowTooltip ? finalTooltipContent : null}
         position={tooltipPosition}
         showDelay={tooltipShowDelay}
+        childId={baseMenuProps.id}
```
rivka-ungar, line 126: Tooltip has an id prop do we need to add a `childId`?

packages/core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -205,7 +206,7 @@ export default class Tooltip extends PureComponent<TooltipProps> {
         className={cx(styles.tooltip, getStyle(styles, camelCase(theme)), className)}
       >
         {image && <img className={styles.image} src={image} alt="" />}
-        <div className={cx(styles.content)}>
+        <div id={id || `${childId}-tooltip`} className={cx(styles.content)}>
```
rivka-ungar, line 209: Would the id be enough? why need to add `childId`?

rivka-ungar (commented):
Thanks! added some questions. Also would be great if you can add some tests to verify the new behavior.

talkor:
@doronbrikman is this still relevant?

PR #2630, opened 2024-12-05
fix(table): avoid unnecessary re-rendering if TableRowMenu doesn't exist

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7889513842

### DIFF
--- a/packages/core/src/components/Table/context/TableRowMenuContext/TableRowMenuContext.tsx
+++ b/packages/core/src/components/Table/context/TableRowMenuContext/TableRowMenuContext.tsx
@@ -7,6 +7,7 @@ export const TableRowMenuProvider = ({ value, children }: TableRowMenuProviderPr
   const { tableRootRef, hoveredRowRef, isMenuOpen, resetHoveredRow, setHoveredRowRef, setIsMenuOpen } = value;
   const rowMouseLeaveTimeoutId = useRef<ReturnType<typeof setTimeout>>(null);
   const [menuButtonPosition, setMenuButtonPosition] = useState(0);
+  const hasMenuRef = useRef<boolean | null>(null);
 
   const resetTimeout = useCallback(() => {
     clearTimeout(rowMouseLeaveTimeoutId.current);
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Table/context/TableRowMenuContext/TableRowMenuContext.tsx
```diff
@@ -15,12 +16,18 @@ export const TableRowMenuProvider = ({ value, children }: TableRowMenuProviderPr
...
     (rowRef: React.RefObject<HTMLDivElement>) => {
-      if (isMenuOpen || hoveredRowRef?.current === rowRef.current) return;
+      if (hasMenuRef.current === false || isMenuOpen || hoveredRowRef?.current === rowRef.current) return;
+
       resetTimeout();
       setHoveredRowRef(rowRef);
       const tableRootTop = tableRootRef.current.getBoundingClientRect().top;
       const rowTop = rowRef.current.getBoundingClientRect().top;
       setMenuButtonPosition(rowTop - tableRootTop);
+      if (hasMenuRef.current === null) {
```
YossiSaadi, line 26: || true?

YossiSaadi: or even
```suggestion
      if (hasMenuRef.current !== false) {
```

YossiSaadi: to avoid re-entering this if

rivka-ungar: Only the first time it is null. After that it is set to true/false. So checking if it is null will run only the first time. If I check not false than the check will run every time for the tables that don't have a menu

YossiSaadi: ohhh you're right actually
sorry :)

YossiSaadi (approved):
Awesome work!

PR #2634, opened 2024-12-09
feat(Modal): use createPortal and LayerProvider in Modal

YossiSaadi:
- use createPortal
- allow bypassing portal target
	- either as a string selector
	- a React ref
	- document fragment
	- element
- use LayerProvider
- use z-index

https://monday.monday.com/boards/3532714909/pulses/7991096353


packages/core/src/components/ModalNew/Modal/Modal.module.scss
```diff
@@ -1,17 +1,21 @@
 .overlay {
+  --monday-modal-z-index: 10000;
   position: fixed;
   inset: 0;
   background-color: rgba(41, 47, 76, 0.7);
+  z-index: var(--monday-modal-z-index);
 }
 
 .modal {
+  --monday-modal-z-index: 10000;
```
talkor, line 10: You can use a scss variable for the value and assign it to a variable if that helps

packages/core/src/components/ModalNew/hooks/usePortalTarget/usePortalTarget.ts
```diff
@@ -0,0 +1,43 @@
...
+
+      if (portalTarget instanceof Element || portalTarget instanceof DocumentFragment) {
+        return portalTarget;
+      }
+
+      if (portalTarget.current instanceof Element) {
+        return portalTarget.current;
+      }
+
+      console.warn("Invalid portalTarget provided. Falling back to document.body.");
```
talkor, line 30: Do we need to print it in prod?

packages/core/src/components/ModalNew/hooks/usePortalTarget/usePortalTarget.ts
```diff
@@ -0,0 +1,43 @@
...
+      return document.body;
+    };
+
+    const target = resolveTarget();
+    if (target !== resolvedTarget) {
```
talkor, line 35: This is a bit confusing haha

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/__stories__/ModalBasicLayout.stories.tsx
```diff
@@ -31,14 +30,17 @@ export default {
   subcomponents: { ModalBasicLayout, ModalHeader, ModalContent, ModalFooter, ModalFooterWizard, TransitionView },
   argTypes: metaSettings.argTypes,
   decorators: metaSettings.decorators,
-  tags: ["internal"]
+  tags: ["internal"],
+  parameters: {
+    chromatic: {}
+  }
```
talkor, line 36: What's this for?

packages/core/src/components/ModalNew/Modal/Modal.tsx
```diff
@@ -32,13 +35,18 @@ const Modal = forwardRef(
       onClose = () => {},
       anchorElementRef,
       alertModal,
+      container = document.body,
```
talkor, line 38: I don't think the name is good enough, when you think about it the modal is not under a certain container but it takes the whole page

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/__stories__/ModalBasicLayout.stories.tsx
```diff
@@ -31,14 +30,17 @@ export default {
...
 } satisfies Meta<typeof Modal>;
 
 export const Overview: Story = {
   decorators: [(Story, context) => withOpenedModalPreview(Story, { isDocsView: context.viewMode === "docs" })],
-  render: (args, { show, setShow }) => {
+  render: (args, { show, setShow, container }) => {
     return (
-      <Modal id="modal-basic" show={show} size="medium" onClose={() => setShow(false)} {...args}>
+      <Modal id="modal-basic" show={show} size="medium" onClose={() => setShow(false)} container={container} {...args}>
```
talkor, line 43: Why do you need to pass it? Isn't it using the default document.body?

packages/core/src/components/ModalNew/Modal/Modal.module.scss
YossiSaadi: oh good point

YossiSaadi: it actually on 2nd though won't work
as I cannot use the var in overlay if I set it in modal's div, and I cannot use it in modal's div if I set it in the overlay

packages/core/src/components/ModalNew/hooks/usePortalTarget/usePortalTarget.ts
YossiSaadi: will make a tiny change here

packages/core/src/components/ModalNew/hooks/usePortalTarget/usePortalTarget.ts
talkor: I think it's better than printing it in prod, no? We did something similar in deprecateMessage

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/__stories__/ModalBasicLayout.stories.tsx
YossiSaadi: a mistake :)
I want to solve it in another PR, I have an issue where Chromatic captures the Modal story too early

packages/core/src/components/ModalNew/Modal/Modal.tsx
YossiSaadi: I followed MUI's convention (they use container), also we use the same convention in our Dialog and Tooltip components..
what do you suggest?

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/__stories__/ModalBasicLayout.stories.tsx
YossiSaadi: I don't want it to default to document.body, I want it to be under the preview container, I have no choice but to expose it in the stories

packages/core/src/components/ModalNew/Modal/Modal.tsx
talkor: portalTarget, but you make a good point about being consistent, so let's keep it this way

packages/core/src/components/ModalNew/layouts/ModalBasicLayout/__stories__/ModalBasicLayout.stories.tsx
talkor: Got it
Would a decorator work for this? Because it's irrelevant for the stories

talkor (approved):

PR #2638, opened 2024-12-09
fix(List): fix ListTitle getting focus on initialization

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8011242971?userId=59970142


packages/core/src/components/List/__tests__/__snapshots__/List.test.js.snap
```diff
@@ -29,7 +29,7 @@ exports[`List renders correctly with list items 1`] = `
     onFocus={[Function]}
     onMouseEnter={[Function]}
     role="option"
-    tabIndex={0}
+    tabIndex={-1}
```
YossiSaadi, line 32: the bug already existed in the snapshot?? that shows the amount of trust you can get from this kind of testing 😓

YossiSaadi (approved):
💯

PR #2642, opened 2024-12-11
fix(Dropdown): fix valueRenderer return type

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8028440049


talkor (approved):

YossiSaadi:
waiting for confirmation from the dev that the prerelease solved it

PR #2643, opened 2024-12-11
feat(Tooltip): be able to change max width

talkor:
https://monday.monday.com/boards/3532714909/pulses/7980028465

### DIFF
--- a/packages/core/src/components/Tooltip/Tooltip.tsx
+++ b/packages/core/src/components/Tooltip/Tooltip.tsx
@@ -1,7 +1,7 @@
 /* eslint-disable react/jsx-props-no-spreading */
 import { camelCase, isFunction } from "lodash-es";
 import cx from "classnames";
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -196,7 +200,10 @@ export default class Tooltip extends PureComponent<TooltipProps> {
     }
 
     return (
-      <div style={style} className={cx(styles.tooltip, getStyle(styles, camelCase(theme)), className)}>
+      <div
+        style={{ ...style, "--tooltip-max-width": `${maxWidth}px` } as CSSProperties}
```
rivka-ungar, line 204: If maxWidth is not sent wont it cause it to be `undefinedpx`?

talkor: oops, good catch! fixing

rivka-ungar (approved):

PR #2648, opened 2024-12-12
fix(List): getting wrong tabIndex on initialization

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8011242971


YossiSaadi (approved):
Thanks!! 👑

PR #2652, opened 2024-12-15
feat(Tipseen): allow referenceWrapperClassName prop for component

stasshw:
Wrapping the element with a Tipseen wraps it in an empty span, which interferes with the CSS. Passing the missing property to the inner component ensures the correct styling is applied.

https://monday.monday.com/boards/5703104356/pulses/8039110583

image

**another** way to go would be to add a default class name here: https://github.com/stasshw/vibe/blob/24828d7f2ffec00084775f56322518cc04c9a2b8/packages/core/src/components/Dialog/Dialog.tsx#L519

### DIFF
--- a/packages/core/src/components/Tipseen/Tipseen.tsx
+++ b/packages/core/src/components/Tipseen/Tipseen.tsx
@@ -43,6 +43,7 @@ export interface TipseenProps extends VibeComponentProps {
   width?: number;
   moveBy?: MoveBy;
   hideWhenReferenceHidden?: boolean;
+  referenceWrapperClassName?: string;
   /**
    * when false, the arrow of the tooltip is hidden
    */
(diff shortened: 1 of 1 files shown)


stasshw:
another way to go would be to add a default class name here: https://github.com/stasshw/vibe/blob/24828d7f2ffec00084775f56322518cc04c9a2b8/packages/core/src/components/Dialog/Dialog.tsx#L519

YossiSaadi:
Hey @stasshw, what do you mean about a default class?

stasshw:
> Hey @stasshw, what do you mean about a default class?

@YossiSaadi just add any class even without a prop from outside (currently it's a `span` without any class)

YossiSaadi:
> > Hey @stasshw, what do you mean about a default class?
> 
> @YossiSaadi just add any class even without a prop from outside (currently it's a `span` without any class)

you mean so you'll be able to "target" the element in a non-moduled css?
it can work, but it's probably a less good solution

YossiSaadi (approved):

stasshw:
> > > Hey @stasshw, what do you mean about a default class?
> > 
> > 
> > @YossiSaadi just add any class even without a prop from outside (currently it's a `span` without any class)
> 
> you mean so you'll be able to "target" the element in a non-moduled css? it can work, but it's probably a less good solution

ok thank you

PR #2661, opened 2024-12-18
fix(Dropdown): fix menuRenderer type

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8075255489

### DIFF
--- a/packages/core/src/components/Dropdown/Dropdown.types.ts
+++ b/packages/core/src/components/Dropdown/Dropdown.types.ts
@@ -150,7 +150,7 @@ export interface DropdownComponentProps extends CustomMenuBaseProps, CustomOptio
   /**
    * custom menu render function
    */
-  menuRenderer?: React.ReactElement;
+  menuRenderer?: React.ReactElement | ((props: CustomMenuProps) => React.ReactElement);
   /**
    * Default placement of the Dropdown menu in relation to its control. Use "auto" to flip the menu when there isn't enough space below the control.
    */


packages/core/src/components/Dropdown/Dropdown.types.ts
```diff
@@ -150,7 +150,7 @@ export interface DropdownComponentProps extends CustomMenuBaseProps, CustomOptio
   /**
    * custom menu render function
    */
-  menuRenderer?: React.ReactElement;
+  menuRenderer?: React.ReactElement | ((props: CustomMenuProps) => React.ReactElement);
```
YossiSaadi, line 153: I believe the type here is a bit more complex
can you check menu.jsx file?
its `rendererProps`.
everything which is passed to `{Renderer && Renderer(rendererProps)}`
so apparently we're also passing by mistake `innerProps`, the `onScroll`, etc.

what a mess 🤦🏼‍♂️
you can give it an any in the meantime if stuff gets too complicated (see what I did in `valueRenderer`, they needed the `DropdownOption` stuff, so I had to add it, but basically `DropdownOption` is `any`, so the all `props` became `any` at the bundle, so I've worked hard for nothing 🙃)

rivka-ungar: I agree that it is a mess, but itsn't it still better to keep it this way then changing it to `any`?

YossiSaadi: they would have TS issues in the current way if they try to consume something which isn't in your type
we can try that and if they complain we'll release a fix

YossiSaadi (approved):

packages/core/src/components/Dropdown/Dropdown.types.ts
rivka-ungar: I tested it with current implementations and it worked properly, I will keep an eye on it and change if necessary

PR #2666, opened 2024-12-18
feat(Tooltip): add dir prop

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8078910437

### DIFF
--- a/packages/core/src/components/Tooltip/Tooltip.tsx
+++ b/packages/core/src/components/Tooltip/Tooltip.tsx
@@ -132,6 +132,10 @@ interface TooltipBaseProps extends VibeComponentProps {
    * Sets the max width of the Tooltip, defaults to 240px
    */
   maxWidth?: number;
+  /**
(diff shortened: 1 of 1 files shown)


YossiSaadi (approved):
I think what the dev needs is that Typography would pass this on to Tooltip as well (for the ellipsis text). 
But I think that's not urgent, let's wait for tomorrow with that one.

rivka-ungar:
> I think what the dev needs is that Typography would pass this on to Tooltip as well (for the ellipsis text). But I think that's not urgent, let's wait for tomorrow with that one.

The typography doesn't have a dir prop, you mean adding it to there and then passing it to the tooltip, as supposed to adding it in the tooltip directly? what if someone wants to change it on a tooltip that is not from typography?

YossiSaadi:
Your code is needed anyway, but I think (if I recall correctly) they've added a dir to Text. So they need that if a Text is rtl, its ellipsis' Tooltip should also be rtl.
Worth asking.

PR #2667, opened 2024-12-19
Add export for package json for trident toolkit

MosheZemah:
https://monday.monday.com/boards/4400799549/pulses/7929489828

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -20,6 +20,7 @@
     "dist"
   ],
   "exports": {
+    "./package.json": "./package.json",
     ".": {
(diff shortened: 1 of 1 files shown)


sahariko (approved):

YossiSaadi (commented):
don't we have other way to access it?
it is included outside of the `dist` folder by default

sahariko:
> don't we have other way to access it? it is included outside of the `dist` folder by default

As long as `exports` exists and it's not included there, no.

talkor (approved):

PR #2668, opened 2024-12-19
feat(TransitionView): remove previous slide before continuing to next slide, enhance animation

YossiSaadi:
Motion asked to make the sliding experience a bit better, and it is great news for us as we could have removed the ugly logic of absolute and calculating height, we can now keep the slide as part of the layout flow of the browser

https://monday.monday.com/boards/3532714909/pulses/8085577206


talkor (approved):
💯

PR #2671, opened 2024-12-20
chore(TextField): Narrow types for Events on TextField

feclist:
Noticed that we can narrow the types of the event parameters on the `on..` props for the TextField. This makes it trivial to access the `value` of the input element instead of having to cast it first.

- [ ] I have read the [Contribution Guide](../CONTRIBUTING.md) for this project.


YossiSaadi (commented):
Hey @feclist, thanks for creating this PR! 

While this improves type safety and makes accessing the `value` property easier without casting, it is considered a breaking change because existing implementations using broader event types may no longer compile. Consumers will need to update their event handler signatures to comply with the new typing.

So unfortunately, for now, we cannot merge it. We will definitely add a task for our next major, though.

feclist:
Fair enough, will do a typecast for now in my code. Looking forward to the major release 😄

PR #2673, opened 2024-12-23
fix(Dropdown): Fix ESM - use .default if available

sahariko:
Same fix as the one we did for `react-windowed-select` and `react-inlinesvg`.


talkor (approved):
👑

PR #2678, opened 2024-12-26
perf: remove redundant overflow style manipulation

talkor:
https://monday.monday.com/boards/3532714909/pulses/8113296723

### DIFF
--- a/packages/core/src/hooks/useIsOverflowing/useIsOverflowing.ts
+++ b/packages/core/src/hooks/useIsOverflowing/useIsOverflowing.ts
@@ -5,13 +5,9 @@ function checkOverflow(element: HTMLElement, ignoreHeightOverflow: boolean, tole
   if (!element) {
     return false;
   }
-  const curOverflow = element.style.overflow;
-  if (!curOverflow || curOverflow === "visible") element.style.overflow = "hidden";
-  const isOverflowing =
-    element.clientWidth < element.scrollWidth ||
-    (!ignoreHeightOverflow && element.clientHeight + tolerance < element.scrollHeight);
-  element.style.overflow = curOverflow;
-  return isOverflowing;
+  const isOverflowingWidth = element.clientWidth < element.scrollWidth;
+  const isOverflowingHeight = !ignoreHeightOverflow && element.clientHeight + tolerance < element.scrollHeight;
+  return isOverflowingWidth || isOverflowingHeight;
 }
 
 export default function useIsOverflowing({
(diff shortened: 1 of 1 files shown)


packages/core/src/hooks/useIsOverflowing/useIsOverflowing.ts
```diff
@@ -5,13 +5,9 @@ function checkOverflow(element: HTMLElement, ignoreHeightOverflow: boolean, tole
   if (!element) {
     return false;
   }
-  const curOverflow = element.style.overflow;
-  if (!curOverflow || curOverflow === "visible") element.style.overflow = "hidden";
```
talkor, line 9: This probably intended for more accurate calculation but it triggers a redraw which makes this function heavy

YossiSaadi: when you say more accurate, what are we "losing" here? only legacy browsers?

talkor: In modern browsers, elements can have `overflow` styles like `visible`, `hidden`, `auto`, or `scroll`, which control how content is displayed when it exceeds the container's dimensions. Historically, and in some edge cases even today, browsers have had quirks in how they calculate `scrollWidth` and `scrollHeight` when `overflow` is set to `visible:`

1. **`overflow: visible` Behavior**:
    - When `overflow` is `visible`, the browser does not clip or restrict content, so `scrollWidth` and `scrollHeight` might not accurately represent the "overflowing" content dimensions.
    - Some older browsers would inconsistently calculate the `scrollHeight` or `scrollWidth`, leading to unreliable results. For example:
        - If the content extended outside the container, `scrollHeight` might only reflect the visible portion of the element.
2. **Dynamic Overflow Calculations**:
    - Some browsers historically recalculated layout properties like `scrollHeight` and `scrollWidth` lazily or only when `overflow` was set to `hidden`, `auto`, or `scroll`. As a result, developers often had to temporarily set `overflow` to force the browser to update these values.
3. **Rendering Engine Differences**:
    - Older rendering engines (e.g., Trident in Internet Explorer or older versions of WebKit) had subtle bugs in handling overflow. Developers would use workarounds like temporarily changing the `overflow` property to get consistent values.

The approach of temporarily setting `overflow: hidden` is a deliberate attempt to work around these historical inconsistencies. Now, it assumes that these historical quirks do not apply or that the `overflow` property is already managed correctly. In modern browsers, the need for such workarounds has significantly decreased. The major benefit is dramatic performance improvementsץ

YossiSaadi (approved):

PR #2679, opened 2024-12-26
Support event in tipseen onClose props

meromco:
Support event in tipseen onClose props

### DIFF
--- a/src/components/Tipseen/Tipseen.tsx
+++ b/src/components/Tipseen/Tipseen.tsx
@@ -55,7 +55,7 @@ interface TipseenProps extends VibeComponentProps {
    */
   isCloseButtonOnImage?: boolean;
   closeButtonOnImage?: boolean;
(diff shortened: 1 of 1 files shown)


src/components/Tipseen/Tipseen.tsx
```diff
@@ -55,7 +55,7 @@ interface TipseenProps extends VibeComponentProps {
    */
   isCloseButtonOnImage?: boolean;
   closeButtonOnImage?: boolean;
-  onClose?: () => void;
+  onClose?: (event: React.MouseEvent) => void;
```
YossiSaadi, line 58: @meromco can you make it more specific to `React.MouseEvent<HTMLButtonElement>`?
also, please make the `event` optional, as otherwise it would be a breaking change

```suggestion
  onClose?: (event?: React.MouseEvent<HTMLButtonElement>) => void;
```

YossiSaadi:
Updated in a new PR: #2680

PR #2680, opened 2024-12-26
feat(Tipseen): add event in Tipseen onClose prop

meromco:
Support event in tipseen onClose props


YossiSaadi (approved):
Thanks @meromco!

PR #2682, opened 2024-12-26
add data-testid to tipseen close button

meromco:
add data-testid to tipseen close button

### DIFF
--- a/packages/core/src/components/Tipseen/Tipseen.tsx
+++ b/packages/core/src/components/Tipseen/Tipseen.tsx
@@ -152,6 +152,7 @@ const Tipseen: VibeComponent<TipseenProps> & {
                 [styles.dark]: closeButtonTheme === "dark" || closeButtonTheme === "fixed-dark"
               })}
               onClick={onClose}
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Tipseen/Tipseen.tsx
```diff
@@ -152,6 +152,7 @@ const Tipseen: VibeComponent<TipseenProps> & {
                 [styles.dark]: closeButtonTheme === "dark" || closeButtonTheme === "fixed-dark"
               })}
               onClick={onClose}
+              data-testid={getTestId(ComponentDefaultTestId.TIPSEEN_CLOSE_BUTTON)}
```
talkor, line 155: We usually add data-testid for the root elements of components and try to avoid having many data-testids on a lot of elements. We encourage using aria/role attributes as it's the common best practice, in this example you can use the aria-label: `[data-testid="tipseen"] [aria-label="Close"]`

talkor (changes requested):

PR #2683, opened 2024-12-26
feat: support multiline for EditableText

gabriel-amram:
This PR introduces the ability to support multiline in EditableText
if multiline is passed it will render a "textarea" instead of an input that cannot be resized by the user, and instead will grow/shrink according to its content up to a certain max-width or max-height defined by the container.

When using Shift+Enter a new line will be added. Hitting Enter will defocus and change the value

This is useful for users who might want to have a long description and would like to be able to edit it and change the content with line breaks.

When multiline is used, ellipsis is removed from the "view" block and from the "textarea".

- [X] I have read the [Contribution Guide](../CONTRIBUTING.md) for this project.


packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -201,8 +273,16 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+          ellipsis={!multiline}
         >
-          {inputValue || placeholder}
+          {(multiline
+            ? inputValue.split("\n").map((line, index) => (
+                <React.Fragment key={index}>
+                  {line}
+                  {index < inputValue.split("\n").length - 1 && <br />}
+                </React.Fragment>
+              ))
```
talkor, line 284: Why do we need this? Textareas know how to break lines

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +146,56 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+
+        if (multiline) {
+          resizeTextarea();
+        }
+      }
+    }
+
+    /* Dynamically resizes the textarea to fit its content */
+    function resizeTextarea() {
+      console.log("*************************resize called, why?", multiline, inputRef.current);
```
talkor, line 171: Please remove :)

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -188,7 +245,22 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+    }
+
+    return (
+      <div
+        ref={mergedRef}
+        id={id}
+        aria-label={ariaLabel}
+        data-testid={dataTestId}
+        className={cx(styles.editableTypography, className)}
+        role={isEditing ? null : "button"}
```
talkor, line 259: It will not be shown when not editing, plus role null is bad for a11y

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +146,56 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+        setInputHeight("auto");
+
+        requestAnimationFrame(() => {
+          const textarea = inputRef.current as HTMLTextAreaElement;
+
+          if (!textarea) {
+            return;
+          }
+
+          const computedStyle = window.getComputedStyle(textarea);
```
talkor, line 183: `getComputedStyle` is pretty rough for performance, and this will be called on each render

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +146,56 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+        }
+      }
+    }
+
+    /* Dynamically resizes the textarea to fit its content */
+    function resizeTextarea() {
+      console.log("*************************resize called, why?", multiline, inputRef.current);
+      if (inputRef.current) {
+        // Temporarily set the height to "auto" to accurately measure the scroll height of the content inside the textarea.
+        setInputHeight("auto");
```
talkor, line 174: This might cause layout thrashing

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +146,56 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
         inputRef.current?.focus();
+
+        if (multiline) {
+          resizeTextarea();
+        }
+      }
+    }
+
+    /* Dynamically resizes the textarea to fit its content */
+    function resizeTextarea() {
```
talkor, line 170: We already do something similar with the width, will it work for the height as well?

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +146,56 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+          // Calculate the appropriate height by taking into account the scrollable content inside the textarea,
+          // as well as the styles applied to it, such as padding and border widths.
+          const lineHeight = parseFloat(computedStyle.lineHeight) || 16;
+          const paddingTop = parseFloat(computedStyle.paddingTop) || 0;
+          const paddingBottom = parseFloat(computedStyle.paddingBottom) || 0;
+          const borderTopWidth = parseFloat(computedStyle.borderTopWidth) || 0;
+          const borderBottomWidth = parseFloat(computedStyle.borderBottomWidth) || 0;
+
+          const newHeight = textarea.scrollHeight + borderTopWidth + borderBottomWidth;
+          const minHeight = lineHeight + paddingTop + paddingBottom + borderTopWidth + borderBottomWidth;
```
talkor, line 194: This feels really fragile, the line height can change and it will break the logic it, the calculation relies on spacings and borders which can also change
There are better stable alternative ways to adjust the height :)

packages/core/src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -26,6 +26,12 @@
     }
   }
 
+  .textarea {
+    max-height: inherit;
```
talkor, line 30: Where is it defined?

packages/core/src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -26,6 +26,12 @@
...
+    resize: none;
+    box-sizing: border-box;
```
talkor, line 32: We already use border-box for all elements

packages/core/src/components/EditableText/__stories__/EditableText.stories.tsx
```diff
@@ -102,3 +102,25 @@ export const Types = {
     }
   }
 };
+
+const multilineValue = "This is a multiline\nhere's the second line";
```
talkor, line 106: This will not be shown in the story code, let's put it inline

packages/core/src/components/EditableText/__stories__/EditableText.stories.tsx
```diff
@@ -102,3 +102,25 @@ export const Types = {
...
+export const Multiline = {
+  render: () => (
+    <div className={styles.multilineContainer}>
+      <EditableText
+        type={EditableText.types.TEXT1}
+        weight={EditableText.weights.NORMAL}
+        multiline
+        value={multilineValue}
+        className={styles.editableText}
```
talkor, line 115: Let's avoid using css in stories

packages/core/src/components/EditableText/__tests__/EditableText.test.tsx
```diff
@@ -18,6 +18,17 @@ describe("EditableText", () => {
...
 
+  it("should render a textarea in edit mode with multiline", () => {
+    render(<EditableText multiline value="Editable text" />);
+
+    const component = screen.getByRole("button");
+    fireEvent.click(component);
+
+    const input = screen.queryByRole("input");
+    expect(input).toBeInTheDocument();
+    expect(input.tagName).toBe("TEXTAREA");
```
talkor, line 29: Feels redundant IMO, better just to check that the role is `textbox`  (not `input`)

talkor (changes requested):
Thank you for the contribution! Great addition
Please see my comments :)

packages/core/src/components/EditableText/__stories__/EditableText.stories.module.scss
```diff
@@ -14,3 +14,13 @@
...
   }
 }
+
+.multilineContainer {
+  display: flex;
+  flex-direction: row;
+
+  .editableText {
+    width: 400px;
+    max-height: 200px;
```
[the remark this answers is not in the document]
gabriel-amram: just to show the max-height capabilities that it can stop growing, i can remove this if its not necessary for the story...

packages/core/src/components/EditableText/__stories__/EditableText.stories.tsx
gabriel-amram: i am using here the same CSS class other stories here are using

packages/core/src/components/EditableTypography/EditableTypography.module.scss
gabriel-amram: it is to allow the user to define a max-height that we can respect
if there is none, it will let it grow as much as needed

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -43,6 +43,8 @@ export interface EditableTypographyProps extends VibeComponentProps, EditableTyp
   type?: TextType | HeadingType;
   /** Sets the Text/Heading weight */
   weight?: TextWeight | HeadingWeight;
+  /** Controls whether a textarea or a simple input would be rendered, allowing multi-lines */
+  multiline?: boolean;
```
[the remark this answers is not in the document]
gabriel-amram: understand, but how would we use this otherwise in the Typography that both EditableText and EditableHeading are using?

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +146,56 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
     }
 
     const toggleKeyboardEditMode = useKeyboardButtonPressedFunc(toggleEditMode);
 
     function focus() {
       if (inputRef.current) {
         inputRef.current?.focus();
+
+        if (multiline) {
+          resizeTextarea();
```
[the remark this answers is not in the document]
gabriel-amram: it means the textarea just gained focus and we need to calculate its height

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: the width is dependent on the "div" behind the scenes that is growing in width, it is not growing in height as much as i wanted it to :)

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: couldn't find any, but lets discuss

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -160,22 +209,30 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
-        id={id}
-        aria-label={ariaLabel}
-        data-testid={dataTestId}
-        className={cx(styles.editableTypography, className)}
-        role={isEditing ? null : "button"}
-        onClick={onTypographyClick}
-        onKeyDown={toggleKeyboardEditMode}
-      >
-        {isEditing && (
+    function getEditableElement() {
```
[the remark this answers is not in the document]
gabriel-amram: yes

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -160,22 +209,30 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+            ref={inputRef}
+            className={cx(styles.textarea, typographyClassName)}
+            value={inputValue}
+            onChange={handleChange}
+            onKeyDown={handleKeyDown}
+            onBlur={handleBlur}
+            aria-label={ariaLabel}
+            placeholder={placeholder}
+            style={{ width: inputWidth, height: inputHeight }}
+            role="input"
```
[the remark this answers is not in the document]
gabriel-amram: changed

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -188,7 +245,22 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
             style={{ width: inputWidth }}
             role="input"
           />
-        )}
+        );
+      }
+    }
+
+    return (
+      <div
```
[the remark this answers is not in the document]
gabriel-amram: this div existed, it is not new, so shouldn't break anything

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: that part existed already, didn't want to change as i don't know what it breaks, but lets discuss and i'll change accordingly

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: this is not a textarea, that is the TypographyComponent (Text) that is shown once the textarea is not in edit mode any more

packages/core/src/components/EditableTypography/EditableTypography.module.scss
talkor: If it's a feature it should be via a prop, not by className

packages/core/src/components/EditableTypography/EditableTypography.tsx
talkor: why not just set ellipsis to false?

packages/core/src/components/EditableTypography/EditableTypography.tsx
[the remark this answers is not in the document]
talkor: Sorry my bad, the diff was confusing :)

packages/core/src/components/EditableTypography/EditableTypography.tsx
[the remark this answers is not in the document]
talkor: This is bad for performance

packages/core/src/components/EditableTypography/EditableTypography.module.scss
gabriel-amram: you're right, or instead user can commandeer the container and limit its size

packages/core/src/components/EditableTypography/EditableTypography.tsx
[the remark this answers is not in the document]
gabriel-amram: i think i found another way, i'll test it and we'll talk about it in our meeting

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -160,22 +209,30 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+            className={cx(styles.textarea, typographyClassName)}
+            value={inputValue}
+            onChange={handleChange}
+            onKeyDown={handleKeyDown}
+            onBlur={handleBlur}
+            aria-label={ariaLabel}
+            placeholder={placeholder}
+            style={{ width: inputWidth, height: inputHeight }}
+            role="input"
+            rows={1}
```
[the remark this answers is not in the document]
gabriel-amram: no, its ok and lines break. what this does is make the textarea start like a normal input (1 line), as browser defaults is 2 lines

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: ellipsis is for wrapping
we're trying to have "new lines" supported
a div needs <br/> to break into a new line

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: i am requesting an animation frame to decrease the chances of this to happen

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +148,47 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+
+          if (!textarea) {
+            return;
+          }
+
+          // Ensure we at least have 1 line
+          setInputHeight(
+            Math.max(
+              textarea.scrollHeight + textareaBorderBoxSizing.current,
+              textareaLineHeight.current + textareaBorderBoxSizing.current
```
gabriel-amram, line 188: @talkor we'll talk about it in our session, but i refactored this a bit - 
only when focusing i am calculating the getComputedStyle and "cache" it in two refs that i need to use
that way i don't need to call if on every render

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: by the way it is with ellipsis false

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -138,15 +148,47 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+      }
+    }
+
+    /* Dynamically resizes the textarea to fit its content */
+    function resizeTextarea() {
+      if (inputRef.current) {
+        // Temporarily set the height to "auto" to accurately measure the scroll height of the content inside the textarea.
+        setInputHeight("auto");
+
+        requestAnimationFrame(() => {
```
gabriel-amram, line 177: check layoutEffect instead?

gabriel-amram: layout effect does the same thing, and even adds unnecessary calculations on the lineHeight
dropped it
@talkor

packages/core/src/components/EditableTypography/EditableTypography.tsx
talkor: You can use only `white-space: pre-wrap` and it will behave the same as in editing mode and avoid splitting the string with `<br />`s

packages/core/src/components/EditableTypography/EditableTypography.tsx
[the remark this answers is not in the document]
talkor: Nevermind my mistake, it won't be shown in the EditableHeading so all good

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: @talkor works like a charm
changed it

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -141,14 +151,46 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
+
+          if (!textarea) {
+            return;
+          }
+
+          // Ensure we at least have 1 line
+          setInputHeight(
+            Math.max(
+              textarea.scrollHeight + textareaBorderBoxSizing.current,
+              textareaLineHeight.current + textareaBorderBoxSizing.current
```
talkor, line 189: Just one thought, looks like the only difference in the dimensions is 2px, and there's no difference in paddings/line heights etc between the different variants, so looks like doing just:
`setInputHeight(textarea.scrollHeight + 2);` works exactly the same and it can save the getComputeStyle and the other calculations in `calculateTextareaHeightAttrs` (which already have some default static numbers). Giving it `auto` in the beginning will already make it take the the currect box size. WDYT?

talkor:
Looks great @gabriel-amram ! Only one thought which I commented, and I did some quick QA and found two things worth checking:
- When focused on the textarea and clicking outside in the same block it will not blur
- When going into edit mode the cursor will be in the beginning of the text instead of the last character which I think makes the experience better

packages/core/src/components/EditableTypography/EditableTypography.tsx
gabriel-amram: actually the question gets back to you :)
technically its only 2px because the border-top/bottom-width is 1px (or 2px overall)
border-box should have included the border as well as the padding, but it doesn't

do we want to just rely on always having this border/padding setting and that the user cannot add anything on their own using overwriting css?
@talkor

talkor: Well it is border-box, it's just that `scrollHeight` is based purely on the content and padding of the element's scrollable content, without the borders. We don't want users to override CSS as it can make design inconsistent so I wouldn't take it into consideration, generally speaking we support the variants we officially provide so I think it's ok :)

talkor:
> * When focused on the textarea and clicking outside in the same block it will not blur

Screenshot 2024-12-31 at 11 49 50
I mean when clicking on the purple area. I think it's because the 400px width, probably changing it to `max-width` instead will fix it

gabriel-amram:
> Looks great @gabriel-amram ! Only one thought which I commented, and I did some quick QA and found two things worth checking:
> 
> * When focused on the textarea and clicking outside in the same block it will not blur
> * When going into edit mode the cursor will be in the beginning of the text instead of the last character which I think makes the experience better

@talkor 
the first one, is just a story thing, i removed the wrapping div so its clearer and it works
the second one, this is the default behaviour of browsers (input on focus goes to end of text, textarea to start). i added a piece of code on focus that makes sure that we always go to end of text for both
(this might feel broken though if people are more used to one rather than the other)

talkor (approved):
Thank you!
IMO worth simplifying the height calculation nevertheless

PR #2685, opened 2024-12-29
fix(Button): ensure icons className is not added when children are non-renderable

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8124747909


rivka-ungar (approved):
Nice tests!

PR #2689, opened 2024-12-30
fix(Dropdown): fix a11y props when searchable is false

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8085299898

### DIFF
--- a/packages/core/src/components/Dropdown/Dropdown.tsx
+++ b/packages/core/src/components/Dropdown/Dropdown.tsx
@@ -192,7 +192,8 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
         insideOverflowContainer,
         controlRef,
         insideOverflowWithTransformContainer,
-        withGroupDivider
+        withGroupDivider,
+        searchable
       });
 
       type BaseStyles = typeof baseStyles;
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -432,7 +434,7 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
         placeholder={placeholder}
         isDisabled={disabled}
         isClearable={!readOnly && clearable}
-        isSearchable={!readOnly && searchable}
+        isSearchable={!readOnly}
```
YossiSaadi, line 437: why is it removed? intentional?

YossiSaadi (approved):
Nice! had a question just to make sure

packages/core/src/components/Dropdown/Dropdown.tsx
rivka-ungar: Yep, I stopped sending the searchable prop to react-select because when it's set, the library creates a dummy input that is missing the proper aria labels. Instead, I handle the behavior directly in the Input component to make sure the accessibility labels are correct.

shaharzil (approved):

PR #2700, opened 2025-01-05
fix(tooltip): fix text with line breaks

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8138323310

Following this [pr](https://github.com/mondaycom/vibe/pull/2383), since the containerSelector is no longer `tooltips-container` which provided the `white-space: pre-wrap;` style, we need to add it directly to the component.


YossiSaadi (approved):
first of all - 👑 
second - can you make sure it is not a breaking one please? 🙏🏼

PR #2701, opened 2025-01-05
fix(EditableTypography): improve performance

talkor:
https://monday.monday.com/boards/3532714909/pulses/8162783724
https://monday.monday.com/boards/3532714909/pulses/8162188621

### DIFF
--- a/packages/core/src/components/EditableTypography/EditableTypography.tsx
+++ b/packages/core/src/components/EditableTypography/EditableTypography.tsx
@@ -1,4 +1,4 @@
-import React, { ElementType, forwardRef, useEffect, useRef, useState, useLayoutEffect } from "react";
+import React, { ElementType, forwardRef, useEffect, useRef, useState } from "react";
 import cx from "classnames";
 import useMergeRef from "../../hooks/useMergeRef";
 import VibeComponentProps from "../../types/VibeComponentProps";
@@ -10,6 +10,7 @@ import { TooltipProps } from "../Tooltip/Tooltip";
 import usePrevious from "../../hooks/usePrevious";
 import { TextType, TextWeight } from "../Text/Text.types";
 import { HeadingType, HeadingWeight } from "../Heading/Heading.types";
+import useIsomorphicLayoutEffect from "../../hooks/ssr/useIsomorphicLayoutEffect";
 
 export interface EditableTypographyImplementationProps {
   /** Value of the text */
@@ -212,7 +213,7 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> = 
       }
     }, [autoSelectTextOnEditMode, isEditing]);
 
-    useLayoutEffect(() => {
+    useIsomorphicLayoutEffect(() => {
       if (!typographyRef.current) {
         return;
       }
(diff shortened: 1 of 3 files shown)


packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -215,13 +212,13 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
       }
     }, [autoSelectTextOnEditMode, isEditing]);
 
-    useEffect(() => {
+    useLayoutEffect(() => {
```
YossiSaadi, line 215: `useIsomorphicEffect`

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -215,13 +212,13 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
       if (!typographyRef.current) {
         return;
       }
 
       const { width } = typographyRef.current.getBoundingClientRect();
-      setInputWidth(width);
+      inputRef?.current?.style.setProperty("--input-width", `${width}px`);
```
YossiSaadi, line 221: it doesn't trigger re-render, is it cool?

YossiSaadi: or it's actually the fix itself - the fact we're not re-rendering?

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -193,12 +191,11 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
...
-          setInputHeight(
-            Math.max(
-              textarea.scrollHeight + textareaBorderBoxSizing.current,
-              textareaLineHeight.current + textareaBorderBoxSizing.current
-            )
+          const height = Math.max(
+            textarea.scrollHeight + textareaBorderBoxSizing.current,
+            textareaLineHeight.current + textareaBorderBoxSizing.current
           );
+          inputRef.current?.style.setProperty("--input-height", `${height}px`);
```
YossiSaadi, line 198: same

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -183,7 +181,7 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
     function resizeTextarea() {
       if (inputRef.current) {
         // Temporarily set the height to "auto" to accurately measure the scroll height of the content inside the textarea.
-        setInputHeight("auto");
+        inputRef.current?.style.setProperty("--input-height", "auto");
```
YossiSaadi, line 184: same

packages/core/src/components/EditableTypography/EditableTypography.tsx
talkor: yep, we don't need a re-render but a recalc/style, so now it doesn't interfer with react's renders when the state changes

packages/core/src/components/EditableTypography/EditableTypography.tsx
talkor: Good point, changing

YossiSaadi (approved):

PR #2702, opened 2025-01-05
feat(Modal): allow accepting custom arias, allow passing ReactNode to ModalHeader's title

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8129253861
https://monday.monday.com/boards/3532714909/pulses/8129247306

### DIFF
--- a/packages/core/src/components/Modal/ModalHeader/ModalHeader.types.ts
+++ b/packages/core/src/components/Modal/ModalHeader/ModalHeader.types.ts
@@ -28,6 +28,6 @@ export type ModalHeaderProps = {
    * Main heading text of the modal.
    * When supplied, would also add an aria-labelledby attribute to the modal dialog element.
    */
-  title: string;
+  title: string | React.ReactNode;
 } & (WithDescription | WithoutDescription) &
   VibeComponentProps;
--- a/packages/core/src/components/Modal/ModalHeader/ModalHeader.types.ts
+++ b/packages/core/src/components/Modal/ModalHeader/ModalHeader.types.ts
@@ -9,7 +9,12 @@ interface WithoutDescription {
 interface WithDescription {
   /**
    * Descriptive text or content below the title.
-   * When supplied, would also add an aria-describedby attribute to the modal dialog element.
+   * - If you pass a **string**, this will automatically set an internally generated `aria-describedby` on the parent Modal.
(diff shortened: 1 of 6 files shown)


packages/core/src/components/Modal/ModalHeader/ModalHeader.types.ts
```diff
@@ -9,7 +9,12 @@ interface WithoutDescription {
 interface WithDescription {
   /**
    * Descriptive text or content below the title.
-   * When supplied, would also add an aria-describedby attribute to the modal dialog element.
+   * - If you pass a **string**, this will automatically set an internally generated `aria-describedby` on the parent Modal.
```
talkor, line 12: Nice, very detailed explanation

packages/core/src/components/Modal/ModalHeader/ModalHeader.types.ts
```diff
@@ -26,8 +31,16 @@ interface WithDescription {
...
+   *   so that `aria-labelledby` is set automatically (unless `Modal` receives `aria-labelledby` prop).
+   * - If you pass a **ReactNode** (such as a custom component), **you must**:
+   *   1. Assign an **`id`** to that element (or a nested element), and
+   *   2. Pass that **same `id`** as the `aria-labelledby` prop to the `Modal`.
+   *
+   * This ensures that assistive technologies know which element is the modal's title.
+   * @see [WAI-ARIA Authoring Practices for Dialog (Modal)](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#wai-ariaroles,states,andproperties)
    */
-  title: string;
+  title: string | React.ReactNode;
```
talkor, line 44: ElementContent? Or we don't want it because it's less descriptive?

talkor: Also, can it be an array?

talkor (approved):

packages/core/src/components/Modal/ModalHeader/ModalHeader.types.ts
YossiSaadi: It's a good point, let's see if a request arises, and we can add it later.

I believe most of the use cases would be wrapped under a single `div`, to flex it and position it properly, and in worst case people can just use fragment.

I generally don't like the `ElementContent` rather than just saying `ReactNode`, I think it is a bit redundant.

PR #2703, opened 2025-01-06
feat(Modal): wrap the overlay and modal inside a container, to allow portaling with layer provider to the container

YossiSaadi:
This is done to fix LayerProvider items are wrapped behind the modal's z-index.
and it seem to be a better practice to wrap the overlay and modal.

https://monday.monday.com/boards/3532714909/pulses/8165813476

### DIFF
--- a/packages/core/src/components/Modal/Modal/Modal.tsx
+++ b/packages/core/src/components/Modal/Modal/Modal.tsx
@@ -103,9 +103,6 @@ const Modal = forwardRef(
       : modalAnimationCenterPopVariants;
 
     const zIndexStyle = zIndex ? ({ "--monday-modal-z-index": zIndex } as React.CSSProperties) : {};
-    // useEffect(() => {
-    //   containerRef.current?.style?.setProperty("--monday-modal-z-index", zIndex?.toString() || "");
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Modal/Modal/Modal.tsx
```diff
@@ -103,28 +103,28 @@ const Modal = forwardRef(
       : modalAnimationCenterPopVariants;
 
     const zIndexStyle = zIndex ? ({ "--monday-modal-z-index": zIndex } as React.CSSProperties) : {};
-    const modalStyle = { ...zIndexStyle, ...style };
+    // useEffect(() => {
+    //   containerRef.current?.style?.setProperty("--monday-modal-z-index", zIndex?.toString() || "");
+    // }, [zIndex, containerRef]);
```
talkor, line 108: Remove?

talkor (approved):

packages/core/src/components/Modal/Modal/Modal.tsx
YossiSaadi: oh 🥲

PR #2707, opened 2025-01-08
fix(Dropdown): aria labels not passed to dropdown menu

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8085300950

### DIFF
--- a/packages/core/src/components/Dropdown/components/menu/menu.jsx
+++ b/packages/core/src/components/Dropdown/components/menu/menu.jsx
@@ -10,6 +10,7 @@ const Menu = ({ children, Renderer, selectProps, dropdownMenuWrapperClassName, i
     ...props,
     innerProps: {
       ...props.innerProps,
+      onScroll: props.onScroll,
       id,
       role: "listbox",
       "aria-label": ariaLabel
@@ -30,7 +31,7 @@ const Menu = ({ children, Renderer, selectProps, dropdownMenuWrapperClassName, i
   return (
     <components.Menu
       {...props}
-      innerProps={{ ...props.innerProps, onScroll: props.onScroll }}
+      innerProps={{ ...rendererProps.innerProps }}
       className={cx(
         styles.dropdownMenuWrapper,
         {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Dropdown/components/menu/menu.jsx
```diff
@@ -30,7 +31,7 @@ const Menu = ({ children, Renderer, selectProps, dropdownMenuWrapperClassName, i
   return (
     <components.Menu
       {...props}
-      innerProps={{ ...props.innerProps, onScroll: props.onScroll }}
+      innerProps={{ ...rendererProps.innerProps }}
```
YossiSaadi, line 34: this is breaking, isn't it?

YossiSaadi: wait, who is `...props.innerProps`? it seem like this attr doesn't even exist on `props` of `menu`

rivka-ungar: Why would it be breaking? Now there is a bug that the `props.innerProps` + scroll is passed but the       `id`, `role` and `aria-label` are being ignored, so I just added them

YossiSaadi: but is `props.innerProps` even a thing? where does it come from?

YossiSaadi: and if it is a thing, and it can contain `id`, `role`, `aria-label`, it means it's breaking since it overrides this change now always override those 3.

rivka-ungar: It comes from line 12, where the additional props are set

YossiSaadi: but the `Menu` component never gets a prop called `inlineProps`

YossiSaadi: oh it's a "react-select" prop which is automatically passed, I see now
we need to make sure `inlineProps` never includes one of those three
if it does, even in edge scenarios, it is kind of breaking

YossiSaadi: react-select
https://react-select.com/props#replacing-components

YossiSaadi: Is it under our control? is it always not containing it in current scenario? or consumers can somehow pass something to our `Dropdown` causing it to pass `innerProps` with different `role`, `id`, and `aria-label`?

YossiSaadi (approved):
Didn't I approved it? 😮

rivka-ungar:
This does not entirely solve the issue. It will be resolved in the new Dropdown

PR #2712, opened 2025-01-12
docs(icons): Added @vibe/icons/meta to readme

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/8129522467

### DIFF
--- a/packages/icons/README.md
+++ b/packages/icons/README.md
@@ -39,8 +39,6 @@ The meta export provides detailed metadata for each icon, including:
 - `file`: The SVG file name.
 - `description`: A description of the icon's purpose or usage.
 - `tags`: Associated tags as a comma-separated string.
-- `category`: (Optional) Icon categories such as `Basic`, `Platform`, or `View`.
-- `ignore`: (Optional) Indicates if the icon should be ignored.
 
 This metadata can be used to programmatically work with icons, such as building custom icon pickers or managing icons in your project.
 
--- a/packages/icons/README.md
+++ b/packages/icons/README.md
@@ -31,6 +31,12 @@ import { Close } from "@vibe/icons/lazy";
 import { Close as CloseSvg } from "@vibe/icons/raw";
 ```
 
+#### Using Meta Information
+
+```javascript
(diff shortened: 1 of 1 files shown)


packages/icons/README.md
```diff
@@ -31,6 +31,12 @@ import { Close } from "@vibe/icons/lazy";
 import { Close as CloseSvg } from "@vibe/icons/raw";
 ```
 
+#### Using Meta Information
+
+```javascript
+import iconsMetaData from "@vibe/icons/meta";
+```
+
```
talkor, line 39: I think it worth mentioning what kind of metadata we provide

shaharzil: I agree :) do you thing we also need to explain for what use cases to use it?

talkor: Yeah that might be also helpful

shaharzil: Done :)

packages/icons/README.md
```diff
@@ -31,6 +31,22 @@ import { Close } from "@vibe/icons/lazy";
...
+#### Using Meta Information
+
+The meta export provides detailed metadata for each icon, including:
+
+- `name`: The name of the icon.
+- `file`: The SVG file name.
+- `description`: A description of the icon's purpose or usage.
+- `tags`: Associated tags as a comma-separated string.
+- `category`: (Optional) Icon categories such as `Basic`, `Platform`, or `View`.
+- `ignore`: (Optional) Indicates if the icon should be ignored.
```
talkor, line 43: I'd remove this as it's not relevant for the metadata but only for the catalog, no? @rivka-ungar

shaharzil: The metadata contains this info...

talkor: Yes but how will it benefit the users who use this metadata? 😅

rivka-ungar: It is used internally to filter out icons we don't want to display in the catalog, so it probably doesn't need to be in the readme

shaharzil: Sababi, removing

talkor (approved):
Cool :)

PR #2713, opened 2025-01-13
feat(MenuItem): allow observing submenu content resize when a re-render isn't triggered

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8222833258

Before:
beforeObserveMenu

After:
afterObserveMenu

### DIFF
--- a/packages/core/src/components/Menu/MenuItem/MenuItem.tsx
+++ b/packages/core/src/components/Menu/MenuItem/MenuItem.tsx
@@ -57,13 +57,11 @@ export interface MenuItemProps extends VibeComponentProps {
   "aria-label"?: AriaAttributes["aria-label"];
   submenuPosition?: SubmenuPosition;
   /**
-   * Enables the observation of content resize for the menu item's submenu.
-   * When set to `true`, a ResizeObserver is attached to the popper content,
-   * automatically triggering repositioning when the size of the submenu's content changes.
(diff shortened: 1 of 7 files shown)


packages/core/src/components/Menu/MenuItem/MenuItem.tsx
```diff
@@ -56,6 +56,14 @@ export interface MenuItemProps extends VibeComponentProps {
...
   "aria-label"?: AriaAttributes["aria-label"];
   submenuPosition?: SubmenuPosition;
+  /**
+   * Enables the observation of content resize for the menu item's submenu.
+   * When set to `true`, a ResizeObserver is attached to the popper content,
+   * automatically triggering repositioning when the size of the submenu's content changes.
+   *
+   * This is useful for when submenu's content may grow or shrink without a re-render being triggered.
+   */
+  observeSubMenuContentResize?: boolean;
```
talkor, line 66: Are the implementation details really matter here? I think we can simplify it and just say that it will respond to resizing of the content
Also worth renaming prop name IMO, maybe `autoAdjustSubMenuOnResize`?

talkor (approved):
Cool! Just one comment about naming

packages/core/src/components/Menu/MenuItem/MenuItem.tsx
talkor: Sounds good

PR #2715, opened 2025-01-13
fix(icons): update folder icon

shaharzil:
https://monday.monday.com/boards/3532715121/pulses/7932464360


YossiSaadi (commented):
is this kind of change not considered a breaking one?

PR #2727, opened 2025-01-21
chore: notice users about package deprecation

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/7960679002

### DIFF
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -1,5 +1,15 @@
 # Vibe Design System - Core
 
+# ⚠️ Deprecation Notice
+
+**monday-ui-react-core is now deprecated.**
+
+We have transitioned to the [`@vibe/core`](https://www.npmjs.com/package/@vibe/core) package, which serves as the successor to `monday-ui-react-core` starting from major version **3.x**.
+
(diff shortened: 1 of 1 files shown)


rivka-ungar (approved):

packages/core/README.md
```diff
@@ -1,5 +1,15 @@
 # Vibe Design System - Core
 
+# ⚠️ Deprecation Notice
```
talkor, line 3: @YossiSaadi It seems like you're deprecating this package as it's on the top and highlighted. Also we referred to it in the main readme also and it has a link to npm and the new package is mentioned in the getting started

YossiSaadi: @talkor What would you like to write there then?

talkor: How about deprecating the npm package instead? It will be shown when you install it which will be more useful IMO

YossiSaadi: You said it is too aggressive, that's why I didn't just do it 🙄🫢

talkor: I said "not at the moment" haha

PR #2728, opened 2025-01-21
docs(Accordion): add description to all props and enable live edit

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/8286797392


packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -5,28 +5,53 @@ import { VibeComponentProps, ElementContent } from "../../../types";
...
   iconSize?: number | string;
   /**
-   * On click callback
+   * Custom callback triggered when the item is clicked.
    */
   onClick?: () => void;
+  /** @ignore */
   open?: boolean;
+  /** @ignore */
   onClickAccordionCallback?: () => void;
```
YossiSaadi, line 26: why?

packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -5,28 +5,53 @@ import { VibeComponentProps, ElementContent } from "../../../types";
...
+  /**
+   * Custom class name to add to the header of the expandable
+   */
   headerClassName?: string;
+  /**
+   * Custom class name to add to the content of the expandable
+   */
   contentClassName?: string;
+  /**
+   * Custom class name to add to the ExpandCollapse component
```
YossiSaadi, line 40: maybe it should be less specific, probably not all devs that are using this component know it is internally with ExpandCollapse

packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -5,28 +5,53 @@ import { VibeComponentProps, ElementContent } from "../../../types";
...
+   */
   expandCollapseComponentClassName?: string;
+  /**
+   * Determines whether click events are captured.
```
YossiSaadi, line 44: maybe we should expand a little more on that, saying capture phase or expand it even more, I was surprised to know not all devs know what capture is

packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
```diff
@@ -5,28 +5,53 @@ import { VibeComponentProps, ElementContent } from "../../../types";
...
+   */
   captureOnClick?: boolean;
+  /**
+   * A classname to be added to the accordion item container
```
YossiSaadi, line 48: you wrote in the previous sections class name, let's keep it same here

packages/core/src/components/Accordion/Accordion/Accordion.tsx
```diff
@@ -33,6 +33,14 @@ export interface AccordionProps extends VibeComponentProps {
    * Array of initial expanded indexes
    */
   defaultIndex?: Array<number>;
+  /**
+   * A classname to be added to the accordion container
```
YossiSaadi, line 37: same for the class name

YossiSaadi (commented):
looks cool, had a few comments

packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
shaharzil: These are internal props and we don't want our users to use them. In the past it even caused some questions in ask-vibe I believe it will be best just to hide it :)

packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
shaharzil: Not sure I understand, the prop name is already super specific(expandCollapseComponentClassName), so why shouldn't the description be?

packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
YossiSaadi: got ya 💯

packages/core/src/components/Accordion/AccordionItem/AccordionItem.tsx
YossiSaadi: hmm there's something to it
let's leave it like that then

talkor (approved):

PR #2729, opened 2025-01-21
feat(Modal): allow disabling autoFocus

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8291737832
https://monday.monday.com/boards/3532714909/pulses/8188699942


rivka-ungar (dismissed):

rivka-ungar (approved):
Looks cool, and great test!
Just wondering if in case of a modal on top of a modal can this cause any issues?

YossiSaadi:
> Details

Good point, just tried it in the playground and it seem the lock mechanism keep them separated correctly, same as it was 2 modals that none manage the focus on their own

PR #2731, opened 2025-01-26
feat(Icons): add ClassicFolder icon

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/7928875316


talkor (approved):
Classic

PR #2733, opened 2025-01-26
chore: remove built icons from ignore

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8325999031


package-lock.json
talkor, line 1: Why has it changed?

packages/icons/.gitignore
```diff
@@ -1,3 +1,2 @@
-src/react
 src/lazy
```
talkor, line 1: Should also be removed, no? If it's used in core it will do the same

talkor (approved):

talkor:
@rivka-ungar Can also be a chore PR, release is not needed, right?

package-lock.json
rivka-ungar: mistake

packages/icons/.gitignore
rivka-ungar: The lazy icons? they are not used in core

PR #2734, opened 2025-01-26
feat(Slider): be able to change value label position and color

talkor:
https://monday.monday.com/boards/3532714909/pulses/8223729904

### DIFF
--- a/packages/core/src/components/Slider/Slider.tsx
+++ b/packages/core/src/components/Slider/Slider.tsx
@@ -77,6 +77,14 @@ export type SliderProps = {
    */
   showValue?: boolean;
   /**
+   * Position of the `value` when `showValue` is true
+   */
+  valueLabelPosition?: "top" | "bottom";
+  /**
+   * Color of the `value` when `showValue` is true
+   */
+  valueLabelColor?: "primary" | "secondary";
+  /**
    * Size small/medium/large of the component (Slider)
    */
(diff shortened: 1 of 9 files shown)


packages/core/src/components/Slider/Slider.tsx
```diff
@@ -76,6 +76,14 @@ export type SliderProps = {
...
    */
   showValue?: boolean;
+  /**
+   * Position of the `value` when `showValue` is true
+   */
+  valueLabelPosition?: "top" | "bottom";
+  /**
+   * Color of the `value` when `showValue` is true
+   */
+  valueLabelColor?: "primary" | "secondary";
```
rivka-ungar, line 86: Can these values be added as a type in the Slider.types.ts file same as `SliderColor`?

packages/core/src/components/Slider/Slider.tsx
```diff
@@ -76,6 +76,14 @@ export type SliderProps = {
    * Always show `value` instead of Tooltip
    */
   showValue?: boolean;
+  /**
+   * Position of the `value` when `showValue` is true
+   */
+  valueLabelPosition?: "top" | "bottom";
```
rivka-ungar, line 82: Can these also be added in the types file instead of inline?

rivka-ungar (approved):

PR #2737, opened 2025-01-26
feat(Modal): add responsiveness for modal

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8177017359


packages/core/src/components/Modal/Modal/Modal.module.scss
```diff
@@ -39,17 +39,17 @@ $full-view-margin: 40px;
 
   &.sizeSmall {
     --modal-max-height: 50%;
-    --modal-width: 480px;
+    --modal-width: 460px;
```
rivka-ungar, line 42: Is this not a breaking change?

rivka-ungar (approved):

packages/core/src/components/Modal/Modal/Modal.module.scss
YossiSaadi: kind of, but we took a decision that we need it

PR #2742, opened 2025-01-29
feat(Search): add prop for hiding clear icon

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8190214567

### DIFF
--- a/packages/core/src/components/Search/Search.types.ts
+++ b/packages/core/src/components/Search/Search.types.ts
@@ -100,5 +100,5 @@ export interface SearchProps extends VibeComponentProps {
   /**
    * Controls the visibility of the clear icon.
    */
-  hasClearIcon?: boolean;
+  showClearIcon?: boolean;
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Search/Search.types.ts
```diff
@@ -97,4 +97,8 @@ export interface SearchProps extends VibeComponentProps {
    * Callback function that is called when a keyboard key is down.
    */
   onKeyDown?: (event: React.KeyboardEvent<HTMLInputElement>) => void;
+  /**
+   * Controls the visibility of the clear icon.
+   */
+  hasClearIcon?: boolean;
```
talkor, line 103: I think we use mostly the `show*` convention in external props such as this, no? Like `showClearIcon`
WDYT?

rivka-ungar: I changed it

talkor (approved):

PR #2745, opened 2025-01-30
fix: change fixed-dark-color to be fixed on all themes

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8350787297


YossiSaadi (commented):
It is kind of a breaking change 🫣

rivka-ungar:
> It is kind of a breaking change 🫣

Yep, thats what the design want. It is apparently a bug that it is called fixed-dark-color and it changes between the themes instead of being fixed.

YossiSaadi (approved):

PR #2749, opened 2025-02-03
feat(Dropdown): allow ellipsis for component's placeholder

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8326192363

### DIFF
--- a/packages/core/src/components/Dropdown/Dropdown.styles.js
+++ b/packages/core/src/components/Dropdown/Dropdown.styles.js
@@ -178,14 +178,14 @@ const control =
   };
 
 const placeholder =
-  ({ forcePlaceholderEllipsis }) =>
+  ({ allowPlaceholderEllipsis }) =>
   (provided, { isDisabled }) => ({
     ...provided,
     ...getFont(),
     color: isDisabled ? getCSSVar("disabled-text-color") : getCSSVar("secondary-text-color"),
     fontWeight: getCSSVar("font-weight-normal"),
     // 22px because we have the inner arrow of opened/closed
-    ...(forcePlaceholderEllipsis && { width: "calc(100% - 22px)" })
+    ...(allowPlaceholderEllipsis && { width: "calc(100% - 22px)" })
   });
 
 const indicatorsContainer =
--- a/packages/core/src/components/Dropdown/Dropdown.styles.js
+++ b/packages/core/src/components/Dropdown/Dropdown.styles.js
@@ -178,12 +178,14 @@ const control =
   };
 
 const placeholder =
-  () =>
+  ({ forcePlaceholderEllipsis }) =>
   (provided, { isDisabled }) => ({
     ...provided,
     ...getFont(),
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Dropdown/Dropdown.styles.js
```diff
@@ -178,12 +178,14 @@ const control =
...
 
 const placeholder =
-  () =>
+  ({ forcePlaceholderEllipsis }) =>
   (provided, { isDisabled }) => ({
     ...provided,
     ...getFont(),
     color: isDisabled ? getCSSVar("disabled-text-color") : getCSSVar("secondary-text-color"),
-    fontWeight: getCSSVar("font-weight-normal")
+    fontWeight: getCSSVar("font-weight-normal"),
```
rivka-ungar, line 186: Isn't 22 too much? Wasn't it supposed to be 13?

packages/core/src/components/Dropdown/Dropdown.styles.js
```diff
@@ -178,12 +178,14 @@ const control =
   };
 
 const placeholder =
-  () =>
+  ({ forcePlaceholderEllipsis }) =>
```
rivka-ungar, line 181: The name sounds pretty intimidating no? Can it be something like `showPlaceholderEllipsis`?

rivka-ungar (approved):
Great!
Had 2 small questions

packages/core/src/components/Dropdown/Dropdown.styles.js
YossiSaadi: I've calculated it to be the exact width as the placeholder's container

I know it's not ideal, but it is the same width for all of the cases + it makes the ellipsis work for readOnly mode as well in a proper way without overflowing

it was suppose to be 11, but 22px seem more "right" value

packages/core/src/components/Dropdown/Dropdown.styles.js
YossiSaadi: I was wondering about the naming as well, wdyt about `allow` maybe?

packages/core/src/components/Dropdown/Dropdown.styles.js
rivka-ungar: Just wondering if in the smaller dropdowns there will be too much white space? But I agree that its better it fits all cases so that there is no case were it gets cut off

YossiSaadi: It fits the same place that the "options" has right now, under all sizes
that means that currently the placeholder overflows the space it should have (probably a react-select glitch? can't understand the reason if this is intentional. if it isn't react-select then it has something to do with our own styling), and that fixes it as well

YossiSaadi: I mean the selectedOption

YossiSaadi: Feb-03-2025 15-45-29

PR #2750, opened 2025-02-03
feat(Tab): add tooltipProps prop

tomer-gillmore:

Resolves #

### DIFF
--- a/packages/core/src/components/Tabs/Tab/Tab.tsx
+++ b/packages/core/src/components/Tabs/Tab/Tab.tsx
@@ -1,5 +1,5 @@
 import cx from "classnames";
-import React, { FC, forwardRef, ReactElement, useRef } from "react";
+import React, { FC, forwardRef, ReactElement, useMemo, useRef } from "react";
 import { noop as NOOP } from "lodash-es";
 import useMergeRef from "../../../hooks/useMergeRef";
 import { getStyle } from "../../../helpers/typesciptCssModulesHelper";
@@ -9,6 +9,7 @@ import { IconType } from "../../Icon";
 import { ComponentDefaultTestId, getTestId } from "../../../tests/test-ids-utils";
 import styles from "./Tab.module.scss";
 import { SubIcon } from "../../../types/SubIcon";
+import Tooltip, { TooltipProps } from "../../Tooltip/Tooltip";
 
 export interface TabProps extends VibeComponentProps {
   /**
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Tabs/Tab/Tab.tsx
```diff
@@ -77,25 +80,27 @@ const Tab: FC<TabProps> = forwardRef(
...
-        aria-selected={active}
-        aria-disabled={disabled}
-        data-testid={dataTestId || getTestId(ComponentDefaultTestId.TAB, id)}
-      >
-        {/* eslint-disable-next-line jsx-a11y/anchor-is-valid,jsx-a11y/click-events-have-key-events */}
-        <a className={cx(styles.tabInner, tabInnerClassName)} onClick={() => !disabled && onClick(value)}>
-          {renderIconAndChildren()}
-        </a>
-      </li>
+      <Tooltip {...tooltipProps} content={disabled && tooltipProps.content}>
```
talkor, line 83: The tooltip props/content should not be conditioned by the disabled, but should be allowed for general tooltips as well (which means that the user of the component should condition the tooltip according to the disable value)

packages/core/src/components/Tabs/Tab/__tests__/Tab.test.tsx
```diff
@@ -1,17 +1,67 @@
...
+      fireEvent.click(screen.getByText(text));
+      expect(onClickMock.mock.calls.length).toBe(1);
+    });
   });
 
-  it("should call the click callback when clicked", () => {
-    fireEvent.click(screen.getByText(text));
-    expect(onClickMock.mock.calls.length).toBe(1);
+  describe("Tooltips", () => {
+    it("should not display tooltip content from tooltipProps if disabled is false", () => {
```
talkor, line 31: This will be redundant when we change the tooltipProps condition of the disabled

packages/core/src/components/Tabs/Tab/__tests__/Tab.test.tsx
```diff
@@ -1,17 +1,67 @@
...
+      jest.advanceTimersByTime(1000);
+      const content = screen.queryByText(tooltipContent);
+      expect(content).toBeFalsy();
+      act(() => {
+        fireEvent.mouseLeave(component);
+      });
+      jest.advanceTimersByTime(1000);
+    });
+
+    it("should display tooltip content from tooltipProps if disabled is true", () => {
```
talkor, line 49: Same

talkor (approved):
Thanks @tomer-gilmore! Let's just remove the condition of the disabled from the TooltipProps

PR #2751, opened 2025-02-03
Fix/hananma/extends dropdown capabilities

hanan-monday:
- Adding const data-testid to modal close button
- Adding new functionalities to testkit


packages/core/src/components/Modal/ModalTopActions/ModalTopActions.tsx
```diff
@@ -17,6 +17,7 @@ const ModalTopActions = ({ renderAction, theme, closeButtonAriaLabel, onClose }:
     <div className={styles.actions} data-no-autofocus={true}>
       {typeof renderAction === "function" ? renderAction(buttonColor) : renderAction}
       <IconButton
+        data-testid={"close-modal-button"}
```
YossiSaadi, line 20: add in `packages/core/src/tests/constants.ts`:
```
...
MODAL_NEXT_OVERLAY = "modal-overlay",
MODAL_NEXT_HEADER = "modal-header",
--> MODAL_NEXT_CLOSE_BUTTON = "modal-close-button",
...
```

and then:
```suggestion
        data-testid={ComponentDefaultTestId.MODAL_NEXT_CLOSE_BUTTON}
```

packages/testkit/components/BaseElement.ts
YossiSaadi, line 1: Those changes are considered breaking changes, no?
the deleted stuff here were accessible for users of @vibe/testkit?

we're open source, we should consider this kind of stuff
also, it would affect current usages within monday, causing tests to break for people upgrading the @vibe/testkit

packages/testkit/components/Dropdown.ts
```diff
@@ -22,7 +22,17 @@ export class Dropdown extends BaseElement {
...
+
+  /**
+   * Open the dropdown.
+   * @returns {Promise<void>}
+   */
+  async open(): Promise<void> {
+    await test.step(`Open ${this.elementReportName}`, async () => {
+      await this.locator.click();
+    });
   }
```
YossiSaadi, line 36: nice

packages/testkit/components/Modal.ts
```diff
@@ -20,13 +21,28 @@ export class Modal {
     this.page = page;
     this.locator = locator;
     this.elementReportName = elementReportName;
-    this.closeModalButton = new IconButton(this.page, this.locator, elementReportName);
+    this.closeModalButton = new IconButton(
+      this.page,
+      this.locator.getByTestId("close-modal-button"),
```
YossiSaadi, line 26: change in here as well afterwards

YossiSaadi: though we already talked about it, I strongly suggest that @vibe/testkit would match **@vibe/core only**. if it can match other custom usages by accident (e.g. other modals, etc.), then it's nice and useful, but everything under `@vibe` namespace, especially as we're open source, should match specifically our @vibe/core library.

also, we're planning on having testkit for unit tests in the future, and it should also match specifically for @vibe/core

packages/testkit/components/TextField.ts
```diff
@@ -19,6 +20,7 @@ export class TextField extends BaseElement {
     super(page, locator, elementReportName);
     this.page = page;
     this.locator = locator;
+    this.elementReportName = elementReportName;
```
YossiSaadi, line 23: are those really needed in addition to the `super`? I think you don't need to re-declare 
```
  override page: Page;
  override locator: Locator;
  override elementReportName: string;
```

as you already have them as part of your inheritance

packages/testkit/components/TextField.ts
```diff
@@ -28,10 +30,28 @@ export class TextField extends BaseElement {
    */
   async setText(text: string): Promise<void> {
     await test.step(`Set text: ${text} in element: ${this.elementReportName}`, async () => {
+      await this.locator.fill("");
```
YossiSaadi, line 33: can you use your own `clearText`?

packages/testkit/components/BaseElement.ts
hanan-monday: I agree.
Since we currently just in the first steps, the method was in use just in the testkit components and not outside and I replaced all usages.

packages/testkit/components/TextField.ts
hanan-monday: Right. Changed it in several files

packages/testkit/components/TextField.ts
hanan-monday: Done

packages/core/src/components/Modal/ModalTopActions/ModalTopActions.tsx
hanan-monday: Done

packages/testkit/components/BaseElement.ts
YossiSaadi: perfect then 💪🏼

packages/testkit/components/Dropdown.ts
hanan-monday: @YossiSaadi - I updated this functionality. Please review again

YossiSaadi (approved):
💯

uziab (approved):

PR #2752, opened 2025-02-04
feat(useIsOverflowing): allow width tolerance

YossiSaadi:
we need it to support weird browser's behaviors, such as Safari with 125% zoom, which causes 1 pixel diff (in clientWidth vs scrollWidth), as Chromium and Safari treat rounding floating pixel differently (round up vs round down)

https://monday.monday.com/boards/3532714909/pulses/8350809092


talkor (approved):
👑

PR #2753, opened 2025-02-04
feat(Flex): introduced flex prop

alexandrzhd:

it's based on the discussion in this thread: https://monday.slack.com/archives/C034VLARPJS/p1738148927052189

### DIFF
--- a/packages/core/src/components/Flex/Flex.tsx
+++ b/packages/core/src/components/Flex/Flex.tsx
@@ -112,15 +112,19 @@ const Flex: VibeComponent<FlexProps> & {
     const flexStyle = useMemo(() => {
       if (!flex) return {};
 
-      if (["string, number"].includes(typeof flex)) {
+      if (["string", "number"].includes(typeof flex)) {
         return { flex };
       }
 
-      return {
-        flexGrow: flex.grow,
-        flexShrink: flex.shrink,
-        flexBasis: flex.basis
-      };
+      if (typeof flex === "object") {
+        return {
+          flexGrow: flex.grow,
+          flexShrink: flex.shrink,
+          flexBasis: flex.basis
+        };
+      }
+
+      return {};
     }, [flex]);
 
     const overrideStyle = useMemo(() => ({ ...style, ...gapStyle, ...flexStyle }), [style, gapStyle, flexStyle]);
--- a/packages/core/src/components/Flex/Flex.tsx
+++ b/packages/core/src/components/Flex/Flex.tsx
@@ -110,10 +110,14 @@ const Flex: VibeComponent<FlexProps> & {
     }, [gap]);
 
     const flexStyle = useMemo(() => {
-      if (!flex) {
-        return {};
+      if (!flex) return {};
+
+      if (["string, number"].includes(typeof flex)) {
+        // which means we got a valid regular css flex value:
+        return { flex };
       }
 
+      // otherwise your regular logic
       return {
         flexGrow: flex.grow,
         flexShrink: flex.shrink,
--- a/packages/core/src/components/Flex/Flex.tsx
+++ b/packages/core/src/components/Flex/Flex.tsx
@@ -8,7 +8,7 @@ import {
   FlexGap as FlexGapEnum,
   FlexJustify as FlexJustifyEnum
 } from "./FlexConstants";
-import { FlexDirection, FlexJustify, FlexAlign, FlexGap } from "./Flex.types";
+import { FlexDirection, FlexJustify, FlexAlign, FlexGap, FlexShorthand } from "./Flex.types";
 import { ElementContent, VibeComponent, VibeComponentProps, withStaticProps } from "../../types";
 import { getStyle } from "../../helpers/typesciptCssModulesHelper";
 import styles from "./Flex.module.scss";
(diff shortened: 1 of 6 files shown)


packages/core/src/components/Flex/Flex.tsx
```diff
@@ -70,7 +72,19 @@ const Flex: VibeComponent<FlexProps> & {
...
+      if (!flex) {
+        return {};
+      }
+
+      return {
+        flexGrow: flex.grow,
+        flexShrink: flex.shrink,
+        flexBasis: flex.basis
+      };
+    }, [flex]);
```
YossiSaadi, line 85: If we are to enable also supplying string/number (using `CSSProperties["flex"]`):
```suggestion
    const flexStyle = useMemo(() => {
      if (!flex) return {};

      if (["string, number"].includes(typeof flex)) {
        // which means we got a valid regular css flex value:
        return { flex };
      }

      // otherwise your regular logic
      return {
        flexGrow: flex.grow,
        flexShrink: flex.shrink,
        flexBasis: flex.basis
      };
    }, [flex]);
```

packages/core/src/components/Flex/Flex.tsx
```diff
@@ -8,7 +8,7 @@ import {
   FlexGap as FlexGapEnum,
   FlexJustify as FlexJustifyEnum
 } from "./FlexConstants";
-import { FlexDirection, FlexJustify, FlexAlign, FlexGap } from "./Flex.types";
+import { FlexDirection, FlexJustify, FlexAlign, FlexGap, FlexShorthand } from "./Flex.types";
```
YossiSaadi, line 11: I'm wondering if FlexShorthand is a good-enough name for the type (remember, this is exported to consumers, and should be readable and understandable where it belongs to from its name)

YossiSaadi: @talkor wdyt?

packages/core/src/components/Flex/Flex.tsx
YossiSaadi: cc @tducasse @talkor

YossiSaadi (commented):
Great contribution! Let's discuss my comments, let me know what you think

packages/core/src/components/Flex/Flex.tsx
talkor: I agree but I can't think of a better name 😅

packages/core/src/components/Flex/Flex.tsx
talkor: Giving the ability to use the string is really cool IMO!

talkor:
@alexandrzhd this is super cool! Looks really good, I think giving the option to use the shorthand as a string would also be useful, as @YossiSaadi mentioned

packages/core/src/components/Flex/Flex.tsx
alexandrzhd: @YossiSaadi thanks! ability to use flex as string is cool!

packages/core/src/components/Flex/Flex.tsx
alexandrzhd: The same for me, I named it FlexType first, but FlexShorthand looked more clear to me

talkor (approved):
Amazing @alexandrzhd, thanks you for this awesome addition!
Wrote one comment about examples and testing the string usage

packages/core/src/components/Flex/Flex.types.ts
```diff
@@ -1,7 +1,15 @@
...
 
 export type FlexGap = "xs" | "small" | "medium" | "large";
 
 export type FlexDirection = "row" | "column";
+
+export type FlexShorthand = {
+  grow?: CSSProperties["flexGrow"];
+  shrink?: CSSProperties["flexShrink"];
+  basis?: CSSProperties["flexBasis"];
+};
```
YossiSaadi, line 15: add the string type as well

YossiSaadi:
Hey, @talkor @alexandrzhd, do we want to push on the efforts on that one?

talkor:
> Hey, @talkor @alexandrzhd, do we want to push on the efforts on that one?

Yes please! This is a super useful, I think that adding the ability to give the string value will be even more useful if it's a quick win. @alexandrzhd WDYT?

alexandrzhd:
> > Hey, @talkor @alexandrzhd, do we want to push on the efforts on that one?
> 
> Yes please! This is a super useful, I think that adding the ability to give the string value will be even more useful if it's a quick win. @alexandrzhd WDYT?

Wow, sorry for long answer. I didn't see the comment :(

Sure, let's go on with it. I've updated branch to latest upstream. Should be fine to go.

packages/core/src/components/Flex/Flex.types.ts
talkor: @alexandrzhd we need the `| CSSProperties["flex"]` for the sharthand, no?

alexandrzhd: @talkor I think I've addressed all issues, should be good to go now

talkor (approved):
Amazing!!

PR #2754, opened 2025-02-04
feat(ModalFooter): allow tooltip for the modal footer's buttons

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8401165563

### DIFF
--- a/packages/core/src/components/Modal/footers/ModalFooterBase/ModalFooterBase.tsx
+++ b/packages/core/src/components/Modal/footers/ModalFooterBase/ModalFooterBase.tsx
@@ -4,14 +4,24 @@ import Button from "../../../Button/Button";
 import Flex from "../../../Flex/Flex";
 import { ModalFooterBaseProps } from "./ModalFooterBase.types";
 import cx from "classnames";
+import { Tooltip } from "../../../Tooltip";
 
 const ModalFooterBase = forwardRef(
   (
     { primaryButton, secondaryButton, renderAction, id, className, "data-testid": dataTestId }: ModalFooterBaseProps,
     ref: React.ForwardedRef<HTMLDivElement>
   ) => {
-    const { text: primaryButtonText, ...primaryButtonProps } = primaryButton;
-    const { text: secondaryButtonText, ...secondaryButtonProps } = secondaryButton || {};
+    const {
+      text: primaryButtonText,
+      tooltipProps: primaryButtonTooltipProps = {},
+      ...primaryButtonProps
+    } = primaryButton;
+    const {
+      text: secondaryButtonText,
+      tooltipProps: secondaryButtonTooltipProps = {},
+      ...secondaryButtonProps
+    } = secondaryButton || {};
+
     return (
       <Flex
         ref={ref}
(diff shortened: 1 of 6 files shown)


YossiSaadi:
prerelease: `@vibe/core => 3.25.0-alpha-3706c.0`

packages/core/src/components/Modal/footers/utils/getPropsForButton.ts
```diff
@@ -0,0 +1,14 @@
...
+import cx from "classnames";
+
+export function getPropsForButton(button?: ModalFooterActionProps, buttonClassName?: string) {
+  if (!button) return undefined;
+  const { tooltipProps, className, ...rest } = button;
+  return {
+    ...rest,
+    className: tooltipProps ? className : cx(className, buttonClassName),
+    tooltipProps: tooltipProps
+      ? { ...tooltipProps, referenceWrapperClassName: cx(tooltipProps.referenceWrapperClassName, buttonClassName) }
```
talkor, line 11: Why do you need the referenceWrapperClassName? Does it add you another wrapping element?

packages/core/src/components/Modal/footers/ModalFooterBase/ModalFooterBase.tsx
```diff
@@ -21,11 +31,15 @@ const ModalFooterBase = forwardRef(
         className={cx(styles.footer, className)}
         data-testid={dataTestId}
       >
-        <Button {...primaryButtonProps}>{primaryButtonText}</Button>
+        <Tooltip {...primaryButtonTooltipProps} content={primaryButtonTooltipProps.content}>
```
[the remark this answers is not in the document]
YossiSaadi: otherwise typescript yells 🫠

packages/core/src/components/Modal/footers/utils/getPropsForButton.ts
YossiSaadi: I need to add the button order class name to the first-wrapper, it can be either the button itself or the span of the Refable

YossiSaadi: to allow the first child of the flex receive the "order" css attribute

talkor: But shouldn't it be on the button element regardless?

packages/core/src/components/Modal/footers/ModalFooterBase/ModalFooterBase.tsx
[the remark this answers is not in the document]
talkor: Oh because content is mandatory?

packages/core/src/components/Modal/footers/utils/getPropsForButton.ts
YossiSaadi: if the button's are wrapped with the span, then the span is the direct child of flex, so if the buttons themselves get the "order" css, it would have no effect. the order should be applied on the direct child of the flex container

packages/core/src/components/Modal/footers/ModalFooterBase/ModalFooterBase.tsx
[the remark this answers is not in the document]
YossiSaadi: yep

PR #2758, opened 2025-02-10
feat: testkit - playwright should be a peer dependency

hanan-monday:
This is in order to avoid playwright versions conflicts.
Anyone who require testkit, should requiring it as a peerDependency


YossiSaadi (approved):

YossiSaadi (approved):
Notice that peer dep might be lower if it's not critical

PR #2765, opened 2025-02-16
feat(EditableText)--be-able-to-set-max-row-for-multiline-variant

talkor:
https://monday.monday.com/boards/3532714909/pulses/8279852367


packages/core/src/components/EditableTypography/EditableTypography.module.scss
```diff
@@ -30,7 +30,7 @@
 
   .textarea {
     resize: none;
-    overflow: hidden;
+    overflow: auto;
```
rivka-ungar, line 33: Is this not considered breaking change?

packages/core/src/components/EditableTypography/EditableTypography.tsx
```diff
@@ -248,7 +250,7 @@ const EditableTypography: VibeComponent<EditableTypographyProps, HTMLElement> =
           tooltipProps={tooltipProps}
           weight={weight}
           type={type}
-          ellipsis={!multiline}
+          maxLines={multilineMaxRows}
```
rivka-ungar, line 253: The `maxLines` replaces the `ellipsis` behavior? if it is not multiline than shouldn't the `ellipsis` stay?

rivka-ungar (approved):

packages/core/src/components/EditableTypography/EditableTypography.module.scss
talkor: It had an effect only in the multiline variant but it didn't really do anything as the text was growing and never achieving overflow. But when limiting the rows we need it to be auto for the scrollbar. Do you think it can break anyway?

packages/core/src/components/EditableTypography/EditableTypography.tsx
talkor: TBH I don't know why it was present anyway before, but ellipsis={false} disables both tooltip and and overflow and we need both

PR #2766, opened 2025-02-16
docs: add example of links on surfaces

talkor:
Screenshot 2025-02-16 at 11 49 30

https://monday.monday.com/boards/3532714909/pulses/8436115722


packages/core/src/components/Link/__stories__/Example.tsx
talkor, line 1: I was thinking about extracting it to storybook-blocks but need to see if it's a repeating pattern before adding it as a component

rivka-ungar (approved):

PR #2767, opened 2025-02-16
feat(Label): Add content colors

shaharzil:
https://monday.monday.com/boards/3532714909/pulses/8222602622

### DIFF
--- a/packages/core/src/components/Label/Label.tsx
+++ b/packages/core/src/components/Label/Label.tsx
@@ -5,8 +5,9 @@ import { getStyle } from "../../helpers/typesciptCssModulesHelper";
 import React, { forwardRef, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import Text from "../Text/Text";
 import Leg from "./Leg";
-import { LabelColor as LabelColorEnum, LabelKind as LabelKindEnum, mapSizesToTextSize } from "./LabelConstants";
+import { LabelAllowedColor as LabelColorEnum, LabelKind as LabelKindEnum, mapSizesToTextSize } from "./LabelConstants";
 import { LabelColor, LabelKind } from "./Label.types";
+import { contentColors } from "../../utils/colors-vars-map";
 import { VibeComponent, VibeComponentProps, withStaticProps } from "../../types";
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Label/Label.tsx
```diff
@@ -70,6 +71,13 @@ const Label: VibeComponent<LabelProps> & {
       [kind, color, isLegIncluded, labelClassName, isClickable, size]
     );
 
+    const backgroundColorStyle = useMemo(() => {
+      type ContentColor = (typeof contentColors)[number];
```
talkor, line 75: It's already defined in the types file

packages/core/src/components/Label/Label.types.ts
```diff
@@ -1,5 +1,8 @@
+import { contentColors } from "../../utils/colors-vars-map";
+
 export type LabelSizes = "small" | "medium";
 
 export type LabelKind = "fill" | "line";
 
-export type LabelColor = "primary" | "dark" | "negative" | "positive";
+type ContentColor = (typeof contentColors)[number];
```
talkor, line 7: export?

packages/core/src/components/Label/Label.tsx
```diff
@@ -70,6 +71,13 @@ const Label: VibeComponent<LabelProps> & {
...
+      if (contentColors.includes(color as ContentColor)) {
+        return { backgroundColor: `var(--color-${color})` };
```
talkor, line 77: Why not use `getElementColor` (with adjustments) like we use in Chips?

shaharzil: the getElementColor dont support all the colors that label supports

packages/core/src/components/Label/Label.tsx
shaharzil: Fixed, thanks

packages/core/src/components/Label/Label.tsx
talkor: What's missing? Wouldn't it be an inconsistency? Maybe we can adjust the current function for that? It might be confusing to have two ways to get content colors

talkor (approved):

PR #2769, opened 2025-02-18
fix(VirtualizedList): sync scroll position to handle React 18 batching

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/113184182/pulses/8510096252


talkor (approved):
👑 👑 👑 👑 👑

PR #2770, opened 2025-02-19
fix(Tipseen): setting width more than max-width in tooltip default width has no effect

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/113184182/pulses/8516533100


rivka-ungar:
> Good catch! Do you know why it hasn't been reproduce in the playground?

It does reproduce :)

PR #2771, opened 2025-02-19
feat(List): add role props

talkor:
Example:
```
<List role="list">
  <ListItem role="listitem" />
</List>  
``` 
https://monday.monday.com/boards/3532714909/pulses/8109543863

### DIFF
--- a/packages/core/src/components/List/List.tsx
+++ b/packages/core/src/components/List/List.tsx
@@ -1,6 +1,7 @@
 import cx from "classnames";
 import React, {
   AriaAttributes,
+  AriaRole,
   CSSProperties,
   forwardRef,
   ReactElement,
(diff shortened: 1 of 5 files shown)


packages/core/src/components/List/List.tsx
```diff
@@ -52,6 +53,14 @@ export interface ListProps extends VibeComponentProps {
...
   renderOnlyVisibleItems?: boolean;
   style?: CSSProperties;
+  /**
+   * ARIA role for the list. Default is "listbox".
+   */
+  role?: AriaRole;
+  /**
+   * ARIA role for the list items. Default is "option".
+   */
+  itemRole?: AriaRole;
```
YossiSaadi, line 63: Will all the list items necessarily would have the same role if we wanna allow changing the role?

YossiSaadi (approved):
Just a thought.
Also, do we really wanna allow it? If you need to change it, doesn't it mean this is probably not the component you need?

packages/core/src/components/List/List.tsx
talkor: Changed it to be a prop in the ListItem if you want to take another look
This is a legitimate case as there are several legit roles for a list

rivka-ungar (approved):

packages/core/src/components/List/utils/ListUtils.ts
```diff
@@ -1,4 +1,4 @@
-import { MutableRefObject, useState } from "react";
+import { AriaRole, MutableRefObject, useState } from "react";
```
rivka-ungar, line 1: `AriaRole` is not used

YossiSaadi (approved):

PR #2775, opened 2025-02-20
fix(useIsOverflowing) - default width tolerance of 2 to fix incorrect overflow in safari

itaymndy:


YossiSaadi:
Solved at #2820

PR #2779, opened 2025-02-25
docs: enable live-edit to all stories

shaharzil:
https://monday.monday.com/boards/7538077190/pulses/7538089840


rivka-ungar (commented):
Instead of defining scope separately for each story, it can be added once in the default export with all possible values. That way, it’s cleaner and easier to maintain.

rivka-ungar (approved):

PR #2791, opened 2025-03-10
fix(Modal): closeButtonTheme fixed color

rivka-ungar:
https://monday.monday.com/boards/6965884836/views/151260495

### DIFF
--- a/packages/core/src/components/Modal/ModalTopActions/ModalTopActions.tsx
+++ b/packages/core/src/components/Modal/ModalTopActions/ModalTopActions.tsx
@@ -3,16 +3,15 @@ import styles from "./ModalTopActions.module.scss";
 import { ModalTopActionsButtonColor, ModalTopActionsTheme, ModalTopActionsProps } from "./ModalTopActions.types";
 import IconButton from "../../IconButton/IconButton";
 import { CloseMedium } from "@vibe/icons";
-import { ButtonColor } from "../../Button/ButtonConstants";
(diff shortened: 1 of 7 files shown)


packages/core/src/components/Modal/ModalTopActions/ModalTopActions.tsx
```diff
@@ -3,16 +3,15 @@ import styles from "./ModalTopActions.module.scss";
...
 import IconButton from "../../IconButton/IconButton";
 import { CloseMedium } from "@vibe/icons";
-import { ButtonColor } from "../../Button/ButtonConstants";
 import { ComponentDefaultTestId } from "../../../tests/constants";
 
 const colorToButtonColor: Record<ModalTopActionsTheme, ModalTopActionsButtonColor> = {
-  dark: ButtonColor.ON_INVERTED_BACKGROUND,
-  light: ButtonColor.ON_PRIMARY_COLOR
+  dark: "fixed-dark",
+  light: "fixed-light"
```
shaharzil, line 10: Is it going to affect only modals with media? We don't want otherwise the color won't change in dark/light mode for basic modal

rivka-ungar: As far as I understand the dark and light variations were only ment for the modals with the media, the basic layout should only be used with the default primary color

talkor (approved):

PR #2794, opened 2025-03-11
feat(Link): add color variants

talkor:
https://monday.monday.com/boards/3532714909/pulses/8513913191


packages/core/src/components/Link/Link.module.scss
```diff
@@ -10,6 +10,14 @@
   text-decoration: none;
   writing-mode: horizontal-tb;
   @include focus-style();
+
```
rivka-ungar, line 13: Since now the default is primary should the default color be added in primary class?

```
 &.colorPrimary {
    color: var(--link-color);
  }
```

rivka-ungar (approved):

packages/core/src/components/Link/Link.module.scss
talkor: Hmm it can, WDYT? Better for consistency?

talkor: Oh yeah it should since it's adding the class anyway, I'll add 👍🏼

PR #2798, opened 2025-03-13
types: export typography types

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/113184182/pulses/8690016083


rivka-ungar:
> Why do we need it? It's an internal component, isn't it?

We got a request for the typography types

talkor:
> > Why do we need it? It's an internal component, isn't it?
> 
> We got a request for the typography types

I think we should export just the types and not the component

talkor (approved):

PR #2807, opened 2025-03-18
fix(Dropdown): use require for test envs only

talkor:
https://monday.monday.com/boards/3532714909/pulses/8766251042


packages/core/src/components/Dropdown/Dropdown.tsx
```diff
@@ -150,7 +150,7 @@ const Dropdown: VibeComponent<DropdownComponentProps, HTMLElement> & {
           isRequireAvailable = false;
         }
 
-        if (isRequireAvailable) {
+        if (isRequireAvailable && process.env.NODE_ENV === "test") {
```
MosheZemah, line 153: I don't think it works on browser

MosheZemah (changes requested):

MosheZemah (approved):

packages/core/src/components/Dropdown/Dropdown.tsx
talkor: Added rollup plugin to replace it

rivka-ungar (approved):

PR #2809, opened 2025-03-20
feat(Tipseen): add submitButtonIcon prop

talkor:
https://monday.monday.com/boards/3532714909/pulses/8729751543


rivka-ungar (approved):
Should an icon be added to the dismiss button as well so they have consistent behavior?

talkor:
> Should an icon be added to the dismiss button as well so they have consistent behavior?

Currently as a design decision it's only required on the Submit

PR #2811, opened 2025-03-20
chore(DropdownNew): add events

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/113184182/pulses/8687640350

### DIFF
--- a/packages/core/src/components/DropdownNew/Dropdown.tsx
+++ b/packages/core/src/components/DropdownNew/Dropdown.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useRef, forwardRef } from "react";
+import React, { useMemo, useRef, forwardRef, useState } from "react";
 import cx from "classnames";
 import { DialogContentContainer } from "../DialogContentContainer";
 import { BaseInput } from "../BaseInput";
@@ -31,8 +31,16 @@ const Dropdown: VibeComponent<BaseDropdownProps<BaseListItemProps>, HTMLDivEleme
       disabled,
       readOnly,
       error,
+      onBlur,
+      onChange,
+      onClear,
+      onFocus,
       onInputChange,
+      onKeyDown,
+      onMenuOpen,
+      onMenuClose,
       onOptionSelect,
+      onScroll,
       className,
       id,
       "data-testid": dataTestId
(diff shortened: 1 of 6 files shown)


packages/core/src/components/DropdownNew/Dropdown.tsx
```diff
@@ -105,8 +139,8 @@ const Dropdown: VibeComponent<BaseDropdownProps<BaseListItemProps>, HTMLDivEleme
         <div
           style={{
             ...popoverStyles.popper,
-            visibility: isOpen ? "visible" : "hidden",
-            width: dropdownMergedRef.current?.offsetWidth
+            width: dropdownMergedRef.current?.offsetWidth,
+            zIndex: 2
```
talkor, line 143: Is it needed?

packages/core/src/components/DropdownNew/Dropdown.tsx
```diff
@@ -105,8 +139,8 @@ const Dropdown: VibeComponent<BaseDropdownProps<BaseListItemProps>, HTMLDivEleme
         <div
           style={{
             ...popoverStyles.popper,
-            visibility: isOpen ? "visible" : "hidden",
```
talkor, line 108: No need for it anymore?

packages/core/src/components/DropdownNew/__tests__/Dropdown.test.tsx
```diff
@@ -214,4 +190,179 @@ describe("DropdownNew", () => {
       }
     });
   });
+
+  describe("event handlers", () => {
```
talkor, line 194: Nice

talkor (approved):

packages/core/src/components/DropdownNew/Dropdown.tsx
rivka-ungar: yep :(
for the dialog content to be on top of the input

packages/core/src/components/DropdownNew/Dropdown.tsx
rivka-ungar: It was my mistake I added it in the first place, It is set in the [hook](https://github.com/mondaycom/vibe/blob/55f0d346d3e98c5ad4d5fa5bacaa485b8b25e39c/packages/core/src/hooks/usePopover.ts#L51), so no need to set it again.

PR #2813, opened 2025-03-23
feat(Menu): update selectItem method to be more strict

hanan-monday:
Update method to filter using regex with exact word instead of partial

### DIFF
--- a/packages/testkit/package.json
+++ b/packages/testkit/package.json
@@ -37,7 +37,7 @@
     "url": "https://github.com/mondaycom/vibe/issues"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.0",
+    "@playwright/test": "1.49.1",
     "@typescript-eslint/eslint-plugin": "^5.36.1",
     "@typescript-eslint/parser": "^5.36.1",
     "eslint": "^8.23.0",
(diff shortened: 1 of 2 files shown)


packages/testkit/package.json
```diff
@@ -37,7 +37,7 @@
     "url": "https://github.com/mondaycom/vibe/issues"
   },
   "devDependencies": {
-    "@playwright/test": "^1.50.0",
+    "@playwright/test": "1.49.1",
```
YossiSaadi, line 40: why exact version? and are you sure about a downgrade?

packages/testkit/package.json
```diff
@@ -47,6 +47,6 @@
     "prettier": "^2.5.1"
   },
   "peerDependencies": {
-    "@playwright/test": "^1.50.0"
+    "@playwright/test": "1.49.1"
```
YossiSaadi, line 50: same

packages/testkit/package.json
hanan-monday: Because we are using playwright image when running in CI.
playwright image must be compatible with the playwright version so we want to make it exact version - otherwise we will get a comparabilities issue.

https://playwright.dev/docs/docker#image-tags

PR #2819, opened 2025-03-25
feat(usePopover): add fallbackPlacements prop

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8773662592


YossiSaadi (approved):
I like it better than the previous code
btw, is it instead the disableFlip?

PR #2824, opened 2025-03-27
feat(Dropdown): use layer context for the dropdown in case used inside Dialog or Modal

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8794876358

### DIFF
--- a/packages/core/src/components/Modal/layouts/ModalSideBySideLayout/__stories__/ModalSideBySideLayout.stories.tsx
+++ b/packages/core/src/components/Modal/layouts/ModalSideBySideLayout/__stories__/ModalSideBySideLayout.stories.tsx
@@ -179,6 +179,7 @@ export const Wizard: Story = {
               <FieldLabel labelText="Language preferences" />
               <Dropdown
                 insideOverflowWithTransformContainer
+                insideLayerContext
                 size="small"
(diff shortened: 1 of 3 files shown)


rivka-ungar (approved):

packages/core/src/components/Modal/layouts/ModalSideBySideLayout/__stories__/ModalSideBySideLayout.stories.tsx
```diff
@@ -179,6 +179,7 @@ export const Wizard: Story = {
               <FieldLabel labelText="Language preferences" />
               <Dropdown
                 insideOverflowWithTransformContainer
+                insideLayerContext
```
talkor, line 182: I'd also maybe rename the prop

talkor (approved):
Great!

packages/core/src/components/Modal/layouts/ModalSideBySideLayout/__stories__/ModalSideBySideLayout.stories.tsx
YossiSaadi: I couldn't think of a better name 🤷🏼 @rivka-ungar wdyt?

PR #2827, opened 2025-04-01
feat(Text): add error color

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8830567066


YossiSaadi (approved):
Does it need an additional changes? Or is it that straightforward?
Also maybe worth documenting it with design in SB.

rivka-ungar:
> Does it need an additional changes? Or is it that straightforward? Also maybe worth documenting it with design in SB.

The designers want to allow the option when needed but not have it documented in the stories.

PR #2828, opened 2025-04-01
docs: add ARIA roles section

benyitzhaki:
docs(Dropdown): add ARIA roles section and update do's and don'ts formatting

- [ ] I have read the [Contribution Guide](../CONTRIBUTING.md) for this project.

Resolves #


packages/core/src/components/Dropdown/__stories__/Dropdown.mdx
```diff
@@ -37,6 +38,32 @@ import { Dropdown } from "@vibe/core";
 
 <PropsTable />
 
+## ARIA Roles
+
+The Dropdown component exposes the following ARIA roles and attributes for testing and accessibility:
+
+### Default Roles
```
laviomri, line 45: If this changes, how will we know to keep this doc up to date? (Does it make sense that the roles will change with time? @talkor )

PR #2830, opened 2025-04-02
fix(Modal): update background color to use new token

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8814383107


talkor (approved):

YossiSaadi (approved):
QUEEN 👑

PR #2831, opened 2025-04-03
feat: add data-testid to Box

alexandrzhd:
Added missing data-testid to Box component

Resolves # https://monday.monday.com/boards/6965884836/pulses/8818397254

### DIFF
--- a/packages/core/src/components/Box/Box.tsx
+++ b/packages/core/src/components/Box/Box.tsx
@@ -170,6 +170,7 @@ const Box: VibeComponent<BoxProps> & {
     return React.createElement(
       elementType,
       {
+        "data-testid": props["data-testid"],
         ref: mergedRef,
         className: cx(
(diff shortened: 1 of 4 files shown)


YossiSaadi (approved):

packages/core/src/components/Box/Box.tsx
```diff
@@ -170,6 +170,7 @@ const Box: VibeComponent<BoxProps> & {
     return React.createElement(
       elementType,
       {
+        "data-testid": props["data-testid"],
```
talkor, line 173: @alexandrzhd does it have a default value like the other components?

alexandrzhd: @talkor fair enough! added default data-testid like in other components

talkor (approved):
Awesome! Thanks @alexandrzhd !

PR #2833, opened 2025-04-06
feat: add metadata for LLMs

YossiSaadi:
https://monday.monday.com/boards/3532715121/pulses/8869841166


packages/core/.eslintrc.cjs
```diff
@@ -51,7 +51,20 @@ const commonRules = {
...
+    "error",
+    {
+      devDependencies: [
+        "src/scripts/**/*.ts",
+        "src/storybook/**/*",
+        "src/tests/**/*.ts",
+        "src/utils/jest-utils.js",
+        "**/__tests__/**/*.{js,ts,tsx}",
+        "**/__stories__/**/*.{js,ts,tsx}",
+        "src/components/Combobox/components/ComboboxConstants.ts"
```
talkor, line 64: Why this?

packages/core/src/scripts/generate-metadata.ts
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    newRecords.push(r);
+  }
+
+  return newRecords;
+}
+
+function main() {
+  const startTime = performance.now();
```
talkor, line 376: Do we need it in the CI?

packages/core/package.json
```diff
@@ -66,9 +70,10 @@
     "test:update": "yarn test -u",
     "test:coverage": "yarn test -- --coverage",
     "start": "yarn build:esm -- --watch",
-    "build": "yarn build:esm && yarn build:esm:mock-classnames",
+    "build": "yarn build:esm && yarn build:esm:mock-classnames && yarn build:metadata",
```
talkor, line 73: Separate it from build? Since it takes 2min

talkor (commented):
Looks great! Let's just separate the step outside build

packages/core/package.json
YossiSaadi: done!

packages/core/src/scripts/generate-metadata.ts
YossiSaadi: Just for debugging, I've added it intentionally for CI

packages/core/.eslintrc.cjs
YossiSaadi: Because of what we saw in the issue of ComboboxConstants with preact import
I'll handle it in a different PR to not mix בשר & חלב

talkor (approved):

PR #2835, opened 2025-04-06
feat(MenuTitle): be able to provide a ReactNode

talkor:
https://monday.monday.com/boards/3532714909/pulses/8869281367


YossiSaadi (approved):

YossiSaadi:
change the JSDoc maybe as well

talkor:
> change the JSDoc maybe as well

When we added all the prop descriptions we tried to avoid to document the types or accepted values/defaults in the description as they can change while the descriptions will most likely be left outdated, and you can get the correct type/value using TS anyway

PR #2850, opened 2025-04-23
fix(Combobox): update scrollRef type to MutableRefObject

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8975685665


packages/core/src/components/Combobox/components/ComboboxConstants.ts
```diff
@@ -1,6 +1,5 @@
 import { SubIcon } from "../../../types";
 import React from "react";
-import { MutableRef } from "preact/hooks";
```
YossiSaadi, line 3: wth? 🫠

rivka-ungar (approved):

PR #2851, opened 2025-04-23
fix(docs): docgen can't handle ComponentType

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/8988817874

Also:
Switched to use react-docgen-typescript in our Storybook default docgen. Check Storybook docs:
<https://storybook.js.org/docs/api/main-config/main-config-typescript#reactdocgen>
> react-docgen-typescript invokes the TypeScript compiler, which makes it slow but generally accurate. react-docgen performs its own analysis, which is much faster but incomplete.

### DIFF
--- a/packages/core/.storybook/main.ts
+++ b/packages/core/.storybook/main.ts
@@ -54,7 +54,8 @@ const config: StorybookConfig = {
     check: true,
     checkOptions: {
       async: false
-    }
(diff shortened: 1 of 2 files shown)


rivka-ungar (approved):
👑 👑 👑

PR #2852, opened 2025-04-24
fix: Fix `<ButtonGroup>` border collapse issue

sahariko:
Ensure individual buttons within the `<ButtonGroup>` component don't have double borders.

This is achieved by a set of 5 CSS rules:

1. All buttons have 0 left and right borders.
1. The first button always has a left border.
1. The last button always has a right border.
1. The active button always has all borders.
1. Any button after a button that is not active has a left border.

#### Before

image

#### After

image

### DIFF
--- a/packages/core/src/components/ButtonGroup/ButtonGroup.module.scss
+++ b/packages/core/src/components/ButtonGroup/ButtonGroup.module.scss
@@ -55,10 +55,6 @@
     }
   }
 
-  // .button:not(.activeButton) + .button {
-  //   border-inline-start-width: 1px;
-  // }
-
   .kindTertiary & {
     .button {
       border-color: transparent;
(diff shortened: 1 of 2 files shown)


packages/core/src/components/ButtonGroup/ButtonGroup.module.scss
```diff
@@ -13,21 +13,56 @@
...
+    border-bottom-width: 1px;
+    background-color: var(--primary-background-color);
+    border-color: var(--ui-border-color);
 
-.buttonsContainer .button.activeButton {
-  background-color: var(--primary-selected-color);
-  border-color: var(--primary-color);
-}
+    &:first-of-type {
+      border-left-width: 1px;
```
rivka-ungar, line 27: Could we switch to logical properties?

rivka-ungar (approved):

YossiSaadi:
not a blocker - we can create a task to fix this kind of usages, but check this one
https://62037c72e0e4d6004a9a450f-vlsjobbmec.chromatic.com/?path=/story/components-buttongroup--disabled-singe-button

YossiSaadi:
> not a blocker - we can create a task to fix this kind of usages, but check this one [62037c72e0e4d6004a9a450f-vlsjobbmec.chromatic.com?path=/story/components-buttongroup--disabled-singe-button](https://62037c72e0e4d6004a9a450f-vlsjobbmec.chromatic.com/?path=/story/components-buttongroup--disabled-singe-button)

I would have probably expected it to look like this
image
and it currently looks like this
image
probably need not to use first-of-type, but first and last

PR #2862, opened 2025-05-03
fix(Table): Memoized row for optimized row rendering

rivka-ungar:
Addresses performance issues in TableVirtualizedBody, particularly noticeable during fast scrolling with many columns or complex cell content. This often resulted in excessive re-renders, blank spaces appearing, and compatibility problems with sticky columns.
Changes:
- Internal Row Memoization: Implements React.memo internally for rendered rows. This significantly reduces unnecessary re-renders when scrolling, as rows only update if their data changes, not just their position.
Before
Screenshot 2025-05-03 at 22 33 20

After
Screenshot 2025-05-03 at 22 32 47

- Configurable Overscanning: Adds an optional overscanCount prop (defaulting to 1). Consumers can increase this value to render more rows outside the viewport, potentially smoothing out fast scrolls at the cost of rendering more items.


talkor (approved):
Wowwww!! Amazing work!!! 🥇 👑 👏🏼

PR #2873, opened 2025-05-08
feat: add llm context to docs

yuhgto:
# Description
Added "llm-context" – instructions to make LLMs more likely to use Vibe correctly. 

I am working on some tools for developers who are using AI-assisted coding. One of them is "LLM context", basically some readme files that make it more likely for LLMs to use Vibe, GraphQL, etc correctly. 

Initially I had them as their own repo but Vlad suggested I add them to the Vibe project directly, so I made a PR.

Hope this contribution was appropriate! Let me know if you have any questions.

This is part of an epic to make our public dev tools easier to use with AI. 

## Open questions
- Is the location appropriate? I put these rules inside `/docs` folder
- Please review [README.md](https://github.com/yuhgto/vibe/blob/feat/dipro/add-llm-context/docs/llm-context/README.md) and let me know if it's understandable
- Boilerplate for each component was copied from our Storybook ([example](https://github.com/yuhgto/vibe/blob/feat/dipro/add-llm-context/docs/llm-context/components/vibe-avatargroup-component.mdc)). Could we automate how we generate these files?

# Demo

## Without context
Without the context, Cursor makes up syntax for adding a table component (this is from a real prompt): 

image

This is the code it wrote: incorrect usage of Table, 
```
<div className="table-wrapper">
  <Table
    columns={columns}
    items={users}
  />
</div>
```

## With context

If you attach a context file that has some boilerplate in it, it does much better:

image
```
<div className="table-wrapper">
  <Table
    columns={columns}
    emptyState={<TableEmptyState />}
    errorState={<TableErrorState />}
  >
    <TableHeader>
      {columns.map((column) => (
        <TableHeaderCell
          key={column.id}
          title={column.title}
        />
      ))}
    </TableHeader>
    <TableBody>
      {users.map((user) => (
        <TableRow key={user.id}>
          <TableCell>{user.name}</TableCell>
          <TableCell>{user.email}</TableCell>
          <TableCell>{user.type}</TableCell>
        </TableRow>
      ))}
    </TableBody>
  </Table>
</div>
```

- [X] I have read the [Contribution Guide](../CONTRIBUTING.md) for this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ import { Button } from "@vibe/core";
 - [monday-ui-style](https://github.com/mondaycom/vibe/blob/master/packages/style/README.md): Styling foundations (included in @vibe/core)
 - [vibe-storybook-components](https://github.com/mondaycom/vibe/blob/master/packages/storybook-blocks/README.md): Vibe Storybook Blocks
 - [storybook-addon-playground](https://github.com/mondaycom/storybook-addon-playground/): A Component Playground Addon for Storybook
+- [llm-context](https://github.com/mondaycom/vibe/blob/master/docs/llm-context/README.md): Rule files for AI-assisted coding
 
 ## Older Versions
 
--- a/docs/llm-context/README.md
+++ b/docs/llm-context/README.md
@@ -1,0 +1,37 @@
+# Context & Documentation for AI-assisted coding
+
+This package contains context and rules for LLM-assisted coding. 
+
+With context, an LLM is more likely to write correct code. Without it, they may write nonsensical JSX or use components in the wrong way. 
+
+The context is bundled as:
+- vibe-usage-best-practices.md - General best practices for importing and using Vibe components
+- /components - Rules and boilerplate for specific components
+
+Each context file is a markdown file with some YAML front matter. The rules are split into multiple files to prevent LLM confusion. 
+
+## Installation
+
+You can copy the files directly or clone the subdirectory using degit. 
+```
+npx degit https://github.com/mondaycom/vibe/docs/llm-context
+```
+
+Clone into a subdirectory: 
+```
+npx degit https://github.com/mondaycom/vibe/docs/llm-context .cursor/rules
+```
+
+## Usage with Cursor
+
+1. Copy the files into `.cursor/rules` directory (project rules)
+2. Add the files to Cursor's context by @mentioning them in the Cursor chat window (recommended). 
+
+[Cursor Project Rules Documentation](https://docs.cursor.com/context/rules#project-rules)
+
+## Usage with Copilot
+
+1. Copy the files into a project-level context folder
+2. Attach the relevant file (or all of them) to your query with #mention
+
+[Copilot Chat Context Documentation](https://code.visualstudio.com/docs/copilot/chat/copilot-chat-context)
--- a/docs/llm-context/components/vibe-accordion-component.mdc
+++ b/docs/llm-context/components/vibe-accordion-component.mdc
@@ -1,0 +1,70 @@
+---
+description: Accordion component usage and examples
+globs: 
+alwaysApply: false
+---
+# Accordion component
+
+- Use the accordion component for expandable lists.
+- The child of `AccordionItem` is the content to be hidden or expanded. 
+
+## MultiActive
+
+```jsx
+<Accordion className="small-wrapper" allowMultiple defaultIndex={[1, 3]}>
+      <AccordionItem title="Notifications">
+        <div className="small-box" />
+      </AccordionItem>
+      <AccordionItem title="Setting">
+        <div className="small-box" />
+      </AccordionItem>
+      <AccordionItem title="Connectivity">
+        <div className="small-box" />
+      </AccordionItem>
+      <AccordionItem title="Integration">
+        <div className="small-box" />
+      </AccordionItem>
+      <AccordionItem title="Assets">
+        <div className="small-box" />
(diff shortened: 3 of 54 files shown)


YossiSaadi (commented):
Hey!
Thank you for your contribution to Vibe! It's evident that you've invested significant effort into this PR, and your initiative is genuinely appreciated. 

We've been exploring ways to expose our component metadata - already available at @vibe/core/meta - through an MCP Tool. Regarding docs - we want to implement a RAG approach - also with an MCP Tool.
We believe that, with those two, using Vibe with an LLM would be a 10x better and more accurate experience. 

Regarding the .md files you've introduced for each component, I have a few considerations:

1. Maintenance Overhead: Given that our components are frequently updated, maintaining individual .md files for each could become challenging. Without an automated update mechanism, there's a risk of these files becoming outdated, leading to inconsistencies. 

2. User-Side Limitations: Storing those rules in the user's .cursor/rules directory means we lack control over updates. This setup could hinder our ability to ensure users have the most recent and accurate rules, potentially leading to discrepancies in their development environment. 

I understand that your approach might offer great immediate benefits, especially in terms of improved results. However, considering the points above, I believe utilizing our existing metadata might offer a more sustainable and scalable solution. We're aware it still does not expose our docs, and we're looking into tackling it, as I've stated above. 

That said, I'm open to understanding more about your approach. Have you seen similar implementations elsewhere? Are there best practices or patterns you're drawing inspiration from? Sharing these insights could provide valuable context and help us make an informed decision.

Looking forward to your thoughts and further discussion!

yuhgto:
Hey thanks for your message and feedback @YossiSaadi !

What about including these rules as a short-term solution for developers? We can remove them when we have our MCP set up. 

**I have some other questions/direct comments:** 

Vibe MCP - What tools are you thinking of adding to this? I'm struggling to imagine it. Most of the MCPs I've seen do tasks (such as add items to a monday board, run build tasks, etc), but this use case is to help the LLM write better code. 

Metadata - this is super interesting. I didn't know about it. How do I use it? 

Exposing docs using RAG - this is what happens under the hood when you @mention a doc in Cursor/Copilot. 

**Further context about this PR:** 

This PR was intended as a quick, stopgap solution while we are still building out our MCPs. The API team had a similar plan (upcoming MCP) and I've submitted some rules for their SDK nevertheless. 

I chose this approach because I got feedback from a few developer partners that they are using cursor and the rules help them out. 

Plus, Cursor limits how many MCP tools you can use at a time (40 tools), so this is a way around that.

YossiSaadi:
> Hey thanks for your message and feedback @YossiSaadi !
> 
> What about including these rules as a short-term solution for developers? We can remove them when we have our MCP set up.
> 
> **I have some other questions/direct comments:**
> 
> Vibe MCP - What tools are you thinking of adding to this? I'm struggling to imagine it. Most of the MCPs I've seen do tasks (such as add items to a monday board, run build tasks, etc), but this use case is to help the LLM write better code.
> 
> Metadata - this is super interesting. I didn't know about it. How do I use it?
> 
> Exposing docs using RAG - this is what happens under the hood when you @mention a doc in Cursor/Copilot.
> 
> **Further context about this PR:**
> 
> This PR was intended as a quick, stopgap solution while we are still building out our MCPs. The API team had a similar plan (upcoming MCP) and I've submitted some rules for their SDK nevertheless.
> 
> I chose this approach because I got feedback from a few developer partners that they are using cursor and the rules help them out.
> 
> Plus, Cursor limits how many MCP tools you can use at a time (40 tools), so this is a way around that.
> 
> Thank you again!

Hey!
Having those as a short-term solution can work out, but let's say 20 devs would use it, and when we decide to remove this, how can we make sure they remove it as well? they would always have outdated usage info of Vibe.
As part of our internal tools, did you tried use one of those? Slack me for more info on internal MCPs if you're not familiar (they're doing pretty good job!).

And regarding to a public MCP - currently Cursor does not support "Resources", but once it will, it would be a good idea to expose the tools I mentioned earlier - as resources. Currently this is the way to go..

I'll give quick example of a possible tool:
```
  name: 'get-vibe-component-metadata',
  description:
    'Get the metadata for a specific @vibe/core or @vibe/core/next component, or all components if no name is provided.',
  inputSchema: {
    componentName: z
      .string()
      .optional()
      .describe(
        'Optional. The name of the component to get metadata for (e.g., Button, TextField). If omitted, returns metadata for all components.'
      ),
  },
  if (componentName) {
        // Filter metadata by display name
        const componentMeta = vibeMetadata.filter(component => component.displayName === componentName);

        if (componentMeta.length === 0) {
          return {
            content: `Error in get-vibe-component-metadata: Component '${componentName}' not found.`,
            isError: true,
          };
        }

        // Prefer the 'next' version if exists
        const nextComponent = componentMeta.find(component => component.aggregator === 'next');
        const result = nextComponent || componentMeta[0];
        return {
          content: JSON.stringify(result, null, 2),
        };
      }

      // Return all metadata for all components if componentName is not provided
      return {
        content: JSON.stringify(vibeMetadata, null, 2),
      };
```

Regarding RAG - I'm aware this is how it works with Cursor, but having an MCP would mean you'll always get the updated version. a Rule file would NEVER be updated unless you do it manually. which means that on every version, if we automate the process, we would need to tell the user to copy & paste the new files to the folder.

Let's discuss more possible solutions that would not harm users with outdated documentation/rules. wdyt?

yuhgto:
Closing this! We implemented this as an MCP tool instead: #2901

PR #2878, opened 2025-05-11
docs(LiveEdit): remember last known valid code for better DX

YossiSaadi:
How it looks now:
image

It helps avoid "jumping" in the page when editing. it always remember the last known valid code.

https://monday.monday.com/boards/3532714909/pulses/6495484558

### DIFF
--- a/packages/core/src/storybook/components/live-preview/LivePreview.module.scss
+++ b/packages/core/src/storybook/components/live-preview/LivePreview.module.scss
@@ -1,0 +1,28 @@
+.wrapper {
+  position: relative;
+}
+
+.overlay {
+  position: fixed;
+  inset: 0;
+  color: #fff;
+  background-color: rgba(0, 0, 0, 0.7);
+  padding: 16px;
+  margin: 0;
+  z-index: 1;
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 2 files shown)


packages/core/src/storybook/components/live-preview/LivePreview.module.scss
talkor, line 1: You can use tokens in here

talkor (approved):
Nice!

packages/core/src/storybook/components/live-preview/LivePreview.module.scss
YossiSaadi: yep, it's on purpose
we did same for Playground
it is intended to be open sourced (one day), so I don't want to use anything related to Vibe

talkor: Actually regardless of Vibe if we'd use tokens here anyway (of some sort) it will give users the ability to customize

YossiSaadi: hmm that's a good point
I haven't done it in the Playground
it can really help making this customizable
I'll create two tasks for those

PR #2884, opened 2025-05-15
chore(DropdownNew): Improve type safety with conditional props for multi select

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/9116167582

### DIFF
--- a/packages/core/src/components/DropdownNew/Dropdown.types.ts
+++ b/packages/core/src/components/DropdownNew/Dropdown.types.ts
@@ -4,7 +4,38 @@ import { VibeComponentProps } from "../../types";
 import { BaseListItemData } from "../BaseListItem/BaseListItem.types";
 
 export type DropdownGroupOption<Item = Record<string, unknown>> = ListGroup<Item>[] | BaseListItemData<Item>[];
-export interface BaseDropdownProps<Item extends BaseListItemData<Record<string, unknown>>> extends VibeComponentProps {
+
+interface MultiSelectSpecifics<Item extends BaseListItemData<Record<string, unknown>>> {
+  /**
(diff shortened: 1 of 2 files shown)


packages/core/src/components/DropdownNew/Dropdown.types.ts
```diff
@@ -133,16 +156,12 @@ export interface BaseDropdownProps<Item extends BaseListItemData<Record<string,
    * Callback fired when an option is selected.
    */
   onOptionSelect?: (option: BaseListItemData<Item>) => void;
```
YossiSaadi, line 158: is this fired on all cases?

packages/core/src/components/DropdownNew/Dropdown.types.ts
YossiSaadi, line 1: `valueRenderer` as well, no?
`onOptionSelect`?

packages/core/src/components/DropdownNew/Dropdown.types.ts
rivka-ungar: yes

packages/core/src/components/DropdownNew/Dropdown.types.ts
rivka-ungar: Good catch, added `valueRenderer`. The `onOptionSelect` is fired on both cases

YossiSaadi (approved):

packages/core/src/components/DropdownNew/Dropdown.types.ts
rivka-ungar: This is the behavior of the current dropdown, i will check it out, and if needed change in a later pr

PR #2886, opened 2025-05-15
chore(DropdownNew): add showSelectedOptions prop

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/9116167582

Screenshot 2025-05-15 at 19 24 14

### DIFF
--- a/packages/core/src/components/DropdownNew/hooks/useDropdownCombobox.ts
+++ b/packages/core/src/components/DropdownNew/hooks/useDropdownCombobox.ts
@@ -1,4 +1,4 @@
-import { useCallback, useMemo } from "react";
+import { useCallback, useMemo, useState } from "react";
 import { useCombobox } from "downshift";
 import useDropdownFiltering from "./useDropdownFiltering";
 import { BaseListItemData } from "../../BaseListItem";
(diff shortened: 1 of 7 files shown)


packages/core/src/components/DropdownNew/hooks/useDropdownCombobox.ts
```diff
@@ -9,14 +9,27 @@ function useDropdownCombobox<T extends BaseListItemData<Record<string, unknown>>
   autoFocus?: boolean,
   isMenuOpen?: boolean,
   closeMenuOnSelect?: boolean,
-  onChange?: (option: T | T[]) => void,
+  onChange?: (option: T | null) => void,
```
YossiSaadi, line 12: why?

packages/core/src/components/DropdownNew/hooks/useDropdownCombobox.ts
```diff
@@ -9,14 +9,27 @@ function useDropdownCombobox<T extends BaseListItemData<Record<string, unknown>>
...
-  filterOption?: (option: T, inputValue: string) => boolean
+  filterOption?: (option: T, inputValue: string) => boolean,
+  showSelectedOptions?: boolean
 ) {
-  const { filteredOptions, filterOptions } = useDropdownFiltering<T>(options, filterOption);
+  const [currentSelectedItem, setCurrentSelectedItem] = useState<T | null>(null);
+
+  const memoizedSelectedItemForFiltering = useMemo(() => {
+    return currentSelectedItem ? [currentSelectedItem] : [];
+  }, [currentSelectedItem]);
```
YossiSaadi, line 24: I think it's cool to just pass it inline to `useDropdownFiltering`, this does not include heavy calculation to consider having it in a memo

YossiSaadi (approved):

packages/core/src/components/DropdownNew/hooks/useDropdownCombobox.ts
rivka-ungar: It causes an infinite re-render loop

packages/core/src/components/DropdownNew/hooks/useDropdownCombobox.ts
rivka-ungar: Mistake, fixed

packages/core/src/components/DropdownNew/hooks/useDropdownCombobox.ts
YossiSaadi: it sound like there's other issue then 👀 why does memoizing it fixes it 🤔

rivka-ungar: useDropdownFiltering takes an array to support both single- and multi-select; if we pass [currentSelectedItem] inline a new array is created on every render, the hook thinks the selection changed each time and refilters forever, so we memoise it.

PR #2888, opened 2025-05-17
fix(Table): make row hover state immediate, robust and more reliable

YossiSaadi:
this was changed because on some scenarios, when moving from menu button to row and back, the menu disappeared - especially in edge cases involving rapid mouse movement

https://monday.monday.com/boards/3532714909/pulses/9173208813

#### 1. **Immediate Hover State Management**
- **Removed all setTimeouts** for hover state transitions between row and menu.
- Hover state (`hoveredRowRef`) is now updated immediately on mouse events, resulting in a more responsive and predictable UI.
- The code now uses a simple `isMenuHovered` ref to track when the mouse is over the menu, ensuring the menu remains visible when moving between row and menu.

#### 2. **Simplified Mouse Event Logic**
- `onMouseOverRow` and `onMouseOverRowMenu` now always immediately set the hover state and cancel any pending state clears.
- `onMouseLeaveRow` and `onMouseLeaveRowMenu` immediately clear the hover state if appropriate, without delay.
- This eliminates race conditions and flickering that could occur with delayed state changes.

### DIFF
--- a/packages/core/src/components/Table/TableContainer/TableContainer.module.scss
+++ b/packages/core/src/components/Table/TableContainer/TableContainer.module.scss
@@ -7,7 +7,7 @@
     --cell-width: 40px;
     position: absolute;
     overflow: hidden;
-    left: calc(calc(var(--cell-width) * -1) + 1px);
+    left: calc(1px - var(--cell-width));
     width: var(--cell-width);
     height: 100%;
   }
--- a/packages/core/src/components/Table/TableContainer/TableContainer.module.scss
+++ b/packages/core/src/components/Table/TableContainer/TableContainer.module.scss
@@ -7,7 +7,7 @@
     --cell-width: 40px;
     position: absolute;
     overflow: hidden;
-    left: calc(var(--cell-width) * -1);
+    left: calc(calc(var(--cell-width) * -1) + 1px);
     width: var(--cell-width);
     height: 100%;
   }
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Table/TableContainer/TableContainer.module.scss
```diff
@@ -7,7 +7,7 @@
     --cell-width: 40px;
     position: absolute;
     overflow: hidden;
-    left: calc(var(--cell-width) * -1);
+    left: calc(calc(var(--cell-width) * -1) + 1px);
```
YossiSaadi, line 10: this has 0 effect on UI, and means that when leaving the row menu towards the row, there isn't a flicker, and no setTimeout needed to compensate on this few ms of moving from one item to another

packages/core/src/components/Table/context/TableRowMenuContext/TableRowMenuContext.tsx
```diff
@@ -5,49 +5,46 @@ const TableRowMenuContext = createContext<ITableRowMenuContext | undefined>(unde
...
+      if (isMenuOpen) return;
 
-      resetTimeout();
       setHoveredRowRef(rowRef);
       const tableRootTop = tableRootRef.current.getBoundingClientRect().top;
       const rowTop = rowRef.current.getBoundingClientRect().top;
       setMenuButtonPosition(rowTop - tableRootTop);
-      if (hasMenuRef.current === null) {
-        setTimeout(() => {
-          hasMenuRef.current = !!document?.querySelector(`[data-row-menu-id]`);
```
rivka-ungar, line 28: If it is not needed any more the attribute `data-row-menu-id={rowId}` can be removed from the component

rivka-ungar (approved):
So much better!!

packages/core/src/components/Table/context/TableRowMenuContext/TableRowMenuContext.tsx
YossiSaadi: good catch!

PR #2903, opened 2025-05-22
chore: improve mcp metadata generation script

talkor:
https://monday.monday.com/boards/3532714909/pulses/9211164933


packages/core/scripts/generate-metadata.ts
```diff
@@ -225,11 +290,45 @@ function aggregatorMain(): AggregatorRecord[] {
...
+  if (filesToProcess.length === 0) {
+    return cachedResults;
+  }
+
+  // Configure parser with optimized settings
   const parser = withDefaultConfig({
     savePropValueAsString: true,
     shouldExtractLiteralValuesFromEnum: true,
     shouldRemoveUndefinedFromOptional: true,
+    skipChildrenPropWithoutDoc: true, // Skip children prop for performance
```
rivka-ungar, line 331: Are we sure we want to skip it?

talkor: No 😅 
Removed

rivka-ungar (approved):

PR #2905, opened 2025-05-24
fix(mcp): integrate metadata fetching from unpkg (with cache) instead of consuming @vibe/core

YossiSaadi:
- Deleted the `copy-metadata.js` script and its associated prebuild step.
- Introduced `MetadataService` to fetch and cache metadata from an external source.
- Updated `get-vibe-component-metadata.ts` and `list-vibe-public-components.ts` to utilize the new metadata fetching logic.


talkor (approved):
Amazing!

PR #2911, opened 2025-05-29
feat(mcp): add icon tools 

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/9256658050

### DIFF
--- a/packages/mcp/src/server/tools/icon-metadata-service.ts
+++ b/packages/mcp/src/server/tools/icon-metadata-service.ts
@@ -19,18 +19,14 @@ export class IconMetadataService {
 
   static async getIconMetadata(): Promise<IconMetadata[]> {
     if (this.cachedIconMetadata) {
-      console.error("[Vibe MCP] Using cached icon metadata");
       return this.cachedIconMetadata;
     }
 
     try {
-      console.error("[Vibe MCP] Fetching icon metadata from unpkg...");
       const metadata = await this.fetchWithRetry();
-      console.error(`[Vibe MCP] Fetched ${metadata.length} icons, caching for session`);
       this.cachedIconMetadata = metadata;
       return metadata;
     } catch (error) {
-      console.error("[Vibe MCP] Failed to fetch icon metadata:", error);
       throw new Error(
         "Unable to fetch icon metadata from remote sources. Please check your internet connection and try again."
       );
@@ -52,8 +48,6 @@ export class IconMetadataService {
       // The new format is: const e="Basic",t="Platform",i="View";var o=[...];export{o as default}
       return this.parseJavaScriptModule(jsCode);
     } catch (fetchError) {
-      console.error("[Vibe MCP] Fetch failed (likely corporate proxy/SSL), trying curl fallback...");
-
       // Fallback to curl for corporate networks
       try {
         const { stdout } = await execAsync(`curl -s -L "${this.UNPKG_URL}"`);
@@ -61,7 +55,6 @@ export class IconMetadataService {
         // Parse the JavaScript module output from curl
         return this.parseJavaScriptModule(stdout);
       } catch (curlError) {
-        console.error("[Vibe MCP] Curl fallback also failed:", curlError);
         throw fetchError; // Re-throw original fetch error
       }
     }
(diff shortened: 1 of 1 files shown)


packages/mcp/README.md
```diff
@@ -40,9 +40,17 @@ Create or update your VSCode MCP configuration file at `.vscode/mcp.json` in you
...
 ### Tools
 
+#### Component Tools
+
 - `get-vibe-component-metadata`: Returns metadata for a specific Vibe component (or all components if no name is provided).
 - `list-vibe-public-components`: Get a list of all public @vibe/core & @vibe/core/next components names.
 
+#### Icon Tools
+
+- `list-vibe-icons`: Get a list of all available Vibe icons from the @vibe/icons package with their descriptions, categories, and tags.
```
YossiSaadi, line 50: do we really want to expose it as structured data? and not just names?

packages/mcp/README.md
```diff
@@ -40,9 +40,17 @@ Create or update your VSCode MCP configuration file at `.vscode/mcp.json` in you
...
+- `get-vibe-icon-details`: Get detailed information about a specific Vibe icon including description, tags, category, and usage examples.
+- `search-vibe-icons`: Search and filter Vibe icons by text query, category, or tags. Supports filtering by categories ('Basic', 'Platform', 'View') and limiting results.
```
YossiSaadi, line 52: I'm not sure about the value of it? Agent would probably make an early call to `list-vibe-icons` and then call `get`

packages/mcp/src/server/tools/icon-metadata-service.ts
```diff
@@ -0,0 +1,153 @@
...
+      // Fallback to curl for corporate networks
+      try {
+        const { stdout } = await execAsync(`curl -s -L "${this.UNPKG_URL}"`);
+        return JSON.parse(stdout);
+      } catch (curlError) {
+        console.error("[Vibe MCP] Curl fallback also failed:", curlError);
+        throw fetchError; // Re-throw original fetch error
+      }
+    }
+  }
```
YossiSaadi, line 70: we're repeating ourself (in core tools as well). not for this PR (!) - for future - but can you make a task to make it a generic reusable?

packages/mcp/src/server/tools/icon-metadata-service.ts
```diff
@@ -0,0 +1,153 @@
...
+      console.error(`[Vibe MCP] Attempting to import icon metadata from workspace...`);
+
+      // For now, return a basic set of common icons if we can't read the full metadata
+      // This provides a fallback when the full metadata isn't available
+      return this.getFallbackIconMetadata();
+    } catch (error) {
+      console.error("[Vibe MCP] Error reading local icon metadata:", error);
+      throw error;
+    }
+  }
```
YossiSaadi, line 90: I think we should avoid this

YossiSaadi: that's only beneficial for us, so we don't need it, redundant logic

packages/mcp/src/server/tools/icon-metadata-service.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  private static getFallbackIconMetadata(): IconMetadata[] {
+    // Provide some common icons as fallback
+    return [
+      {
+        name: "Add",
+        file: "Add.svg",
```
YossiSaadi, line 97: let's remove this
it's better to not give data instead of giving partial data
we also have a fallback curl so the chances for failing are not high

packages/mcp/src/server/tools/list-vibe-icons.ts
```diff
@@ -0,0 +1,50 @@
...
+    try {
+      const allIcons = await IconMetadataService.getIconMetadata();
+
+      // Format the response to include relevant information
+      const formattedIcons = allIcons.map(icon => ({
+        name: icon.name,
+        description: icon.description,
+        category: icon.category || [],
+        tags: icon.tags.split(", ").map(tag => tag.trim())
+      }));
```
YossiSaadi, line 19: you can combine get-vibe-icon-details with list-vibe-icons if you're planning to return all the metadata in the "list".
you can add an optional param here of icon name - so if a user doesn't supply an icon name it would receive everything, otherwise receive specific supplied name

YossiSaadi (commented):
if you're planning to return full structure in `list`, then `search` is redundant as the Agent would be smart enough without the `search`.
also, I think combining `get` and `list` to a single `get` may be enough if `list` returns full structure

talkor (approved):

PR #2915, opened 2025-06-01
feat(icons): add NotificationChecked

talkor:
https://monday.monday.com/boards/3532714909/pulses/9256187147


packages/icons/plop/icon/index.js
```diff
@@ -44,7 +44,7 @@ export default {
             }
             console.log(lintStdout);
 
-            exec("yarn build:react-icons", (buildErr, buildStdout) => {
+            exec("yarn build", (buildErr, buildStdout) => {
```
rivka-ungar, line 47: Why was this removed?

rivka-ungar (approved):

packages/icons/plop/icon/index.js
talkor: I change it to run the entire build script since some other scripts were added to add, so it's more accurate to run the `build` script already

PR #2923, opened 2025-06-07
feat(tabs): added stretchedUnderline prop

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/8297159370


YossiSaadi (approved):
Excellent

PR #2925, opened 2025-06-08
feat: proof of concept for creating emptyState component from Figma using the new agentic workflow

YossiSaadi:
Almost perfect

image
image


YossiSaadi:
The new NumberField and AttentionBox were created with the new workflow - can be considered success so far

PR #2927, opened 2025-06-09
docs: steps story tipseen alignment

talkor:
https://monday.monday.com/boards/3532714909/pulses/5618073395

### DIFF
--- a/packages/core/src/components/Steps/__stories__/Steps.stories.tsx
+++ b/packages/core/src/components/Steps/__stories__/Steps.stories.tsx
@@ -159,33 +159,31 @@ export const StepsInsideATipseen = {
     }, []);
 
     return (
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Steps/__stories__/Steps.stories.tsx
```diff
@@ -159,33 +159,31 @@ export const StepsInsideATipseen = {
...
+            nextButtonProps={{
+              kind: "primary",
+              size: "small",
+              onClick: stepNext
+            }}
+            onFinish={() => {}}
+          />
+        }
+      >
+        <div className="monday-storybook-steps_container" />
```
rivka-ungar, line 185: Can this class be removed? I don't see that it has any styles and it is confusing for users who copy the code snippets

PR #2933, opened 2025-06-17
refactor: tokens

talkor:
https://monday.monday.com/boards/3532714909/pulses/4640358445


packages/core/src/components/Box/_utilities.scss
```diff
@@ -0,0 +1,263 @@
+@use "sass:map";
+@use "sass:string";
+@use "sass:list";
+@import "~monday-ui-style/dist/functions";
```
YossiSaadi, line 4: not sure I understand the need for this file

packages/style/src/files/tokens.json
```diff
@@ -0,0 +1,1138 @@
+{
+  "light-app-theme": {
```
YossiSaadi, line 2: this file should probably be available only at build-time, no?

YossiSaadi (approved):

packages/style/src/files/tokens.json
talkor: I thought about it but decided it's worth it have it in git so that if tokens are changed accidentally it will be visible as a change, but I'm not sure. WDYT?

packages/core/src/components/Box/_utilities.scss
talkor: It used some functions that were present in the style library only for the purpose of the Box so I moved it here since it's not a generic util

packages/style/src/files/tokens.json
YossiSaadi: you can have a test like `exports` in @vibe/core
but we can do it later, that's a non-blocker for me

YossiSaadi (approved):

packages/style/src/files/tokens.json
talkor: Good idea actually!

PR #2934, opened 2025-06-17
feat: Add steps component to testkit

hanan-monday:
https://monday.monday.com/boards/5374157346/pulses/9396388628

### DIFF
--- a/packages/testkit/components/Steps.ts
+++ b/packages/testkit/components/Steps.ts
@@ -1,0 +1,287 @@
+import { test, Locator, Page } from "@playwright/test";
+import { BaseElement } from "./BaseElement";
+import { Button } from "./Button";
+
+/**
+ * Class representing a Steps element.
+ * Extends the BaseElement class.
+ * Provides methods to interact with all Steps component functionality including navigation buttons,
+ * step indicators, and step content.
+ */
+export class Steps extends BaseElement {
+  private backButton: Button;
+  private nextButton: Button;
+  private finishButton: Button;
+
+  /**
+   * Create a Steps element.
+   * @param {Page} page - The Playwright page object.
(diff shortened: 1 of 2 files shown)


packages/testkit/components/Steps.ts
```diff
@@ -0,0 +1,263 @@
...
+   * @returns {Promise<Locator[]>} An array of step dot locators.
+   */
+  async getStepDots(): Promise<Locator[]> {
+    let stepDots: Locator[] = [];
+    await test.step(`Get all step dots in ${this.elementReportName}`, async () => {
+      const dotsContainer = this.locator.locator('[role="group"]');
+      stepDots = await dotsContainer.locator("button").all();
+    });
+    return stepDots;
+  }
```
or-monday: Do you intend to use this function in tests? If not, I think it should be private.

or-monday:
```suggestion
  /**
   * Get all step dots (for gallery type steps).
   * @returns {Promise<Locator[]>} An array of step dot locators.
   */
  private async getStepDots(): Promise<Locator[]> {
    let stepDots: Locator[] = [];
    await test.step(`Get all step dots in ${this.elementReportName}`, async () => {
      const dotsContainer = this.locator.locator('[role="group"]');
      stepDots = await dotsContainer.locator("button").all();
    });
    return stepDots;
  }
```

or-monday (commented):
@hanan-monday added a small comment to consider, but other than that it looks great.

packages/testkit/components/Steps.ts
hanan-monday: You are right.Thanks

YossiSaadi (approved):

PR #2939, opened 2025-06-22
docs(Link): update story structure and typings for improved clarity

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/9426150162


packages/core/src/components/Link/__stories__/Link.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import React from "react";
-import Link from "../Link";
+import { Meta, StoryObj } from "@storybook/react";
+import Link, { LinkProps } from "../Link";
 import { createStoryMetaSettingsDecorator } from "../../../storybook";
 import { ExternalPage, Info, Link as IconLink } from "@vibe/icons";
-import { createComponentTemplate } from "vibe-storybook-components";
 import "./Link.stories.scss";
```
rivka-ungar: Any chance we can remove the styles and make them inline?

rivka-ungar: Actually seems like the styles are not in use, so I guess the file can just be removed?

YossiSaadi: Good catch! Removed it

PR #2952, opened 2025-06-25
ci: force inner tasks to report "skip", to prevent ci stuck on "waiting for status..."

YossiSaadi:
This should prevent this behavior:
image

Where if we're changing something in the root, without changes to "packages" folder, the CI is stuck and we need to force merge

https://monday.monday.com/boards/3532714909/pulses/9449860055

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,12 +5,18 @@ name: Test & Lint
 
 on:
   workflow_call:
+    inputs:
+      has_changes:
+        description: Whether there are any changes in the monorepo.
+        type: string
+        required: true
     secrets:
       npm_token:
         required: true
 
 jobs:
   checks:
+    if: ${{ inputs.has_changes == 'true' }}
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -41,6 +47,8 @@ jobs:
         run: yarn lerna run ${{ matrix.command }} $SINCE_FLAG
 
   e2e:
+    name: E2E Tests
```
talkor, line 51: Did you notice the alert?

talkor (approved):
Finally! Yes!

.github/workflows/test.yml
YossiSaadi: which one 😟

talkor: Screenshot 2025-06-25 at 15 37 34

YossiSaadi: why it only showed it now 🤔 
I'll take a look
we anyway don't allow external users to run the workflows themselves, only we can approve the run

PR #2966, opened 2025-07-05
docs(icons): cleanup stories

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/9426150102


packages/core/src/components/Icon/__stories__/Icon.stories.tsx
```diff
@@ -84,7 +86,42 @@ export const Color = {
 };
 
 export const IconsListStory = {
-  render: IconsList.bind({}),
-  name: "Icons List",
```
YossiSaadi, line 88: maybe leave this `name` attribute (or just change the `const` name to IconsList`), so it won't be shown as "Icons List Story" in the docs page

YossiSaadi (approved):

PR #2981, opened 2025-07-13
docs(Flex): refactor stories and styles for improved layout and readability

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/9426149892

### DIFF
--- a/packages/core/src/components/Flex/__stories__/Flex.stories.module.scss
+++ b/packages/core/src/components/Flex/__stories__/Flex.stories.module.scss
@@ -1,9 +1,3 @@
 .flex-chip {
   margin: 0;
 }
-
-.story-container {
-  & > * {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Flex/__stories__/Flex.stories.module.scss
```diff
@@ -1,9 +1,3 @@
 .flex-chip {
   margin: 0;
```
rivka-ungar: This can be removed as well, since we have a new prop `noMargin` for the chip

YossiSaadi: niceee wasn't aware, thanks!!

rivka-ungar (approved):

PR #2991, opened 2025-07-20
feat(stylelint): add new stylelint rule for new spacing tokens system and update configuration

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/9628470815

### DIFF
--- a/packages/style/package.json
+++ b/packages/style/package.json
@@ -18,6 +18,16 @@
   "author": "monday.com",
   "license": "MIT",
   "main": "dist/index.css",
+  "exports": {
+    ".": "./dist/index.css",
+    "./dist/*": "./dist/*",
+    "./src/*": "./src/*",
+    "./stylelint-config/*": "./stylelint-config/*",
+    "./stylelint": {
+      "import": "./stylelint-config/index.js",
+      "default": "./stylelint-config/index.js"
+    }
+  },
   "files": [
     "src/",
     "dist/",
--- a/packages/style/stylelint-config/index.js
+++ b/packages/style/stylelint-config/index.js
@@ -1,7 +1,8 @@
 module.exports = {
   customSyntax: require.resolve("postcss-scss"),
-  plugins: ["./rules/use-defined-css-var-when-available/index.js"],
+  plugins: ["./rules/use-defined-css-var-when-available/index.js", "./rules/use-new-spacing-tokens/index.js"],
   rules: {
-    "monday-ui-style/use-defined-css-var-when-available": true
+    "monday-ui-style/use-defined-css-var-when-available": true,
+    "monday-ui-style/use-new-spacing-tokens": [true, { severity: "warning" }]
   }
 };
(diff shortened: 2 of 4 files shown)


YossiSaadi:
I've added to the rule an autofix with `stylelint --fix`. that means that in Vibe4 we can upgrade to the new spacing system in one cli command 🎉

packages/style/package.json
```diff
@@ -18,6 +18,16 @@
   "author": "monday.com",
   "license": "MIT",
   "main": "dist/index.css",
+  "exports": {
+    ".": "./dist/index.css",
+    "./dist/*": "./dist/*",
+    "./src/*": "./src/*",
```
talkor: TBH if we need to add these to not break it, I prefer not adding exports at al (and adding the "right" exports in the next major)

packages/core/.stylelintrc.json
```diff
@@ -1,6 +1,7 @@
 {
-  "extends": "../../.stylelintrc.json",
+  "extends": ["../../.stylelintrc.json", "monday-ui-style/stylelint"],
   "rules": {
+    "monday-ui-style/use-defined-css-var-when-available": null,
```
talkor, line 4: Why?

packages/style/stylelint-config/index.js
```diff
@@ -1,7 +1,8 @@
 module.exports = {
   customSyntax: require.resolve("postcss-scss"),
-  plugins: ["./rules/use-defined-css-var-when-available/index.js"],
+  plugins: ["./rules/use-defined-css-var-when-available/index.js", "./rules/use-new-spacing-tokens/index.js"],
   rules: {
-    "monday-ui-style/use-defined-css-var-when-available": true
+    "monday-ui-style/use-defined-css-var-when-available": true,
+    "monday-ui-style/use-new-spacing-tokens": [true, { severity: "warning" }]
```
talkor: @rivka-ungar and I came to the conclusion that it's best not to mix it up right now. WDYT? Let's talk about it. The rule is perfect but I'm not sure we'd want to enforce it in this major (even with a warning)

packages/style/stylelint-config/rules/use-new-spacing-tokens/index.js
talkor, line 1: That's awesome!

talkor (commented):
Looks perfect but I think we should discuss the enforcement for this major, plus the exports

packages/style/package.json
```diff
@@ -18,6 +18,16 @@
...
+    "./stylelint-config/*": "./stylelint-config/*",
+    "./stylelint": {
```
talkor: Do we need to bring back stylelint as a dep?

YossiSaadi: it needs to be a dev dep

talkor: Yeah actually it's already a peer dep

packages/core/.stylelintrc.json
YossiSaadi: I've turned it back on at the next PR

packages/style/package.json
YossiSaadi: removed

packages/style/stylelint-config/index.js
YossiSaadi: Not enabled by default now 💪🏼

PR #3006, opened 2025-07-23
docs(Tooltip): cleanup stories

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/9426160014


packages/core/src/components/Tooltip/__stories__/Tooltip.stories.tsx
```diff
@@ -21,24 +20,34 @@ export default {
...
+              mainAxis: false
+            }
+          },
+          {
+            name: "flip",
+            options: {
+              // @ts-ignore
+              fallbackPlacements: []
+            }
+          }
```
talkor, line 44: Are all of these necessary? Maybe some not, like we had in Dialog

packages/core/src/components/Tooltip/__stories__/Tooltip.stories.tsx
```diff
@@ -85,31 +110,49 @@ export const TooltipWithImage = () => (
...
           </Tooltip>
         </div>
-        <div>
+        <div style={{ padding: "0 32px 8px 32px", margin: "0 32px" }}>
           <Tooltip modifiers={modifiers} content="Left" position="right" shouldShowOnMount open>
             <div />
           </Tooltip>
         </div>
-        <div>
+        <div style={{ padding: "0 64px 8px 64px", margin: "0 64px" }}>
```
talkor, line 150: I think we should think of a simpler layout to prevent such cases (not right now)

talkor (approved):

packages/core/src/components/Tooltip/__stories__/Tooltip.stories.tsx
rivka-ungar: Unfortunately Yes :(

packages/core/src/components/Tooltip/__stories__/Tooltip.stories.tsx
rivka-ungar: 100%, maybe as part of removing the flex :)

PR #3025, opened 2025-08-03
fix: add check for ResizeObserver support

talkor:
https://monday.monday.com/boards/3532714909/pulses/9712991279

### DIFF
--- a/packages/core/src/components/Modal/layouts/useLayoutScrolledContent.ts
+++ b/packages/core/src/components/Modal/layouts/useLayoutScrolledContent.ts
@@ -22,6 +22,10 @@ const useLayoutScrolledContent = () => {
   }, [checkScroll]);
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


YossiSaadi (commented):
is it for SSR? if so, you can check `isServer` from `ssr-utils`

YossiSaadi:
also see Qodo's comment

talkor:
> is it for SSR? if so, you can check `isServer` from `ssr-utils`

No, for tests in jest-dom env where ResizeObserver is not defined. 

> also see Qodo's comment

Good call! But I've checked and React internally checks typeof returnValue === 'function' before calling cleanup, and when you return early (returning undefined), React recognizes this and does not attempt cleanup, so no memory leak. Plus if I'm not mistaken ResizeObserver is supported in all of our browserslist so we're good

YossiSaadi (approved):

PR #3029, opened 2025-08-03
chore: enforce importing types as "type" for Vite-reasons

YossiSaadi:
https://monday.monday.com/boards/3532714909/pulses/9732997220


talkor (approved):
👏🏼 💯

PR #3033, opened 2025-08-05
ci: add a bundle check

talkor:


.github/workflows/bundle-size.yml
YossiSaadi, line 1: why won't we add it to the "test" workflow? it won't automatically be a blocker, so it's ok

you can add `if: ${{ inputs.has_changes == 'true' && anotherCheckHereThatChecks if "event.pull_request.paths" includes the paths you said }}`

Let's check that again in the future if we're not doing it now

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,90 @@
...
+          fetch-depth: 0
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
```
YossiSaadi:
```suggestion
      - name: Run Setup
        uses: ./.github/actions/setup
```

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,90 @@
...
+
+  bundle-size:
+    runs-on: ubuntu-latest
+    needs: build
+
+    steps:
+      - name: Checkout PR
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
YossiSaadi:
```suggestion
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0
```

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,90 @@
...
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Download build artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: ci-builds-${{ github.run_id }}-${{ github.run_attempt }}
+          path: packages/
```
YossiSaadi:
```suggestion
      - uses: ./.github/actions/download-builds
```

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Run size-limit on PR
+        run: |
+          yarn install --frozen-lockfile
+          node scripts/bundle-check/generate-size-limit-config.js
+          ./node_modules/.bin/size-limit --json > pr.json
```
YossiSaadi:
```suggestion
      - name: Run size-limit on PR
        run: |
          node scripts/bundle-check/generate-size-limit-config.js
          ./node_modules/.bin/size-limit --json > pr.json
```

can't you in `generate-size-limit-config.js` make sure that it won't be inside `node_modules`? that it would make the `pr.json` from the scripts automatically?

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,90 @@
...
+            echo ".size-limit.js found on base commit. Running size-limit."
+            yarn install --frozen-lockfile
+            yarn build
+            yarn add --dev -W size-limit @size-limit/preset-small-lib
+            ./node_modules/.bin/size-limit --json > base.json
+          else
+            echo ".size-limit.js not found on base commit. Creating empty base.json."
+            echo "[]" > base.json
+          fi
+        continue-on-error: true
```
YossiSaadi: you can do
```
        env:
          BASE_SHA: ${{ github.event.pull_request.base.sha }}
          HEAD_SHA: ${{ github.sha }}
```

to get the same shas that you're calculating inside, you get them from github's context

also, you can just use the "actions/checkout" instead of doing `git checkout --force $PR_HEAD_SHA" EXIT`
just add a step of "now I'm checkouting to commit $BASE_SHA" like this:
```
- uses: actions/checkout@v4
   env:
    PR_HEAD_SHA: ${{ github.event.pull_request.base.sha }}
   with:
     ref: $PR_HEAD_SHA
```

and then do again the run setup, run build and whatever you need later

YossiSaadi: or just of course straight:
```
- uses: actions/checkout@v4
  with:
    ref: ${{ github.event.pull_request.base.sha }}
```

should only make sure if it is `base.sha` or `head.sha`, I don't remember tbh

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Compare sizes
+        run: node scripts/bundle-check/compare-bundles.js
+
+      - name: Post comment
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: bundle-size
+          path: bundle-sizes.md
```
YossiSaadi: Cool!

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,90 @@
...
+            yarn add --dev -W size-limit @size-limit/preset-small-lib
+            ./node_modules/.bin/size-limit --json > base.json
+          else
+            echo ".size-limit.js not found on base commit. Creating empty base.json."
+            echo "[]" > base.json
+          fi
+        continue-on-error: true
+
+      - name: Compare sizes
+        run: node scripts/bundle-check/compare-bundles.js
```
YossiSaadi, line 55: maybe worth passing them here to the script, and make the script accept two arguments for base and PR.

(future)

.gitignore
```diff
@@ -14,4 +14,10 @@ coverage
...
 packages/testkit/test-results
-packages/testkit/reports
\ No newline at end of file
+packages/testkit/reports
+
+# Bundle size analysis files
+base.json
+pr.json
+bundle-sizes.md
+.size-limit.js
```
YossiSaadi: maybe put them in a folder in the script and ignore the entire folder?
/bundle-comparison or something?

just in case in the future if you add files to the script they won't sneak into the bundled package in npm

YossiSaadi, line 19:
```suggestion
size-limit.js
```

?

package.json
```diff
@@ -8,11 +8,12 @@
     "lint": "lerna run lint",
     "test": "lerna run test",
     "storybook": "lerna run storybook --parallel --stream",
-    "postinstall": "lerna run build --scope={vibe-storybook-components,@vibe/icons}",
-    "size": "lerna run size --scope=@vibe/core"
+    "postinstall": "lerna run build --scope={vibe-storybook-components,@vibe/icons}"
   },
   "devDependencies": {
+    "@size-limit/preset-small-lib": "^11.2.0",
```
YossiSaadi, line 14: in the root? why not in @vibe/core?

package.json
```diff
@@ -8,11 +8,12 @@
...
     "lerna": "^8.1.2",
+    "size-limit": "^11.2.0",
```
YossiSaadi, line 16: same

.github/workflows/bundle-size.yml
talkor: I don't mind adding it to test, I just didn't want that workflow to be bloated with something that isn't specifically a "test" but a check. WDYT?

.github/workflows/bundle-size.yml
talkor: Not sure I understand

.github/workflows/bundle-size.yml
talkor: That's actually seem much nicer, I'll use that, thanks!

.gitignore
talkor: Good idea!

talkor: That's actually intended, that's the format (though it can be customized but it won't be relevant if I ignore the entire folder

package.json
talkor: Since we're using it in a workflow only and not in the package specifically, and it can be used in further packages as well if we decide

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,79 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+          cache: "yarn"
```
YossiSaadi: you don't need it, it exists already on "./.github/actions/setup"

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,79 @@
+name: Bundle Size Analysis
+
+on:
+  pull_request:
+    paths:
+      - "packages/core/src/components/**"
+      - "packages/core/package.json"
+
+permissions:
+  pull-requests: write
```
YossiSaadi, line 10: should work without it

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,79 @@
...
+          trap "echo 'Checking out back to PR commit'; git checkout --force $PR_HEAD_SHA" EXIT
+
+          BASE=$(git merge-base origin/master HEAD)
+          echo "Base commit: $BASE"
+          git checkout $BASE
+          yarn install --frozen-lockfile
+          yarn build
+          yarn add --dev -W size-limit @size-limit/preset-small-lib
+          ./node_modules/.bin/size-limit --json > scripts/bundle-check/reports/base.json
+        continue-on-error: true
```
YossiSaadi, line 52: the additional checkout didn't work?

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,79 @@
...
+
+      - name: Run Setup
+        uses: ./.github/actions/setup
+
+      - name: Download build artifacts
+        uses: ./.github/actions/download-builds
+
+      - name: Run size-limit on PR
+        run: |
+          yarn install --frozen-lockfile
```
YossiSaadi: probably not needed

YossiSaadi: already happens on `setup`

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,79 @@
...
+          node scripts/bundle-check/generate-size-limit-config.js
+          mkdir -p scripts/bundle-check/reports
+          ./node_modules/.bin/size-limit --json > "scripts/bundle-check/reports/pr.json"
+
+      - name: Run size-limit on base commit
+        run: |
+          PR_HEAD_SHA=$(git rev-parse HEAD)
+          trap "echo 'Checking out back to PR commit'; git checkout --force $PR_HEAD_SHA" EXIT
+
+          BASE=$(git merge-base origin/master HEAD)
```
YossiSaadi, line 45: can also be: `github.event.pull_request.base.sha`

.github/workflows/bundle-size.yml
YossiSaadi, line 52: I believe this entire step can be simplified using GitHub's Context, and separate the checkout to another step, but if you tried and didn't work let's check that out later

package.json
YossiSaadi: you can by the way install it only inside the workflow itself if it only gets used in the CI

.github/workflows/bundle-size.yml
YossiSaadi: but don't spend time on it, we can test it later

YossiSaadi (approved):
just remove the `yarn install` and `setup-node`, they are redundant.

edit: this would save you at least 30secs

.github/workflows/bundle-size.yml
talkor: It's needed for the sticky comments: https://github.com/marocchino/sticky-pull-request-comment?tab=readme-ov-file#basic

.github/workflows/bundle-size.yml
talkor: You're right, removed

.github/workflows/bundle-size.yml
talkor: Tried it but I need the trap here, otherwise I'm having some issues with branching

.github/workflows/bundle-size.yml
talkor: Nope :/

.github/workflows/bundle-size.yml
YossiSaadi: hmm got ya

PR #3068, opened 2025-08-25
feat(Tabs): add ariaControls

rivka-ungar:
https://monday.monday.com/boards/3532714909/views/80492480/pulses/9806634098


packages/core/src/components/Tabs/TabsContext/TabsContext.tsx
```diff
@@ -44,6 +53,19 @@ const TabsContext: FC<TabsContextProps> = forwardRef(
       [activeTabIdState]
     );
 
+    // Collect TabPanel ids for aria-controls relationship
+    const tabPanelIds = useMemo(() => {
+      const ids: string[] = [];
+      React.Children.forEach(children, (child: TabsChild) => {
```
talkor, line 59: I think it's deprecated in React 19?

packages/core/src/components/Tabs/TabsContext/TabsContext.tsx
```diff
@@ -44,6 +53,19 @@ const TabsContext: FC<TabsContextProps> = forwardRef(
...
+        if (child.type.isTabPanels) {
```
talkor, line 60: Is this ok? Where is it defined?

rivka-ungar: It is defined [here](https://github.com/mondaycom/vibe/blob/48f94e690d0c9944040a5259bc438941dbd35074/packages/core/src/components/Tabs/TabPanels/TabPanels.tsx#L68), and used [here](https://github.com/mondaycom/vibe/blob/48f94e690d0c9944040a5259bc438941dbd35074/packages/core/src/components/Tabs/TabsContext/TabsContext.tsx#L63) as well

packages/core/src/components/Tabs/TabsContext/TabsContext.tsx
rivka-ungar: It is the same logic used [here](https://github.com/mondaycom/vibe/blob/48f94e690d0c9944040a5259bc438941dbd35074/packages/core/src/components/Tabs/TabsContext/TabsContext.tsx#L54), so i guess if we will need to change it it will be in both places. Do you think it is ok to keep for now or should I change it?

talkor: Yes if it's already present in other places let's keep it this way for now and use non-deprecated api all at once

talkor (approved):

PR #3092, opened 2025-09-07
docs: add Monday Vibe banner

talkor:
https://monday.monday.com/boards/3532714909/pulses/9958580790


packages/core/src/storybook/stand-alone-documentaion/welcome/banner/tracking.ts
```diff
@@ -0,0 +1,37 @@
+export const BANNER_EVENTS = {
+  MONDAY_VIBE_BANNER_VIEWED: "monday_vibe_banner_viewed",
```
rivka-ungar: Are these event names unique? How can we be sure they don't exist already?

talkor: I checked in BB

PR #3393, opened 2026-06-02
feat(mcp): auto-detect Vibe version for correct metadata

talkor:
## Summary
- Adds a version detection module that identifies whether the consumer uses Vibe 3 or 4 by checking `node_modules/@vibe/core/package.json` (walking up directories for monorepo support), then falling back to `package.json` dependency ranges
- Updates `MetadataService` to fetch from `@vibe/core@3` or `@vibe/core@4` on unpkg instead of `@latest` — fixing the bug where v3 consumers received v4 metadata
- Defaults to v4 when no version is detected

## Test plan
- [x] Unit tests for version detection (node_modules lookup, package.json fallback, devDependencies, default)
- [x] Unit tests for semver range parsing edge cases (`^3.0.0`, `~4.1.0`, `>=4.0.0`, `*`, `latest`, `workspace:*`)
- [ ] Manual: verify MCP server fetches v3 metadata when run in a project with `@vibe/core@3` installed
- [ ] Manual: verify MCP server fetches v4 metadata when run in a project with `@vibe/core@4` installed


rivka-ungar (commented):
Looks great, What about the examples?

talkor:
> Looks great, What about the examples?

Added a second commit that moves the example and accessibility extraction scripts from `@vibe/mcp` into `@vibe/core`'s build pipeline.

This means `@vibe/core` will now publish component examples and accessibility data in its dist (`dist/metadata/examples/` and `dist/metadata/accessibility/`), making them fetchable from unpkg at runtime — just like `metadata.json` already is today. The old `dist/metadata.json` path is preserved as a copy for backwards compat with existing MCP versions.

This is a prerequisite for making the MCP tools fetch all data at runtime (instead of bundling it), which enables the single-package approach that supports both Vibe 3 and Vibe 4 consumers.

rivka-ungar (approved):
Awesome!

PR #3413, opened 2026-07-02
fix(Dropdown): accessibility fixes for searchable single & multi select

rivka-ungar:
Clean PR (based on `master`) with the Dropdown accessibility fixes only. The Storybook pages / docs for these are intentionally **not** included here — they're kept in the other PR.

## Single searchable
- Selected value lives **inside the input** (exposed to assistive tech) instead of a visual overlay; filter resets on close; focus stays on the input after selecting.
- `aria-haspopup="dialog"` on the combobox input; `helperText` linked via `aria-describedby` (needs an `id`).

## Searchable (single + multi)
- **Space** selects/toggles the highlighted option (when `aria-activedescendant` is set) instead of inserting a literal space.
- **Chevron** gets a real accessible name: `aria-labelledby` → the visible `label`, else an `aria-label` string — never the listbox.

## Multi select
- Each selected chip renders with a labelled remove (**×**) button; chips wrapper exposed as `role="group"` `aria-label="selected items"`.
- Current selection announced via a visually hidden element referenced with `aria-describedby`.
- `interactiveChips`: keyboard-navigable chips (Arrow keys to move, Backspace/Delete to remove), selection kept out of the input value so type-to-search stays available.
- **+N overflow dialog**: opens on Enter, focus moves to the first chip, Esc returns focus to the +N counter.
- Correct `aria-selected` on options in multi-select.

## Supporting change
- `Flex` gains a `role` prop (used by the chips group).

## Notes
- Nothing is gated behind a new prop — these apply to `searchable` / `multi` / the existing `interactiveChips` as noted.
- 89 Dropdown unit tests pass; lint/types clean. The `+N` overflow focus behavior needs manual/Storybook verification (jsdom can't trigger overflow).


packages/core/src/components/Dropdown/components/MultiSelectedValues/MultiSelectedValues.tsx
```diff
@@ -73,17 +108,23 @@ function MultiSelectedValues<Item extends BaseItemData<Record<string, unknown>>>
...
     return selectedItems.map((item, index) => {
       const isVisible = index < visibleCount;
+      const extraProps = isVisible && getChipContainerProps ? getChipContainerProps(item, index) : {};
+      const { ref: extraRef, ...extraAttrs } = extraProps;
 
       return (
         <div
           key={`dropdown-chip-visible-${item.value}`}
-          ref={itemRefs[index]}
+          ref={el => {
```
talkor, line 116: Maybe memo?

packages/core/src/components/Dropdown/components/MultiSelectedValues/MultiSelectedValues.tsx
```diff
@@ -131,13 +178,18 @@ function MultiSelectedValues<Item extends BaseItemData<Record<string, unknown>>>
             <Dialog
               content={dialogContent}
               showTrigger={["click", "enter"]}
-              hideTrigger={["clickoutside", "enter"]}
+              // "enter" is intentionally NOT a hide trigger: Enter keydown opens and the Enter keyup
+              // would otherwise immediately toggle it shut. Close via Esc or click-outside instead.
+              hideTrigger={["clickoutside"]}
```
talkor, line 182: Is it breaking?

packages/core/src/components/Dropdown/components/Trigger/DropdownInput.tsx
```diff
@@ -17,44 +37,71 @@ const DropdownInput = ({ inputSize, fullWidth }: { inputSize?: "small" | "medium
...
-  const multipleSelectionDropdownProps = getDropdownProps ? getDropdownProps({ preventKeyAction: isOpen }) : {};
+  // interactiveChips: menu is always open, so isOpen would permanently suppress Backspace chip-nav.
+  // Instead suppress only when the input has text (Backspace should delete chars, not navigate chips).
+  const preventKeyAction = interactiveChips ? !!(inputValue && inputValue.length > 0) : isOpen;
+  const multipleSelectionDropdownProps = getDropdownProps ? getDropdownProps({ preventKeyAction }) : {};
+
+  // Stable id for the visually hidden element that announces the current selection.
+  // Only needed for multi-select chips; single-select already keeps the value inside the input.
+  const selectedValueId = useRef(`dropdown-selected-${Math.random().toString(36).slice(2, 9)}`).current;
+  const selectedValueText = useMemo(() => (multi ? getSelectedValueText(selectedItems) : ""), [multi, selectedItems]);
```
talkor: Better move these to the top to avoid react hook violations

packages/core/src/components/Dropdown/components/Trigger/Trigger.module.scss
```diff
@@ -105,3 +95,15 @@
...
 }
+
+.visuallyHidden {
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
```
talkor, line 116: Needed? nitpick...

packages/core/src/components/Dropdown/Dropdown.types.ts
```diff
@@ -20,10 +20,20 @@ interface MultiSelectSpecifics<Item extends BaseItemData<Record<string, unknown>
    * Callback fired when an option is removed in multi-select mode. Only available when multi is true.
    */
   onOptionRemove?: (option: Item) => void;
+  /**
+   * If true, chips are always visible and support keyboard navigation: pressing Backspace or Left arrow
+   * from the input moves focus to the last chip; Left/Right navigates between chips; Delete/Backspace
+   * removes the focused chip. Only applies when searchable=true.
+   */
+  interactiveChips?: boolean;
```
talkor, line 28: Shouldn't this be always true?

packages/core/src/components/Dropdown/Dropdown.types.ts
```diff
@@ -59,6 +69,13 @@ interface SingleSelectSpecifics<Item extends BaseItemData<Record<string, unknown
    * The function to call to render the selected value on single select mode.
    */
   valueRenderer?: (option: Item) => React.ReactNode;
+  /**
```
talkor, line 72: Shouldn't this always be true?

talkor (approved):

packages/core/src/components/Dropdown/components/MultiSelectedValues/MultiSelectedValues.tsx
rivka-ungar: Its fixing wrong behavior, so don't think it is considered breaking

packages/core/src/components/Dropdown/Dropdown.types.ts
rivka-ungar: Not yet :( design didn't approve yet, and we need a solution, that is why we are releasing it with opt in, and not by default for all

PR #3417, opened 2026-07-12
perf(ColorPickerItemComponent): skip Tooltip render when showColorNameTooltip is false

rivka-ungar:
## Motivation

`ColorPickerItemComponent` wraps every color swatch in `<Tooltip content={tooltipContent}>` regardless of whether `tooltipContent` has a value. When `showColorNameTooltip` is `false` (the default), `tooltipContent` is `undefined` for every swatch, yet each one still mounts `Tooltip` → `Dialog` with its ~49 hooks.

A ColorPicker grid typically renders 20–60 swatches, so in the default configuration this mounts up to 60 needless Dialog trees on every render.

This is the same guard applied to `Tooltip` itself in PR #3416 — extend the pattern down one level.

## Changes

- Extract the `<li>` element into a `liElement` variable.
- When `tooltipContent` is falsy, return `liElement` directly (no Tooltip/Dialog mount).
- When `tooltipContent` is truthy, return `<Tooltip content={tooltipContent}>{liElement}</Tooltip>` as before.
- No changes to props, types, styles, or external API.

## Safety

- Behavior is **identical** when `tooltipContent` is provided — the Tooltip wraps the swatch exactly as before.
- The `Tooltip` import is retained; it is still used when needed.
- All existing ColorPicker tests pass (10/10).

## Test plan

- [x] `yarn workspace @vibe/core test ColorPicker` — 10/10 tests pass
- [ ] Storybook: verify color name tooltips still appear when `showColorNameTooltip={true}`
- [ ] Storybook: verify no visual regression when `showColorNameTooltip={false}` (default)

### DIFF
--- a/packages/core/src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.tsx
+++ b/packages/core/src/components/ColorPicker/components/ColorPickerItemComponent/ColorPickerItemComponent.tsx
@@ -116,47 +116,52 @@ const ColorPickerItemComponent = forwardRef(
 
     const shouldRenderIcon = isSelected || ColorIndicatorIcon;
     const colorIndicatorWrapperStyle = shouldRenderIndicatorWithoutBackground ? { color: colorAsStyle } : {};
-    return (
-      <Tooltip content={tooltipContent}>
-        <li
-          id={id}
-          role="option"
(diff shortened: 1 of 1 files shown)


rivka-ungar:

PR #3422, opened 2026-07-12
perf(SliderThumb): skip Tooltip render when showValue is true

rivka-ungar:
## Motivation

When `showValue=true`, `SliderThumb` renders the current value as an inline label inside the thumb element. In this mode, the tooltip content is `null` — but `<Tooltip>` (and the underlying `<Dialog>`) was still mounted, causing 49+ hooks to run on every render for no benefit.

This mirrors the guard added to `Tooltip` itself in PR #3416.

## Changes

- Extract `tooltipContent = showValue ? null : valueText`
- Extract the thumb `<div>` into a local `thumb` variable
- When `tooltipContent` is falsy, return `thumb` directly (no `<Tooltip>` mounted)
- When `tooltipContent` is truthy, wrap `{thumb}` with `<Tooltip>` as before, preserving the `open={active === index || dragging === index}` controlled-open behavior

**File changed**: `packages/core/src/components/Slider/SliderBase/SliderThumb.tsx`

## Safety

- The `open` prop (active/dragging state → tooltip visible) only applies when `tooltipContent` is truthy, i.e. `showValue=false`. In that scenario nothing has changed.
- When `showValue=true` the tooltip was already invisible (null content); we now avoid mounting Dialog entirely.
- All 70 Slider tests pass.

## Test Plan

- [ ] `yarn workspace @vibe/core test Slider` — 70 tests, all passing
- [ ] Verify Slider with `showValue=false`: tooltip still appears on hover/drag showing the value
- [ ] Verify Slider with `showValue=true`: inline label renders correctly, no tooltip mounted

### DIFF
--- a/packages/core/src/components/Slider/SliderBase/SliderThumb.tsx
+++ b/packages/core/src/components/Slider/SliderBase/SliderThumb.tsx
@@ -85,55 +85,65 @@ const SliderThumb: FC<SliderThumbProps> = ({ className, index = 0, onMove = NOOP
     }
   }, [focused, index]);
 
+  const tooltipContent = showValue ? null : valueText;
+
+  const thumb = (
+    <div
+      aria-label={ariaLabel}
+      aria-labelledby={ariaLabelledby}
+      aria-valuemax={max}
+      aria-valuemin={min}
+      aria-valuenow={value}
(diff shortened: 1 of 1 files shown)


rivka-ungar:

PR #3426, opened 2026-07-12
perf(Tab): skip Tooltip render when no tooltip content

rivka-ungar:
## Motivation

Every `Tab` renders unconditionally inside a `<Tooltip>`, even though the default for `tooltipProps` is `{} as TooltipProps`, meaning `tooltipProps.content` is always `undefined` in the typical case.

`Tooltip` delegates to `Dialog` which mounts **49+ hooks** on every render. A tab bar with 10 tabs therefore mounts ~490 unnecessary hook calls on every render cycle, even when no tooltip is needed.

This mirrors the same optimization made to `Tooltip` itself in PR #3416.

## Changes

- `packages/core/src/components/Tabs/Tab/Tab.tsx`: Extract the `<li>` into a local `tab` constant. Return it directly when `tooltipProps?.content` is falsy. Only wrap with `<Tooltip>` when content is truthy.

## Safety

- **No behavior change** when `tooltipProps.content` is provided — the `<Tooltip>` renders exactly as before with all spread props.
- **Zero visual change** for the 99 % of tabs that have no tooltip — the `<li>` is identical to what was previously rendered inside the pass-through Tooltip.
- All existing tests pass (15/15 Tab tests green).

## Test plan

- [x] `yarn workspace @vibe/core test Tab` — 95 tests pass across 12 test files
- [ ] Visually verify tab bar renders correctly in Storybook
- [ ] Verify tooltip still appears when `tooltipProps.content` is set on a Tab

### DIFF
--- a/packages/core/src/components/Tabs/Tab/Tab.tsx
+++ b/packages/core/src/components/Tabs/Tab/Tab.tsx
@@ -126,32 +126,36 @@ const Tab: FC<TabProps> = forwardRef(
       }
     }
 
-    return (
-      <Tooltip {...tooltipProps} content={tooltipProps.content}>
-        <li
-          ref={mergedRef}
-          key={id}
-          className={cx(styles.tabWrapper, className, {
-            [styles.active]: active,
-            [styles.disabled]: disabled,
-            [styles.tabFocusVisibleInset]: focus,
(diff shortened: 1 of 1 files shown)


rivka-ungar:

PR #3428, opened 2026-07-12
perf(ModalFooterBase): skip Tooltip render on buttons when no tooltip content

rivka-ungar:
## Motivation

`ModalFooterBase` defaults both `primaryButtonTooltipProps` and `secondaryButtonTooltipProps` to empty objects `{}`, so `tooltipProps.content` is always `undefined` unless a caller explicitly provides it. As a result, every modal footer unconditionally mounts a `<Tooltip>` around both buttons — and Tooltip internally mounts Dialog, which runs 49+ hooks — even when no tooltip text is ever shown.

This is the same issue addressed for the standalone `Tooltip` component in PR #3416 (which added an early-return guard when `content` is falsy).

## Changes

- **Primary button**: only renders `<Tooltip>` when `primaryButtonTooltipProps.content` is truthy; renders `<Button>` directly otherwise.
- **Secondary button**: same guard — only wraps with `<Tooltip>` when `secondaryButtonTooltipProps.content` is truthy.
- The redundant explicit `content={...tooltipProps.content}` spread (which was already included via `{...tooltipProps}`) is removed; the spread alone is sufficient.

## Safety

- Zero behaviour change for any caller that passes a non-empty `content` — the `<Tooltip>` renders exactly as before.
- Callers that never set `content` (the common case) no longer pay the cost of mounting Dialog.
- All 6 existing `ModalFooterBase` unit tests pass unchanged.

## Test plan

- [x] `ModalFooterBase` unit tests (6/6 green)
- [ ] Visual regression: render a modal footer without tooltip props — buttons behave identically
- [ ] Visual regression: render a modal footer with `primaryButtonTooltipProps={{ content: "Tooltip text" }}` — tooltip appears correctly on hover
- [ ] Verify that secondary button tooltip also works when `secondaryButtonTooltipProps.content` is set

### DIFF
--- a/packages/core/src/components/Modal/footers/ModalFooterBase/ModalFooterBase.tsx
+++ b/packages/core/src/components/Modal/footers/ModalFooterBase/ModalFooterBase.tsx
@@ -22,6 +22,13 @@ const ModalFooterBase = forwardRef(
       ...secondaryButtonProps
     } = secondaryButton || {};
 
+    const primaryButtonEl = <Button {...primaryButtonProps}>{primaryButtonText}</Button>;
+    const secondaryButtonEl = secondaryButton ? (
+      <Button {...secondaryButtonProps} kind="tertiary">
+        {secondaryButtonText}
+      </Button>
+    ) : null;
+
     return (
       <Flex
         ref={ref}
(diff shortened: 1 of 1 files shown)


rivka-ungar:

PR #3430, opened 2026-07-12
perf(DropdownBase): skip Tooltip render when no tooltip props

rivka-ungar:
## Motivation

Every `Dropdown` instance in the app was mounting a `<Tooltip>` (which internally mounts `Dialog`) even when no tooltip was configured. `Dialog` runs 49+ hooks on mount — a silent overhead hit that applied to **every** Dropdown regardless of whether it ever shows a tooltip.

This is the same pattern applied to `Tooltip` itself in PR #3416, where an early-return guard (`<>{children}</>`) was added when `content` is falsy.

## Change

**File**: `packages/core/src/components/Dropdown/components/DropdownBase/DropdownBase.tsx`

Before:
```tsx
<Tooltip {...tooltipProps} content={tooltipProps?.content}>
  {coreDropdownElement}
</Tooltip>
```

After:
```tsx
{tooltipProps?.content ? (
  <Tooltip {...tooltipProps}>{coreDropdownElement}</Tooltip>
) : (
  coreDropdownElement
)}
```

When `tooltipProps?.content` is falsy (the common case — no tooltip configured), the core dropdown element is rendered directly, skipping `Tooltip` and `Dialog` entirely.

When `tooltipProps?.content` is truthy, the existing `<Tooltip>` wrapping behavior is preserved unchanged.

## Safety

- No change in behavior when `tooltipProps.content` is provided
- No change to the DOM structure of the dropdown element itself
- The `Tooltip` import remains (used in the truthy branch)
- All existing Dropdown tests pass (90/90)

## Test plan

- [ ] `yarn workspace @vibe/core test` — all Dropdown tests pass
- [ ] Visual check in Storybook: Dropdown with and without `tooltipProps` renders correctly
- [ ] Confirm Dropdown with `tooltipProps.content` still shows tooltip on hover

### DIFF
--- a/packages/core/src/components/Dropdown/components/DropdownBase/DropdownBase.tsx
+++ b/packages/core/src/components/Dropdown/components/DropdownBase/DropdownBase.tsx
@@ -63,9 +63,11 @@ const DropdownBase = ({ dropdownRef, children }: DropdownBaseProps) => {
   return (
     <div dir={dir} className={styles.outerWrapper}>
       {label && <FieldLabel labelText={label} required={required} {...getLabelProps()} />}
-      <Tooltip {...tooltipProps} content={tooltipProps?.content}>
-        {coreDropdownElement}
-      </Tooltip>
+      {tooltipProps?.content ? (
+        <Tooltip {...tooltipProps}>{coreDropdownElement}</Tooltip>
+      ) : (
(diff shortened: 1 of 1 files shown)


rivka-ungar:

PR #3459, opened 2026-08-03
refactor(core): scope Toggle focus-ring selector to a marker class

orrgottlieb:
## What

Replaces the non-specific `.toggleInput:focus-visible ~ div` selector in the Toggle component with `~ .toggleFocusTarget`, a dedicated marker class applied to the drawn toggle `<div>`.

## Why

The old rule matched a bare `div` type selector under a general sibling combinator. It only worked because there happened to be exactly one `div` sibling inside the `<label>` — non-specific and fragile. This scopes the focus ring to an explicit target.

## Changes

- `Toggle.module.scss`: `& ~ div` → `& ~ .toggleFocusTarget`
- `Toggle.tsx`: pass the marker class to `MockToggle` via `className={cx(className, styles.toggleFocusTarget)}`
- `Toggle.snapshot.test.tsx.snap`: regenerated (drawn div's class now includes `toggleFocusTarget`)

## No behavior change

The same drawn toggle `<div>` still receives `focus-style-css()` on keyboard focus — now via an explicit marker instead of a bare `div` match.

## Testing

- ✅ `@vibe/core` Toggle tests: 18/18 pass
- ✅ stylelint clean

> This is the `version3` counterpart of the same change targeting `master`.

### DIFF
--- a/packages/core/src/components/Toggle/Toggle.module.scss
+++ b/packages/core/src/components/Toggle/Toggle.module.scss
@@ -8,7 +8,7 @@
   // When the hidden checkbox will be focused by keyboard navigation events, the toggle appearance will reflect it
   &:focus-visible,
   &.focus-visible { /* stylelint-disable-line selector-class-pattern */
-    & ~ div {
+    & ~ .toggleFocusTarget {
       @include focus-style-css();
     }
   }
--- a/packages/core/src/components/Toggle/Toggle.tsx
+++ b/packages/core/src/components/Toggle/Toggle.tsx
@@ -117,7 +117,7 @@ const Toggle = forwardRef(
           offOverrideText={offOverrideText}
           onOverrideText={onOverrideText}
           disabled={disabled}
-          className={className}
+          className={cx(className, styles.toggleFocusTarget)}
           selectedClassName={toggleSelectedClassName}
           size={size}
         />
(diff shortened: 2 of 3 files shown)


orrgottlieb:
### ⚠️ `bundle-size` failure is pre-existing CI breakage on `version3` — unrelated to this change

This PR only touches `Toggle.module.scss`, `Toggle.tsx`, and a snapshot. All other checks pass (Build, Test, Lint, Stylelint, Chromatic). The `bundle-size` job fails for an infrastructure reason that predates this branch:

- The **base-report** step in `.github/workflows/bundle-size.yml` runs `yarn add --dev -W size-limit @size-limit/preset-small-lib` **unpinned**.
- Unpinned, it resolves to a `size-limit` whose transitive dependency `nanoid@6.0.1` requires Node `^22 || ^24 || >=26`, but CI runs Node `20.12.2`:
  ```
  error nanoid@6.0.1: The engine "node" is incompatible with this module. Expected version "^22 || ^24 || >=26". Got "20.12.2"
  error Found incompatible module.
  ```
- The step has `continue-on-error: true`, so it reports ✓ but never writes `scripts/bundle-check/reports/base.json`. The following `Compare sizes` step then fails with `ENOENT: ... base.json`.

**This was already fixed on `master` in #3443** ("ci: pin size-limit to v12 to fix bundle check on Node 20"), which pins `size-limit@12` / `@size-limit/preset-small-lib@12`. That fix has not been backported to `version3`, so any PR targeting `version3` hits the same failure.

Fix (one line, when someone backports it to `version3`):
```diff
-          yarn add --dev -W size-limit @size-limit/preset-small-lib
+          yarn add --dev -W size-limit@12 @size-limit/preset-small-lib@12
```

Intentionally **not** included here to keep this PR scoped to the Toggle selector change.

